PR #43, opened 2023-02-16
Python language updates

marksmayo:
- cleaned imports
 - converted format strings to f-strings
 - added trailing commas for python 3.5+ styles
 - removed unnecessary inherit of object from Info
 - replace type() with isinstance where possible (faster)
 - simplified some statements

### DIFF
--- a/keep/action/action.py
+++ b/keep/action/action.py
@@ -1,7 +1,4 @@
 import logging
-import re
-
-import requests
 
 from keep.contextmanager.contextmanager import ContextManager
 from keep.exceptions.action_error import ActionError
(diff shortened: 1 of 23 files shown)


talboren:
@marksmayo is this just an automatic linter? what are all those commas that were added all over? they're not really needed. Also seems like there's a bunch of function assignments that you removed where it seems like it can be completely removed (e.g. https://github.com/keephq/keep/pull/43/files#diff-b1df8f166760ffc77a6522a3e7cccb0d44fa7751d13d93d0a44e361c3cf9901bR59).

talboren:
@marksmayo please also follow the https://www.conventionalcommits.org/en/v1.0.0/ and change the PR name to comply with that 🙏🏼

talboren (changes requested):
Left a few comments

talboren:
@marksmayo I've closed this PR since it seemed stale. let me know if this is still relevant and we can work together to push that in. Thanks!

PR #100, opened 2023-04-02
Adding tests for parser

GabrielSamberg:
THIS IS A WORK IN PROGRESS, SO THE PULL REQUEST IS MARKED AS DRAFT. I PLAN TO ADD MORE TESTS AND ONLY THEN I WILL MARK IT AS READY.

Closes #

## ✅ Checks

### DIFF
--- a/keep/parser/tests_for_parser/test_parser.py
+++ b/keep/parser/tests_for_parser/test_parser.py
@@ -1,0 +1,22 @@
+# here we are going to create all needed tests for the parser.py parse function
+import pytest
+import requests
+
+from keep.parser.parser import Parser
+
+
+def test_parse_with_nonexistent_file():
+    self = Parser()
+    # Expected error when a given input does not describe an existing file
+    with pytest.raises(FileNotFoundError):
+        self.parse('')
+
+
+def test_parse_with_nonexistent_url():
+    self = Parser()
+    # Expected error when a given input does not describe an existing URL
+    with pytest.raises(requests.exceptions.ConnectionError):
+        self.parse('https://ThisWebsiteDoNotExist.com')
+
+
+
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,8 @@ black = "^23.1.0"
 isort = "^5.12.0"
 autopep8 = "^2.0.1"
 flake8 = "^6.0.0"
+pytest = "^7.2.2"
+coverage = "^7.2.2"
 
 [build-system]
 requires = ["poetry-core"]


keep/parser/tests_for_parser/test_parser.py
```diff
@@ -0,0 +1,22 @@
+# here we are going to create all needed tests for the parser.py parse function
+import pytest
+import requests
+
+from keep.parser.parser import Parser
+
+
+def test_parse_with_nonexistent_file():
+    self = Parser()
```
shahargl, line 9: parser = Parser()

keep/parser/tests_for_parser/test_parser.py
```diff
@@ -0,0 +1,22 @@
...
+    # Expected error when a given input does not describe an existing file
+    with pytest.raises(FileNotFoundError):
+        self.parse('')
```
shahargl, line 12: parser.parser('non-existing-file')

keep/parser/tests_for_parser/test_parser.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def test_parse_with_nonexistent_url():
+    self = Parser()
+    # Expected error when a given input does not describe an existing URL
+    with pytest.raises(requests.exceptions.ConnectionError):
+        self.parse('https://ThisWebsiteDoNotExist.com')
```
shahargl, line 19: parser.parse

talboren:
@shahargl @GabrielSamberg let's also add the workflow to execute these tests (and then we'll add a gate that makes sure tests are not broken when new PR is created), wdyt?

keep/parser/tests_for_parser/resources/db_disk_space_for_testing.yml
```diff
@@ -0,0 +1,130 @@
+# Database disk space is low (<10%)
```
shahargl, line 1: Can you put this under `keep/tests/alerts`?

keep/parser/tests_for_parser/resources/db_disk_space_for_testing.yml
```diff
@@ -0,0 +1,130 @@
...
+alert:
+  id: db-disk-space
+  tags: ['string_1','string_2']
```
shahargl, line 4: change?

keep/parser/tests_for_parser/resources/providers_for_testing.yaml
```diff
@@ -0,0 +1,3 @@
+slack-demo:
```
shahargl, line 1: this could be added to the alert file itself or be mocked in the test itself (which I think is better)

.github/workflows/test-pr.yml
```diff
@@ -0,0 +1,19 @@
+name: Tests
```
shahargl, line 1: Nice!

keep/parser/tests_for_parser/resources/providers_for_testing.yaml
GabrielSamberg: The reason behind the  separation between the two  is that  we also want to test the case  where the function is called without the necessary  providers file. It is  because of that test I've discovered the dependency of the different calls to the parse function that is a bug.

shahargl (approved):
lgtm

PR #101, opened 2023-04-03
feat: adding postgres provider + multivalue threshold

shahargl:
adding postgres provider + multivalue threshold + some other improvements


deploy.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
talboren, line 1: Wouldn't want this pushed to the repo, it's keep internal

deploy.sh
talboren, line 1: Delete this file, this is for internal/own usage

keep/action/action.py
```diff
@@ -54,9 +54,11 @@ def _check_conditions(self):
                 # TODO - wrap it else
                 if step == "this":
                     continue
-                for condition in full_context.get("steps").get(step).get("conditions"):
+                for condition, condition_results in (
+                    full_context.get("steps").get(step).get("conditions").items()
+                ):
                     # One of the conditions has been met
```
talboren, line 60: is this doc line still relevant?

keep/action/action.py
```diff
@@ -54,9 +54,11 @@ def _check_conditions(self):
...
-                    if condition.get("result"):
+                    if any([c["result"] for c in condition_results]):
```
talboren, line 61: i'm not 100% sure I understand this one

keep/conditions/base_condition.py
```diff
@@ -59,3 +61,14 @@ def get_compare_value(self):
...
+    def get_additional_context(self):
+        """Get additional context for the condition.
+
+        For example, for threshold conditions it'll be the 'level' so the it will be accessible
+        through the context manager.
+
+        Returns:
+            _type_: dict
+        """
+        return self.condition_context
```
talboren, line 74: feels redundant.. why can't I just access self.condition_context?

keep/conditions/condition_factory.py
```diff
@@ -5,9 +5,11 @@
 
 class ConditionFactory:
     @staticmethod
-    def get_condition(condition_type, condition_config) -> BaseCondition:
+    def get_condition(
+        condition_type, condition_name, condition_config
```
talboren, line 9: why was condition_name added? is it so we can access the condition "results"?

keep/conditions/threshold_condition.py
```diff
@@ -10,19 +10,56 @@ class ThresholdCondition(BaseCondition):
...
+            levels = self.condition_config.get("level")
+            if len(levels.split(", ")) != len(compare_to.split(",")):
+                raise ValueError(
+                    "Number of levels and number of thresholds do not match"
+                )
+            self.levels = [l.strip() for l in levels.split(",")]
+            return True
+        return False
+
+    def _apply_multithreshold(self, compare_to, compare_value):
```
talboren, line 41: side note: we have to change "compare_to" and "compare_value", it makes no sense

talboren, line 41: add docs 🙏🏼

keep/conditions/threshold_condition.py
```diff
@@ -10,19 +10,56 @@ class ThresholdCondition(BaseCondition):
...
 
-    def __init__(self, condition_type, condition_config):
-        super().__init__(condition_type, condition_config)
+    def __init__(self, *kargs, **kwargs):
+        super().__init__(*kargs, **kwargs)
+        self.levels = []
 
-    def apply(self, compare_to, compare_value) -> bool:
-        """apply the condition.
+    def _check_if_multithreshold(self, compare_to):
```
talboren, line 17:
```suggestion
    def _check_if_multithreshold_and_set_levels(self, compare_to):
```
or perhaps split to 2 functions?

keep/contextmanager/contextmanager.py
```diff
@@ -61,7 +63,11 @@ def get_full_context(self):
         full_context = {
             "providers": self.providers_context,
             "steps": self._steps_context,
-            "foreach": {"value": self.foreach_context},
+            # This is a hack to support both "before condition" and "after condition"
```
talboren, line 66: don't understand this one at all

keep/contextmanager/contextmanager.py
```diff
@@ -61,7 +63,11 @@ def get_full_context(self):
...
+            # TODO - fix it and make it more elegant
+            "foreach": self.foreach_context
+            if "value" in self.foreach_context
+            else {"value": self.foreach_context},
```
talboren, line 70: what does that else mean?

keep/contextmanager/contextmanager.py
```diff
@@ -89,21 +97,24 @@ def set_condition_results(
             compare_value (_type_): _description_
             result (_type_): _description_
             condition_alias (_type_, optional): _description_. Defaults to None.
-            raw_value (_type_): the raw value which the condition was compared to. this is relevant only for foreach conditions
+            value (_type_): the raw value which the condition was compared to. this is relevant only for foreach conditions
```
talboren, line 100: why is a foreach condition different from other condition?

keep/contextmanager/contextmanager.py
```diff
@@ -112,15 +123,23 @@ def set_condition_results(
     def get_actionable_results(self):
```
talboren, line 123: can you add docs to this function?

keep/conditions/threshold_condition.py
```diff
@@ -62,6 +112,7 @@ def _apply_threshold(self, step_output, threshold):
         Returns:
             _type_: _description_
         """
+        self._validate(step_output, threshold)
```
talboren, line 115: is this the right place for it? maybe it should be in init?

keep/iohandler/iohandler.py
```diff
@@ -1,8 +1,10 @@
 import ast
 import copy
+import datetime
 import json
 import logging
 import re
+from decimal import Decimal
```
talboren, line 7: redundant?

keep/iohandler/iohandler.py
```diff
@@ -1,8 +1,10 @@
 import ast
 import copy
+import datetime
```
talboren, line 3: redundant?

keep/providers/postgres_provider/postgres_provider.py
```diff
@@ -0,0 +1,134 @@
+"""
+PostgresProvider is a class that provides a way to read data from MySQL.
```
talboren, line 2:
```suggestion
PostgresProvider is a class that provides a way to read/write data from/to Postgres.
```

keep/providers/providers_factory.py
```diff
@@ -48,7 +48,14 @@ def get_provider(
                 + provider_type_split[1].title().replace("_", "")
                 + "Provider",
             )
-        return provider_class(provider_id=provider_id, config=provider_config)
+        try:
+            return provider_class(provider_id=provider_id, config=provider_config)
+        except TypeError as exc:
+            error_message = f"Configuration problem while trying to initialize the provider {provider_id}. Probably missing provider config, please check the provider configuration. [supplied configuration: {provider_config}]]"
```
talboren, line 54: This can also be problematic in a "hosted" version since it discloses secrets in logs

talboren: would remove/simplify

keep/step/step.py
```diff
@@ -51,40 +51,50 @@ def _post_step_validations(self):
...
         for value in foreach_actual_value:
             # will be use inside the io handler
             self.context_manager.set_for_each_context(value)
             for condition in self.step_conditions:
                 condition_type = condition.get("type")
                 condition_alias = condition.get("alias")
-                condition = ConditionFactory.get_condition(condition_type, condition)
+                condition_name = (
+                    condition.get("name", None) or condition_type
+                )  # if name is not supplied, its ok to use the type as the name (todo: this is a hack, we should fix this)
```
talboren, line 69: let's maybe fix that now? make it a "breaking change", otherwise I feel like the hole will just get deeper, wdyt?

talboren (changes requested):
See my comments

keep/action/action.py
shahargl: yes

keep/action/action.py
shahargl: there was a bug in the previous code - it assumed that there is one condition result per condition, which is not the case for "foreach". anyway we are going to rewrite the conditions (get them out of steps)

keep/conditions/condition_factory.py
shahargl: because the previous structure was:
- steps
- - conditions results

so steps.bla.conditions[0] was the first condition

the problem is that in "foreach" mode, there is more than one condition result per condition so the structure does not apply anymore, so now its
- steps
- - conditions
- - - results

keep/iohandler/iohandler.py
shahargl: if you remove it, the "eval" fails, I'll add doc

keep/providers/providers_factory.py
shahargl: removed

keep/step/step.py
shahargl: I think this should be implicit until we have e2e tests, then it'll be easy to fix all underlying alerts.

keep/contextmanager/contextmanager.py
shahargl: added a lot of docs to the function but it still should be refactored

keep/contextmanager/contextmanager.py
shahargl: its for all of them now, not only for foreacb

PR #102, opened 2023-04-03
feat: github (stars) provider + improvements

talboren:
## 📑 Description

- Simpler parsing and full chevron (mustache) support
- Github stars provider
- Merged state and context manager since it was redundant
- action/steps can now access previous runs


docs/docs/state.md
```diff
@@ -3,15 +3,17 @@ sidebar_label: State
 sidebar_position: 6
 ---
 
+# State
+
 ## Intro
-Keep support state to support throttling mechanism's and track alerts over time.
+Keep has a state manager so it can support throttling mechanism's as well as track alerts over time.
```
shahargl, line 9:
```suggestion
Keep State Manager currently used for:
1. Throttling
2. Track alerts over time
```

examples/alerts/new_github_stars.yml
```diff
@@ -0,0 +1,35 @@
+# Alert when there are new GitHub Stars utilizing keepstate
+alert:
+  id: new-github-stars
+  description: Get new GitHub Stars
+  steps:
+    - name: get-github-stars
+      provider:
+        type: github.stars
+        config: " {{ providers.github-pat }}"
```
shahargl, line 9: why do I need github PAT for public repo such as Keep? Maybe we can eliminate it? and then the usage can be easier

examples/alerts/new_github_stars.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          repository: "keephq/keep"
+          previous_stars_count: " {{ state.new-github-stars.-1.alert_context.alert_steps_context.get-github-stars.results.stars }} " # state.alert-id.-1 for last run
```
shahargl, line 12: I guess we should find a better way to do it since this is not indicative.

Maybe something like - 
```
          previous_stars_count: " {{ state.previous_stars_count }} "
```
And add the option to keep the state with a state block
```
state: 
  - name: previous_stars_count
     value: steps.get-github-stars.results.new_stargazers
```

keep/alert/alert.py
```diff
@@ -82,6 +81,7 @@ def run_actions(self):
 
     def run(self):
         self.logger.debug(f"Running alert {self.alert_id}")
+        # todo: check why is this needed?
```
shahargl, line 84: ?

keep/contextmanager/contextmanager.py
```diff
@@ -59,7 +63,7 @@ def set_alert_context(self, alert_context):
     def get_alert_id(self):
         return self.alert_context.get("alert_id")
 
-    def get_full_context(self):
+    def get_full_context(self, exclude_state=False):
```
shahargl, line 66: add docs when exclude state should be used?

keep/contextmanager/contextmanager.py
```diff
@@ -187,4 +195,38 @@ def load_step_context(self, step_id, step_results, step_conditions):
...
+        return {"step_id": step_id, "step_context": self.steps_context.get(step_id)}
+
+    def __load_state(self):
+        if self.state_file:
+            # TODO - SQLite
+            try:
+                with open(self.state_file, "r") as f:
+                    self.state = json.load(f)
+            except:
+                self.state = {}
```
shahargl, line 207: add debug logs

keep/contextmanager/contextmanager.py
```diff
@@ -187,4 +195,38 @@ def load_step_context(self, step_id, step_results, step_conditions):
...
+
+    def get_last_alert_run(self, alert_id):
+        if alert_id in self.state:
+            return self.state[alert_id][-1]
+        # no previous runs
```
shahargl, line 212: should it fail?

keep/contextmanager/contextmanager.py
```diff
@@ -187,4 +195,38 @@ def load_step_context(self, step_id, step_results, step_conditions):
...
+        else:
+            return {}
+
+    def set_last_alert_run(self, alert_id, alert_context, alert_status):
```
shahargl, line 216: add docs + how does `set_last_alert_run` is being used from outer class? shouldn't it be private?

examples/alerts/new_github_stars.yml
talboren: it's not mandatory and can live w/o BUT you are very much limited with API rate limits

examples/alerts/new_github_stars.yml
talboren: We can think about simplifying the "syntax" although it's pretty straightforward, I'll think about a way to extend it and will open an issue

keep/alert/alert.py
talboren: You've added the line
`        self.context_manager.set_alert_context(self._get_alert_context())`
but I couldn't see why it is needed

keep/contextmanager/contextmanager.py
talboren: idk, let's talk about it, you defined this one, I just moved it

keep/contextmanager/contextmanager.py
talboren: it's being called from the alert manager, it's something you've defined, let's think about it?

shahargl (approved):
lgtm

PR #123, opened 2023-04-30
feat: add test to throttling

shahargl:
sanity test to throttling


keep/contextmanager/contextmanager.py
```diff
@@ -206,6 +194,7 @@ def load_action_context(self, action_id, action_results, actions_conditions):
                 condition_alias=condition.get("alias"),
                 value=condition.get("value"),
             )
+        self.set_action_status(action_id, status)
```
talboren, line 197: This is a bit confusing, load_action_context calls set_action_status? why?

keep/contextmanager/contextmanager.py
```diff
@@ -116,7 +116,9 @@ def set_condition_results(
         value=None,
         **kwargs,
     ):
-        """_summary_
+        """Save the condition results in the context manager
+
+        TODO: this should be refactored to something more structured
```
talboren, line 121: 👍🏼, should we have an issue for this?

keep/action/action.py
```diff
@@ -116,7 +126,7 @@ def _run_single(self):
                 self.config.get("name"),
                 if_conf,
             )
-            return
+            return ActionStatus.SKIPPED
```
talboren, line 129: update _run_single return type (-> ActionStatus)

tests/test_throttling.py
```diff
@@ -0,0 +1,43 @@
+import io
```
talboren, line 1: would add a comment to this test, since it is based on debug logs (correct me if I am wrong, because of the -vvv) - it'll be flaky since if a line of **log** changes, this will break the test (although the logic might be working).
perhaps it is good to think about checking keepstate?

talboren (commented):
4 minor comments

PR #128, opened 2023-05-07
test: functions.diff

blaklafter:
add pytest to pyproject.toml dev deps
functions: add typing to diff function

## 📑 Description

I tried to run the python tests, but noticed that pytest wasn't installed, [even though the import was used in the sanity tests](https://github.com/keephq/keep/blob/b07a4da58d5038b4745e9752d0ff14e67d65efd7/tests/test_sanity.py#L6).

For fun, I also added some tests around the diff function.  Pointing out a couple of behaviors that might be surprising.  I didn't change any code behavior.

## ✅ Checks

### DIFF
--- a/keep/functions/__init__.py
+++ b/keep/functions/__init__.py
@@ -14,7 +14,7 @@ def all(iterable):
     return next(g, True) and not next(g, False)
 
 
-def diff(iterable):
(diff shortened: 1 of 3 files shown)


shahargl (approved):
lgtm

PR #144, opened 2023-05-21
feat: add keep.encode() to functions

mostafahussein:
This PR adds a new function `keep.encode()` which can be useful in case of adding urls to slack notifications, which needs to be encoded.

Usage:
```
keep.encode("{{ steps.myalert.provider_parameters.query}}")
```


shahargl (approved):
Lgtm!

PR #178, opened 2023-06-16
feat: add auth0 provider

shahargl:


examples/alerts/new_auth0_users.yml
```diff
@@ -0,0 +1,36 @@
+# Alert when there are new GitHub Stars utilizing keepstate
```
talboren, line 1:
```suggestion
# Alert when there are new Auth0 users
```

keep/providers/auth0_provider/auth0_provider.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+    def notify(self, **kwargs):
+        pass  # Define how to notify about any alerts or issues
+
+    def dispose(self):
+        pass
+
+
+class Auth0LogsProvider(Auth0Provider):
```
talboren, line 63: Doesn't really seem to let you query logs but rather more specific to new users?

keep/providers/auth0_provider/auth0_provider.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+if __name__ == "__main__":
+    import logging
+
+    logging.basicConfig(level=logging.DEBUG, handlers=[logging.StreamHandler()])
+
+    # If you want to use application default credentials, you can omit the authentication config
+    config = {
+        # "authentication": {"credentials_path": "/path/to/your/service_account.json"},
```
talboren, line 133: ?

keep/providers/auth0_provider/auth0_provider.py
shahargl: It does (you query by log type)

PR #186, opened 2023-06-21
chore: Enhancement: Merge step and action class implementation

### LINKED ISSUE #180: Enhancement: Merge step and action class implementation (e.g. base class)
Actions and Steps are practically the same entities, but the fact that Actions has conditions (and we want conditions for steps as well) and "throttling" logic, and Step calls `query` when executed while Actions calls `notify`.

The reason this is now needed is that we want to add an `on-failure` strategy for both Actions and Steps and would like to have the same logic implemented for both. (And in general, this can reduce code and complexity).

joaquinto:
Closes #180  

## 📑 Description

This PR consolidates the Actions and Steps class. See reason below;
Actions and Steps are practically the same entities, but the fact that Actions has conditions (and we want conditions for steps as well) and "throttling" logic, and Step calls `query` when executed while Actions calls `notify`.

The reason this is now needed is that we want to add an `on-failure` strategy for both Actions and Steps and would like to have the same logic implemented for both. (And in general, this can reduce code and complexity).

## ✅ Checks


talboren:
Hey @joaquinto,

First of all, thank you so much for this contribution. It looks awesome!

There are 2 points I'd like to touch on here:
1. I think that we should somehow try to consolidate also the "logic" behind actions & steps (e.g. a step should also support conditions and throttling) -- the only difference should be on whether to call the query or notify function (IMO)
2. I'd love it if you can update the documentation if it is needed anywhere (probably on this page: https://docs.keephq.dev/platform/core/syntax/basic-syntax)

Would love to hear your thoughts.


@shahargl any other thoughts you have?

keep/step/step.py
```diff
@@ -1,31 +1,53 @@
...
         self.context_manager = ContextManager.get_instance()
+        self.conditions = self.config.get("condition", [])
 
     @property
     def foreach(self):
         return self.step_config.get("foreach")
-
+    
     def run(self):
+        if self.step_type == StepType.STEP:
```
shahargl, line 46: if action and steps are the same, why is there a "_run_step" and "_run_action"? what's the difference between them?

talboren: > if action and steps are the same, why is there a "_run_step" and "_run_action"? what's the difference between them?

Only difference is calling notify or query

shahargl: so it should be something like:
```
class Runnable():
    def run(...):
         return self._run()
    
class Step(Runnable):
     def run(...):
          return self._query()
          
class Action(Runnable):
     def run(...):
          return self._notify()
```

shahargl: and I'm not even sure why we need to distinguish between query and notify, we can consolidate it too

shahargl: @joaquinto and thanks for your contribution! <3

joaquinto: @talboren ok. I will look into it.

keep/step/step.py
```diff
@@ -1,51 +1,198 @@
...
+                    rendered_providers_parameters.update(extra_context)
+                    self.context_manager.set_step_provider_paremeters(
+                        self.step_id, rendered_providers_parameters
+                    )
+                except Exception as e:
+                    raise StepError(e)
+                
+                return step_output
+            self.provider.notify(**rendered_value)
+            return True
```
talboren, line 180: I think we need to try and keep logic as similar as possible, wdyt about the following:

```suggestion
        if inspect.iscoroutinefunction(self.provider.notify) or inspect.iscoroutinefunction(self.provider.query):
            result = self._run_single_async()
        # else, just run the provider
        else:
              try:
                  rendered_providers_parameters = {}
                  for parameter in self.provider_parameters:
                      rendered_providers_parameters[parameter] = self.io_handler.render(
                          self.provider_parameters[parameter]
                      )
                  step_output =  self.provider.query(**rendered_value) if self.step_type == StepType.STEP else self.provider.notify(**rendered_value)
                  extra_context = self.provider.expose()
                  rendered_providers_parameters.update(extra_context)
                  self.context_manager.set_step_provider_paremeters(
                      self.step_id, rendered_providers_parameters
                  )
              except Exception as e:
                  raise StepError(e)
```

keep/step/step.py
```diff
@@ -1,51 +1,198 @@
...
+    def _run_single_async(self):
+        """For async providers, run them in a new event loop
+
+        Raises:
+            ActionError: _description_
+        """
+        rendered_value = self.io_handler.render_context(self.provider_context)
+        # This is "magically solved" because of nest_asyncio but probably isn't best practice
+        loop = asyncio.new_event_loop()
+        task = loop.create_task(self.provider.notify(**rendered_value))
```
talboren, line 191:
```suggestion
        if self.step_type == StepType.STEP:
            task = loop.create_task(self.provider.query(**rendered_value))
        else:
            task = loop.create_task(self.provider.notify(**rendered_value))
```

keep/parser/parser.py
```diff
@@ -242,25 +242,27 @@ def _get_action(self, action: dict, action_name: str | None = None) -> Action:
...
         return action
 
-    def _parse_actions(self, alert) -> typing.List[Action]:
+    def _parse_actions(self, alert) -> typing.List[Step]:
         self.logger.debug("Parsing actions")
         alert_actions = alert.get("actions", [])
         alert_actions_parsed = []
         for _action in alert_actions:
             parsed_action = self._get_action(_action)
+            parsed_action.step_type = StepType.ACTION
```
talboren, line 260: Seems redundant, as https://github.com/keephq/keep/pull/186/files#diff-d70b456d958384fcd43980d14f1c8335a30d72522715c7d9b6b7a9d02845a12dR245 already handles that, no?

joaquinto: Yes. That has been handled already in the `_get_action` method. Thank you for pointing that out.

keep/step/step.py
joaquinto: Ok. I will add that. Thank you for pointing that out.

talboren:
@shahargl can you also review it, please?

shahargl:
lgtm

PR #193, opened 2023-06-26
feat: adding frontend endpoints

shahargl:


keep/cli/cli.py
```diff
@@ -1,18 +1,31 @@
...
 
 load_dotenv(find_dotenv())
+if not os.getenv("DISABLE_POSTHOG"):
+    posthog_api_key = (
+        os.getenv("POSTHOG_API_KEY")
+        or "phc_muk9qE3TfZsX3SZ9XxX52kCGJBclrjhkP9JxAQcm1PZ"
+    )
+    posthog_client = Posthog(api_key=posthog_api_key, host="https://app.posthog.com")
+else:
+    posthog_client = None
```
talboren, line 26: perhaps put this in a helper and use it also in the API? currently you can't "shut" posthog for the api and this code is duplicate

keep/api/api.py
```diff
@@ -35,6 +38,67 @@ async def dispose_context_manager() -> None:
     ContextManager.delete_instance()
 
 
+class EventCaptureMiddleware(BaseHTTPMiddleware):
+    def __init__(self, app: FastAPI):
+        super().__init__(app)
```
talboren, line 43: is this init being called only once when adding the middleware or for every request? if it's for every request, maybe it's better to initalize posthog client once?

talboren (approved):
looks good, 2 small comments!

keep/api/api.py
shahargl: its once on init

PR #233, opened 2023-08-04
ci: use ci cache to cache poetry

blaklafter:
## 📑 Description
I noticed there wasn't any caching around poetry installed packages.  This can really help keep CI times down if none of the poetry dependencies change.

I set up Actions to cache based on a hash of the poetry lock file.  This also required me to remote `poetry.lock` from `.gitignore` and check it in.  This file is usually checked in, I'm not sure if there are some specific reasons why this file is not being checked in. `poetry.lock` was removed [here](https://github.com/keephq/keep/commit/d3a364d9491011611e4fb24bd076e55b70615f98). (cc @talboren :eyes: )

## ✅ Checks


talboren (approved):
LGTM, thanks for that @blaklafter 
Not sure why I did this back then, I’m gonna look on what’s the best practice here and merge it!

PR #190, opened 2023-06-26
feat: add interval as prop

shahargl:

PR #234, opened 2023-08-06
fix: CI Action Run Keep / Demo has been failing

blaklafter:
## 📑 Description

I noticed that the GH Action for [Run Keep / Demo](https://github.com/keephq/keep/actions/workflows/run-keep.yml) has been failing for nearly 2 months. Although the functionality is technically working - the results are probably not intended.

GitHub limits the running of their runners to a maximum of 6 hours, and that is when this job times out and "fails". Every minute this alert is generated, for a maximum of 360 alerts! Every... day (for this repo and every fork on GitHub). Probably overkill.

I think there is more benefit to having this test run once, than to try to exercise the interval. This will also cut down on the CI hours, and generally be a better net citizen :smiley_cat: 

This break was introduced [here](https://github.com/keephq/keep/pull/190).

## ✅ Checks

## ℹ Additional Information

This is just a simple fix to get back to a reasonable CI process.


talboren:
@blaklafter thank you for this! I actually noticed this a few days ago and had a task to fix this.

talboren (approved):
LGTM!

PR #238, opened 2023-08-08
test: parse providers from env vars

blaklafter:
## 📑 Description

Added some tests around parsing providers from env vars.

## ✅ Checks

## ℹ Additional Information

There might be some behavior when parsing a malformed env var that is unintentional. See comments...

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1,4 +1,5 @@
 # here we are going to create all needed tests for the parser.py parse function
+import json
 import os
 from pathlib import Path
 from unittest.mock import patch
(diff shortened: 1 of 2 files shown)


tests/test_parser.py
```diff
@@ -87,3 +88,93 @@ def test_parse_with_alert_source_with_no_providers_file():
...
+
+        #ASSERT
+        expected = {provider_name.replace("_", "-").lower(): provider_dict}
+        
+        # This might be a bug?
+        # It will create a provider context with an empty string as a provider name: {'': {'hi': 1}}
+        assert parser.context_manager.providers_context == expected
+
+        # I would expect it to not create the provider
+        #assert parser.context_manager.providers_context == {}
```
blaklafter, line 176: This might not be intentional. I wanted to draw attention to it as a corner case.

talboren (approved):
LGTM, thank you for that!

tests/test_parser.py
talboren: i.e. a "nameless" provider. That is an edge case but probably not to interesting to tackle right now.

PR #241, opened 2023-08-08
test: add tests for loading providers by file

blaklafter:
## 📑 Description

Added some unit tests around loading providers from a file

## ✅ Checks

### DIFF
--- a/keep/parser/parser.py
+++ b/keep/parser/parser.py
@@ -181,7 +181,10 @@ class Parser:
     def _get_step_provider(self, _step: dict) -> dict:
         step_provider = _step.get("provider")
         step_provider_type = step_provider.pop("type")
-        step_provider_config = step_provider.pop("config")
+        try:
+            step_provider_config = step_provider.pop("config")
+        except KeyError:
+            step_provider_config = {"authentication": {}}
         provider_id, provider_config = self._parse_provider_config(
             step_provider_type, step_provider_config
         )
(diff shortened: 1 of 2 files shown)


keep/parser/parser.py
```diff
@@ -181,10 +181,7 @@ def _parse_steps(self, alert) -> typing.List[Step]:
     def _get_step_provider(self, _step: dict) -> dict:
         step_provider = _step.get("provider")
         step_provider_type = step_provider.pop("type")
-        try:
-            step_provider_config = step_provider.pop("config")
-        except KeyError:
-            step_provider_config = {"authentication": {}}
```
blaklafter, line 187: This exception should be "impossible" because the `BaseProvider` class requires a config attribute (`ProviderConfig`) : https://github.com/keephq/keep/blob/a1e8ece787e059e1f58ba54d544388e0a16d2362/keep/providers/base/base_provider.py#L22C22-L22C22

I spent a while trying to get this thing to fail as organically as possible... but ended up with the `test_parse_alert_steps` test.

talboren: I'm unsure if it's a leftover or something we had in place for providers that do not have configuration (e.g., PythonProvider, MockProvider, etc.). Need to test if that change wouldn't fail those providers in that case.

talboren (approved):
LGTM, thanks again @blaklafter

Issue #240, opened 2023-08-08
Missing provider Zabbix


SomeAverageDev:
**Is your feature request related to a problem? Please describe.**
no, just a use case

**Describe the solution you'd like**
centralize all Zabbix problems in a single keep pane of glass

**Describe alternatives you've considered**
-

**Additional context**
-

talboren:
@SomeAverageDev thank you for this issue! would love to further discuss this with you so I can get to it ASAP.
Would you mind joining [Keep's Slack community](https://slack.keephq.dev) so we can chat over there?

talboren:
@SomeAverageDev would love for you to try it ;)

talboren:
Closed at #246 and #251

PR #246, opened 2023-08-17
feat: Zabbix provider push capabilities

talboren:


talboren:
This will partially handle #240

talboren:
@SomeRandomDev FYI

PR #254, opened 2023-08-20
feat: cloudwatch webhooks

shahargl:


keep/api/routes/alerts.py
```diff
@@ -139,6 +155,11 @@ def receive_event(
         # Each provider should implement a format_alert method that returns an AlertDto
         # object that will later be returned to the client.
         formatted_event = provider_class.format_alert(event)
+        # If the format_alert does not return an AlertDto object, it means that the event
+        # should not be pushed to the client.
+        if not formatted_event:
+            raise {"status": "ok"}
```
talboren, line 161: raise?

keep/api/routes/alerts.py
```diff
@@ -120,13 +122,27 @@ def get_alerts(
 @router.post(
     "/event/{provider_type}", description="Receive an alert event from a provider"
 )
-def receive_event(
+async def receive_event(
```
talboren, line 125: just putting this down here: we know this one is problematic with cloud run

keep/api/routes/providers.py
```diff
@@ -211,7 +211,12 @@ def delete_provider(
     )
     secret_manager = SecretManagerFactory.get_secret_manager()
     secret_name = f"{tenant_id}_{provider_type}_{provider_id}"
-    secret_manager.delete_secret(secret_name)
+    try:
+        secret_manager.delete_secret(secret_name)
+    except Exception as exc:
+        # TODO: handle it better
+        logger.exception("Failed to delete the provider secret")
+        pass
```
talboren, line 219: This will show in the UI as if the provider was deleted but won't actually deleted (back after refresh)

keep/api/routes/providers.py
```diff
@@ -231,6 +236,11 @@ def install_provider_webhook(
         provider_id, provider_type, provider_config
     )
     api_url = config("KEEP_API_URL")
+    if "0.0.0.0" in api_url or "127.0.0.1" in api_url or "localhost" in api_url:
+        raise HTTPException(
+            status_code=400,
+            detail="Invalid KEEP_API_URL - webhook installation must be a public URL, you can restart keep and run it with USE_NGROK=true to get a public URL",
+        )
```
talboren, line 243: Let's handle this when we meet this
```suggestion
```

PR #299, opened 2023-09-05
db pvc fixes and k8s min version removed

DmarshalTU:
1. fixed keep-db naming in yamls
2. keep-db get pv and pvc based on storageClassName from values.yaml
3. removed min k8s version for more "old" k8s clusters

### DIFF
--- a/chart/keep/values.yaml
+++ b/chart/keep/values.yaml
@@ -102,6 +102,7 @@ database:
   enabled: true
   replicaCount: 1
   size: 5Gi
+  storageClasss: deafult
   image:
     repository: mysql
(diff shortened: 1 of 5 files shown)


chart/keep/values.yaml
```diff
@@ -102,6 +102,7 @@ database:
   enabled: true
   replicaCount: 1
   size: 5Gi
+  storageClasss: deafult
```
talboren, line 105: @DmarshalTU 
is it default?
```suggestion
  storageClasss: default
```

DmarshalTU: thank to point it out, it should be empty

PR #316, opened 2023-09-10
feat: added support for step retry

### LINKED ISSUE #239: Step retry mechanism
Sometimes when a step fails to run (e.g. a simple HTTP request), I'd like to be able to have a retry mechanism before failing the step.

An optional retry mechanism for steps to define what to do when a step fails.

-

**Additional context**
Requested by @configurator

devblin:
Closes #239 

## 📑 Description
Added support for retrying step on failure for specified count and interval.
Example:
```
# Alert if a result queried from the DB is above a certain thershold.
alert:
  id: raw-sql-query
  description: Monitor that time difference is no more than 1 hour
  steps:
    - name: get-max-datetime
      provider:
        type: mysql
        config: "{{ providers.mysql-prod }}"
        with:
          # Get max(datetime) from the random table
          query: "SELECT MAX(datetime) FROM demo_table LIMIT 1"
        on-failure:
          retry: 
            count: 4
            interval: 15
  actions:
    - name: trigger-ntfy
      condition:
      - name: threshold-condition
        type: threshold
```

## ✅ Checks

/claim #239


devblin:
@talboren What are the steps to update doc?

talboren:
> @talboren What are the steps to update doc?

@devblin I think the <current> best place to update the documentation regarding this new feature would be under `Step/Action` in https://docs.keephq.dev/workflows/syntax/basic-syntax in the `basic-syntax.mdx` file.

keep/step/step.py
```diff
@@ -167,28 +172,34 @@ def _run_single(self):
...
-                for parameter in self.provider_parameters:
-                    rendered_providers_parameters[parameter] = self.io_handler.render(
-                        self.provider_parameters[parameter]
+            for curr_retry_count in range(self.__retry_count + 1):
+                try:
+                    rendered_providers_parameters = {}
+                    for parameter in self.provider_parameters:
+                        rendered_providers_parameters[parameter] = self.io_handler.render(
+                            self.provider_parameters[parameter]
+                        )
```
talboren, line 181: This can be taken out of the for loop, there's no need to re-render the parameters every time.

talboren, line 181:
```suggestion
            rendered_providers_parameters = {}
            for parameter in self.provider_parameters:
                rendered_providers_parameters[parameter] = self.io_handler.render(
                    self.provider_parameters[parameter]
                )
            for curr_retry_count in range(self.__retry_count + 1):
                try:
```

keep/step/step.py
```diff
@@ -167,28 +172,34 @@ def _run_single(self):
...
+                            self.step_id, results=step_output, foreach=self.foreach
+                        )
+                    else:
+                        self.provider.notify(**rendered_value)
+
+                    extra_context = self.provider.expose()
+                    rendered_providers_parameters.update(extra_context)
+                    self.context_manager.set_step_provider_paremeters(
+                        self.step_id, rendered_providers_parameters
                     )
```
talboren, line 195: One of these may fail and we probably don't want to re-run the step in that case, @shahargl wdyt?

keep/step/step.py
devblin: @talboren Updated with requested changes

keep/step/step.py
devblin: @talboren I think we should handle these if we consider this as a part of step execution. If you see below yml example, the `on-failure`, should handle each step, and I think whatever is involved in successful execution of the step, it should be handled, wdyt? 
```
# Alert if a result queried from the DB is above a certain thershold.
alert:
  id: raw-sql-query
  description: Monitor that time difference is no more than 1 hour
  steps:
    - name: get-max-datetime
      provider:
        type: mysql
        config: "{{ providers.mysql-prod }}"
        with:
          # Get max(datetime) from the random table
          query: "SELECT MAX(datetime) FROM demo_table LIMIT 1"
      on-failure:
         retry: 
             count: 4
             interval: 15
  actions:
    - name: trigger-ntfy
      condition:
      - name: threshold-condition
        type: threshold
```

talboren: @devblin I'm not sure I'm 100% convinced since 
```python
extra_context = self.provider.expose()
rendered_providers_parameters.update(extra_context)
self.context_manager.set_step_provider_paremeters(
    self.step_id, rendered_providers_parameters
)
```
is a bit more "infrastructure"ish -- if something in there fails, it should be some kind of "WTFException" or something like that.

talboren: @shahargl wdyt ^?

talboren (changes requested):
1. In general, I would love to see a unit test covering the `step` on failure mechanism
2. You reverted the changes for the step parameters rendering mechanism, is that on purpose?

keep/step/step.py
shahargl: so fwiw I think that  the "retry" mechanism is for handling where the *provider* fails and not Keep, so for that matter, only the `notify`/`query` should be in the "retry" block - all other parts are kinda deterministic and retry won't help with that. @talboren

keep/step/step.py
```diff
@@ -167,28 +172,34 @@ def _run_single(self):
...
-                self.context_manager.set_step_provider_paremeters(
-                    self.step_id, rendered_providers_parameters
-                )
-            except Exception as e:
-                raise StepError(e)
+                except Exception as e:
+                    if curr_retry_count == self.__retry_count:
+                        raise StepError(e)
+                    else:
+                        self.logger.info(
```
shahargl, line 200: anyway it worth adding the original exception so we will have some trace (str(e))

keep/step/step.py
```diff
@@ -167,28 +172,34 @@ def _run_single(self):
             result = self._run_single_async()
```
shahargl, line 172: the async should be fixed too (or at least adding a test for that)

keep/step/step.py
devblin: @shahargl you mentioned that `notify`/`query` method(s) execution should be handled for retries, so the `on-failure` prop should be available for `action` step too? @talboren

keep/step/step.py
devblin: @shahargl you mean handling it for retry?

devblin:
> 1. In general, I would love to see a unit test covering the `step` on failure mechanism
> 2. You reverted the changes for the step parameters rendering mechanism, is that on purpose?

@talboren yes, I did the change, but realised that we should have a discussion regarding which part of the `_run_single` method should be handled for retries, so I just reverted it.

talboren (approved):
LGTM! great job @devblin!

PR #350, opened 2023-09-22
feat: Add export options and sqlaclhemy/requests/metrics intrumentation

bck01215:
## 📑 Description
By setting the OTLP_ENDPOINT the otel collector will now try to export data. Setting METRIC_OTEL_ENABLED to true will also enable metrics to be sent. Currently only OTLP endpoints work. Work was also done to auto-instrument sqlalchemy and the requests library


docker/Dockerfile.dev.ui
```diff
@@ -10,7 +10,7 @@ WORKDIR /app
 
 # Copy package.json and package-lock.json
 # to the /app working directory
-COPY keep-ui/package*.json /app
+COPY keep-ui/package*.json /app/
```
bck01215, line 13: I was receiving a build error that I needed a trailing slash for this `COPY`. I can remove this from the PR if it is only affecting my docker env

shahargl (approved):
lgtm

shahargl (approved):

PR #361, opened 2023-09-26
fix: update planner bucket_id detail

devblin:


devblin:
@shahargl Minor doc update for planner.

Issue #423, opened 2023-10-30
add user/password authentication for self hosted deployments


shahargl:
currently, self-hosted deployments are by default deployed with "SINGLE_TENANT" so you don't have users. 
we want to add the functionality of managing teams with self-hosted deployments.

shahargl:
closed by https://github.com/keephq/keep/pull/436

PR #436, opened 2023-11-02
feat: add users management to single tenant

shahargl:


shahargl:
close https://github.com/keephq/keep/issues/423

PR #513, opened 2023-11-19
feat: better auth control

shahargl:


talboren (changes requested):
LGTM, just remove the customAuth file 🚀

PR #536, opened 2023-11-22
feat: add the ability to run workflows with alert trigger from UI

shahargl:


shahargl:
close https://github.com/keephq/keep/issues/537

Issue #543, opened 2023-11-23
add scopes to OpsGenie provider


shahargl:

surendrasomai:
@talboren , Can you provide me detail ?

talboren:
@surendrasomai you can take a look at Datadog Provider/Grafana Provider/Dynatrace Provider for reference on how to implement the `ProviderScopes` and `validate_scopes()` functions.
If it's still not 100% clear, I'd be happy to elaborate further.

sakshambp:
How do we test the change @talboren what is success criteria?

sakshambp:
Also do we know what all scopes to add?

TakshPanchal:
Hey @talboren, I searched for the API endpoint to get permissions, but I couldn't find any! Lastly, we need `userid` or `username` to [get user role](https://docs.opsgenie.com/docs/user-api#get-user).

TakshPanchal:
> Hey @talboren, I searched for the API endpoint to get permissions, but I couldn't find any! Lastly, we need `userid` or `username` to [get user role](https://docs.opsgenie.com/docs/user-api#get-user).

@shahargl can you help me here 😅

shahargl:
@TakshPanchal couldn't find easy way to do it :( what I suggest is that you'll try to delete it after creating (although delete permissions are higher). let's go with it and wait for user feedback.

ezhil56x:
> @TakshPanchal couldn't find easy way to do it :( what I suggest is that you'll try to delete it after creating (although delete permissions are higher). let's go with it and wait for user feedback.

I have implemented a way to validate the users on the next commit. Can you review my pull request

PR #557, opened 2023-11-24
feat: added scopes-OpsGenie-provider

ezhil56x:
Closes # 

## 📑 Description

Added scopes for OpsGenie provider
/claim #543

## ✅ Checks

### DIFF
--- a/keep/providers/opsgenie_provider/opsgenie_provider.py
+++ b/keep/providers/opsgenie_provider/opsgenie_provider.py
@@ -7,7 +7,7 @@ from opsgenie_sdk.rest import ApiException
 
 from keep.contextmanager.contextmanager import ContextManager
 from keep.providers.base.base_provider import BaseProvider
-from keep.providers.models.provider_config import ProviderConfig
+from keep.providers.models.provider_config import ProviderConfig, ProviderScope
 
 
 @pydantic.dataclasses.dataclass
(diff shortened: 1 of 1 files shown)


keep/providers/opsgenie_provider/opsgenie_provider.py
```diff
@@ -30,13 +30,37 @@ class OpsGenieRecipient(pydantic.BaseModel):
...
         super().__init__(context_manager, provider_id, config)
         self.configuration = opsgenie_sdk.Configuration()
         self.configuration.api_key["Authorization"] = self.authentication_config.api_key
-
+        
+    def validate_scopes(self):
+        scopes = {}
+        self.logger.info("Validating scopes")
+        try:
+            self._create_alert(
```
shahargl, line 53: @ezhil56x yo folk, the problem is that we can't just "create alert" on users' OpsGenie, we need to somehow validate it without actually changing the data. I would look for some API that returns the scopes for the creds.

ezhil56x: Missed that will work on it

ezhil56x: > @ezhil56x yo folk, the problem is that we can't just "create alert" on users' OpsGenie, we need to somehow validate it without actually changing the data. I would look for some API that returns the scopes for the creds.

For now only Creating Alert is supported on keep right?

ezhil56x:
@shahargl I hope it will be validated now

keep/providers/opsgenie_provider/opsgenie_provider.py
shahargl: Yes

PR #620, opened 2023-12-11
feat: rule engine (do not merge)

shahargl:


examples/workflows/create_jira_ticket_upon_alerts copy.yml
talboren, line 1: Remove or rename?

keep/rulesengine/rulesengine.py
```diff
@@ -0,0 +1,127 @@
...
+                self.logger.info("Created alert")
+
+        self.logger.info("Rules ran, alerts created")
+        return
+
+    def _extract_subrules(self, expression):
+        # CEL rules looks like '(source == "sentry") && (source == "grafana" && severity == "critical")'
+        # and we need to extract the subrules
+        sub_rules = expression.split(") && (")
+        # the first and the last rules will have a ( or ) at the beginning or the end
```
talboren, line 96: what does that mean?

keep/rulesengine/rulesengine.py
```diff
@@ -0,0 +1,127 @@
...
+        sub_rules[0] = sub_rules[0][1:]
+        sub_rules[-1] = sub_rules[-1][:-1]
```
talboren, line 98: can you add an example?

keep/rulesengine/rulesengine.py
```diff
@@ -0,0 +1,127 @@
...
+        sub_rules = self._extract_subrules(rule.definition_cel)
+        payload = event.dict()
+        # workaround since source is a list
+        # todo: fix this in the future
+        payload["source"] = payload["source"][0]
+        for sub_rule in sub_rules:
+            env = celpy.Environment()
+            ast = env.compile(sub_rule)
+            prgm = env.program(ast)
+            r = prgm.evaluate(payload)
```
talboren, line 112: some docs for this?

talboren: isn't this "dangerous" (couldn't sub_rule have some malicious code?)

keep/rulesengine/rulesengine.py
```diff
@@ -0,0 +1,127 @@
...
+        # CEL rules looks like '(source == "sentry") && (source == "grafana" && severity == "critical")'
+        # and we need to extract the subrules
+        sub_rules = expression.split(") && (")
+        # the first and the last rules will have a ( or ) at the beginning or the end
+        sub_rules[0] = sub_rules[0][1:]
+        sub_rules[-1] = sub_rules[-1][:-1]
+        return sub_rules
+
+    # TODO: a lot of unit tests to write here
+    def _run_rule_on_event(self, rule, event: AlertDto):
```
talboren, line 102: i'm a bit confused about `_run_rule_on_event` and `_run_rule`, if I'm not mistaken, this "validates" whether the event is **relevant** for a rule? and `_run_rule` actually runs the rule?

keep/rulesengine/rulesengine.py
```diff
@@ -0,0 +1,127 @@
...
+from keep.api.core.db import run_rule as run_rule_db
+from keep.api.models.alert import AlertDto
+
+
+class RulesEngine:
+    def __init__(self, tenant_id=None):
+        self.tenant_id = tenant_id
+        self.logger = logging.getLogger(__name__)
+
+    def run_rules(self, events: list[AlertDto]):
```
talboren, line 17: Logs are a bit strange here.
You are "Running rules" twice:
1. once when you validate if an event is relevant to some rule
2. when you actually run the rule.

perhaps logs should be "validating rule relevancy" and "running rule"?

keep/rulesengine/rulesengine.py
```diff
@@ -0,0 +1,127 @@
...
+        self.logger.info("Running rules")
+        rules = get_rules_db(tenant_id=self.tenant_id)
+
+        # first, we need to understand which rules are actually relevant for the events that arrived
+        relevent_rules_for_events = []
+        for rule in rules:
+            self.logger.info(f"Running rule {rule.name}")
+            for event in events:
+                self.logger.info(f"Running rule {rule.name} on event {event.id}")
```
talboren, line 26: i.e.
```suggestion
                self.logger.info(f"Checking rule {rule.name} relevancy for event {event.id}")
```

talboren (changes requested):
a few comments

keep/rulesengine/rulesengine.py
shahargl: added

keep/rulesengine/rulesengine.py
shahargl: Nope since its just Common Expression Logic (CEL) and its not actually running any code - https://github.com/google/cel-spec and the pyhton implementation - https://github.com/cloud-custodian/cel-python

anyway I've added some comments

PR #644, opened 2023-12-24
fix: simplify env var deployments

shahargl:


keep-ui/middleware.tsx
```diff
@@ -1,13 +1,24 @@
...
-    // 2. keep svg (so that it can be displayed on the signin page)
     '/((?!keep_big\\.svg$|signin$).*)',
   ],
 };
+
+export function middleware(req: NextRequest) {
+  const { pathname } = new URL(req.url);
+  // Redirect /backend/ to the API
+  if (pathname.startsWith('/backend/')) {
+    let apiUrl = getApiURL();
```
talboren, line 16:
```suggestion
    const apiUrl = getApiURL();
```

keep-ui/utils/apiUrl.ts
talboren, line 1: format file

PR #657, opened 2023-12-26
feat(ui): Delete Multiple Alerts

### LINKED ISSUE #561: Add an option to delete multiple alerts
Now that we support react-table (thanks to @asharonbaltazar!) we want to be able to delete multiple alerts at once.

Alert menu will be moved to the right as an ellipse.
When clicking an alert row, checkbox to the left will be marked and user will be able to check multiple alerts and hit a delete button.
Under the hood, multiple delete requests will be sent -- in the future, we will add a bulk delete operation on the backend side to make it more efficient.

asharonbaltazar:
Closes #561 

## 📑 Description

- add props to `AlertTable` for row selection. create optional checkbox for row selection
- create `AlertActions` for the panels that support row selection. replaces `ActionPresets` with action buttons
- move `AlertPresets` into `TabContent` so row selection logic can be shared between the two

https://github.com/keephq/keep/assets/58940073/0ae5c378-1b3a-4228-80e5-95b006aeceed

## ✅ Checks


talboren:
2 bugs that I found:

1. Since the position of the menu is now fixed, scrolling the page kinda screws the position of the menu (I'm also a bit worried about the menu being "hidden" in the right sector of the row, it might cause people to miss the menu button when their screen is 2 small, any suggestions for that?, see the second video)

https://github.com/keephq/keep/assets/68807791/ff51f060-b1b9-4c78-83a3-21f3e8e147eb

https://github.com/keephq/keep/assets/68807791/061f9e9e-41a5-4f57-88a4-d24b1f69cc29


2. It seems like although I am only selecting 1 alert to be deleted, all the alerts in the same view gets deleted 🤯 

https://github.com/keephq/keep/assets/68807791/ee34f934-1b1e-4513-a2fd-167e290723d3

keep-ui/app/alerts/alert-menu.tsx
```diff
@@ -161,7 +158,7 @@ export default function AlertMenu({
           leaveFrom="transform opacity-100 scale-100"
           leaveTo="transform opacity-0 scale-95"
         >
-          <Menu.Items className="z-50 relative mt-2 min-w-36 origin-top-right divide-y divide-gray-100 rounded-md bg-white shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none">
+          <Menu.Items className="z-50 fixed right-12 mt-2 divide-y divide-gray-100 rounded-md bg-white shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none">
```
talboren, line 161: This makes the menu scroll down with the user scrolling. We need to think this through.

keep-ui/app/alerts/alerts.tsx
```diff
@@ -237,7 +239,59 @@ export default function Alerts({
...
+
+    const amountOfRowSelections = Object.keys(rowSelection).length;
+
+    return (
+      <TabPanel className="mt-4">
+        {amountOfRowSelections ? (
+          <AlertActions
+            amountOfRowSelections={amountOfRowSelections}
+            onDelete={onDelete}
+            alerts={currentStateAlerts.slice(startIndex, endIndex)}
```
talboren, line 253: I think we're always passing all the alerts in the sliced view here so it deletes everything in the same view, although I only selected 1 alert.

talboren (changes requested):
2 comments, let's discuss this further over Slack! ❤️

keep-ui/app/alerts/alerts.tsx
asharonbaltazar: Should be resolved with https://github.com/keephq/keep/pull/657/commits/2b6fbde2df1e14f23f30be9e05166a8902fc30b8. That was a brain fart on my end 🤯

keep-ui/app/alerts/alert-menu.tsx
asharonbaltazar: Possibly resolved with https://github.com/keephq/keep/pull/657/commits/1fb3a84e30ff05642cb692a5f424627a34b75030. Lemme know what you think of the implementation. It's quite verbose, but the problem cannot (based on this [SO answer](https://stackoverflow.com/a/65535922/12627235)) addressed without dynamic positioning. 

Regarding the `@floatinui/react-dom` package, it's not installed as a dependency, but it's used in one of the other dependencies. I'm not sure – do we need to install it as a dependency?

keep-ui/app/alerts/alert-menu.tsx
```diff
@@ -137,146 +139,167 @@ export default function AlertMenu({
...
+                  enter="transition ease-out duration-100"
+                  enterFrom="transform opacity-0 scale-95"
+                  enterTo="transform opacity-100 scale-100"
+                  leave="transition ease-in duration-75"
+                  leaveFrom="transform opacity-100 scale-100"
+                  leaveTo="transform opacity-0 scale-95"
+                >
+                  <Menu.Items
+                    ref={refs.setFloating}
+                    className="z-50 fixed mt-2 divide-y divide-gray-100 rounded-md bg-white shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none"
```
talboren, line 171:
```suggestion
                    className="z-50 absolute mt-2 divide-y divide-gray-100 rounded-md bg-white shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none"
```

talboren (approved):
LGTM!

PR #698, opened 2024-01-08
chore: improve get-alerts efficiency

talboren:


keep-ui/app/alerts/alert-history-charts.tsx
```diff
@@ -36,9 +36,7 @@ export default function AlertHistoryCharts({
     timeUnit = "Hours";
   }
 
-  const alertsReversed = [...alerts].reverse();
-
-  const rawChartData = alertsReversed.reduce((prev, curr) => {
+  const rawChartData = [...alerts].reverse().reduce((prev, curr) => {
```
asharonbaltazar, line 39: If you use [`toReversed()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed), you don't need to copy arrays.

keep-ui/utils/fatigue.ts
```diff
@@ -18,7 +18,7 @@ export const calculateFatigue = (
   }
 
   // Sort alerts by timestamp
-  const sortedAlerts = alerts.sort(
+  const sortedAlerts = [...alerts].sort(
```
asharonbaltazar, line 21: If you don't want to copy arrays when sorting, you can use [`toSorted()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted)

talboren:
@asharonbaltazar quite strangely toSorted and toReversed didn't work.

> While Visual Studio Code might suggest its existence possibly due to an extension or a typing file, it's important to note that toSorted is not a standard method in JavaScript or TypeScript for array manipulation.

so I reverted the changes back.

PR #706, opened 2024-01-12
feat: Ability to set labels or custom fields like Component/s when creating Jira ticket

### LINKED ISSUE #699: Ability to set labels or custom fields like Component/s when creating Jira ticket
When I create a Jira ticket I would like to be able to set labels or custom fields so that I am able to use those fields for filtering in Jira.

I would like to set the label, component and fill in a custom field.

See:
https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issues/#api-rest-api-3-issue-post
https://support.atlassian.com/jira-cloud-administration/docs/configure-issue-custom-fields/

Min2who:
closes #699 
## 📑 Description

Ability to set labels or custom fields like Component/s when creating Jira ticket

## ✅ Checks

### DIFF
--- a/keep/providers/jira_provider/jira_provider.py
+++ b/keep/providers/jira_provider/jira_provider.py
@@ -255,7 +255,7 @@ class JiraProvider(BaseProvider):
         issue_type: str = "",
         labels: List[str] = None,
         components: List[str] = None,
-        custom_field_value: str = "",
+        custom_fields: dict = None,
         **kwargs: dict,
     ):
         """
@@ -284,7 +284,7 @@ class JiraProvider(BaseProvider):
                 fields["components"] = [{"name": component} for component in components]
 
             if custom_field_value:
-                fields["customfield_10021"] = custom_field_value
+                fields.update(custom_fields)
 
             request_body = {"fields": fields}
 
--- a/keep/providers/jira_provider/jira_provider.py
+++ b/keep/providers/jira_provider/jira_provider.py
@@ -253,9 +253,6 @@ class JiraProvider(BaseProvider):
         summary: str,
         description: str = "",
         issue_type: str = "",
-        labels: List[str] = None,
-        components: List[str] = None,
-        custom_fields: str = "",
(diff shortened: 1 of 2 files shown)


talboren (changes requested):
@Min2who great job! I didn't get a chance to test it yet but seems like a solid solution.
while components and labels seems OK, I didn't really get the `customfield_10027` part. I think we should be able to support X custom fields (as every company is different and has it's own custom fields).
Perhaps have that function get a `custom_fields: dict[str, str]` where the key is the custom_field id and the value is the actual value to put inside? (just a though).

Let me know your thoughts.

Min2who:
> @Min2who great job! I didn't get a chance to test it yet but seems like a solid solution. while components and labels seems OK, I didn't really get the `customfield_10027` part. I think we should be able to support X custom fields (as every company is different and has it's own custom fields). Perhaps have that function get a `custom_fields: dict[str, str]` where the key is the custom_field id and the value is the actual value to put inside? (just a though).
> 
actually i was too confused about that part
> Let me know your thoughts.
looks fine to me working on it

talboren:
> > @Min2who great job! I didn't get a chance to test it yet but seems like a solid solution. while components and labels seems OK, I didn't really get the `customfield_10027` part. I think we should be able to support X custom fields (as every company is different and has it's own custom fields). Perhaps have that function get a `custom_fields: dict[str, str]` where the key is the custom_field id and the value is the actual value to put inside? (just a though).
> 
> actually i was too hesitant about that part
> 
> > Let me know your thoughts.
> 
> looks fine to me working on it

Looking forward

talboren (approved):
lgtm!

talboren:
@Min2who seems like tests are failing. Do you mind taking a look and fixing? we can push that in once they pass.

Min2who:
> @Min2who seems like tests are failing. Do you mind taking a look and fixing? we can push that in once they pass.

done @talboren

Issue #731, opened 2024-01-18
Simplify `deleted` and `assignee` attributes in AlertDto


talboren:
TBD

talboren:
@asharonbaltazar FYI

PR #733, opened 2024-01-19
chore(ui): Move to hooks and Alerts Table Re-render problem

### LINKED ISSUE #719: Alerts table re-renders whenever a new alert is pushed
**Describe the bug**
Whenever Keep gets a new alert (via push/pull), the whole alerts table re-renders.
If the menu is open -> it will close
If the history modal is open -> it will close
If the extra payload is open -> it will close

It makes it quite hard to work with a specific alert when you're viewing the alerts table.

**To Reproduce**
1. Go to Alerts page
2. Open "Extra Payload" or "History" for example
3. Open a new tab
4. Go to Keep
5. Push an alert via webhook (CMD/CTRL+K -> webhook -> "Click to create example alert")
CleanShot 2024-01-16 at 18 22 23
6. See extra payload getting minimized in the first tab

**Expected behavior**
Alert table shouldn't be re-rendered but only the updated (or newly inserted alert) should be updated. (fingerprint is our unique identifier for a row)

**Screenshots**

https://github.com/keephq/keep/assets/68807791/20ed4a6a-7824-4119-b4ca-42f1b1e3cdd0

asharonbaltazar:
Closes #719 

## 📑 Description

- move `useSWR` calls into hook abstractions
- move from prop drilling by using the `use[DataType]` hooks
- lift state (alert history, pagination, column state) to avoid `AlertTable` re-renders 

## ℹ Additional Information

~Still lots of work to be done:~
- ~CRUD operations in the frontend have been remove and are blocked by #731~ 
- ~rendering issues still remain. Most of the work has been done to simplify logic for memoization~


asharonbaltazar:
@talboren I've left some todos in the code. Can you please make sure that functionality works? To refetch data, please use the `mutate` function returned by the appropriate `use[DataType]` hook

talboren:
> @talboren I've left some todos in the code. Can you please make sure that functionality works? Since we're moving the CRUD op to the backend, refetching data uses the `mutate` function returned by the appropriate `use[DataType]` hook

👀

PR #739, opened 2024-01-21
feat: chart with ingress

nmbarel:
Closes # 

## 📑 Description

add openshift route to the chart (controlled by values)

## ✅ Checks


talboren (approved):
LGTM

chart/keep/values.yaml
```diff
@@ -52,6 +52,12 @@ backend:
           - path: /
             pathType: ImplementationSpecific
     tls: []
+  route:
+    enabled: false
+    host: chart-example-backend.local
```
talboren, line 57: what's that exactly? why is it chart-example-backend and not keep-backend?

chart/keep/values.yaml
```diff
@@ -115,6 +121,12 @@ frontend:
           - path: /
             pathType: ImplementationSpecific
     tls: []
+  route:
+    enabled: false
+    host: chart-example.local
```
talboren, line 126: > what's that exactly? why is it chart-example-backend and not keep-backend?

same

chart/keep/values.yaml
nmbarel: taken from your ingress host value as to not deviate.

chart/keep/values.yaml
nmbarel: same as above reply

talboren (approved):
LGTM!

PR #754, opened 2024-01-29
fix: flashing alerts from pusher

asharonbaltazar:
## 📑 Description

- change `sorted` to `toSorted` to maintain data immutability
- spread existing alerts in the new array

### DIFF
--- a/keep-ui/utils/hooks/useAlerts.ts
+++ b/keep-ui/utils/hooks/useAlerts.ts
@@ -170,11 +170,20 @@ export const useAlerts = () => {
             new TextDecoder().decode(decompressedAlert)
           );
 
-          next(null, {
-            alerts: newAlerts,
-            lastSubscribedDate: new Date(),
-            isAsyncLoading: false,
-            pusherChannel,
+          next(null, (data) => {
+            if (data) {
+              return {
+                ...data,
+                alerts: [...newAlerts, ...data.alerts],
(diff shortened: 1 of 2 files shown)


keep-ui/utils/hooks/useAlerts.ts
```diff
@@ -170,11 +170,20 @@ export const useAlerts = () => {
...
+              return {
+                ...data,
+                alerts: [...newAlerts, ...data.alerts],
+              };
+            }
+
+            return {
+              alerts: newAlerts,
+              lastSubscribedDate: new Date(),
+              isAsyncLoading: false,
```
asharonbaltazar, line 184: If this event fires before the `async-done` event (and it always does, I believe), should we return `isAsyncLoading` as `false` or `true`? Since it's instantaneous, we won't be able to show the artificial loading

talboren: Can we not emit it? This can be happening in 2 cases:
1. When the user is sending get /alerts (and then we asyncly fetch more alerts from providers until `async-done` is received)
2. When new alerts gets pushed "on the fly"

talboren (approved):
LGTM!

talboren:
CleanShot 2024-01-29 at 14 06 56
@asharonbaltazar this seems to be a problem :(

PR #756, opened 2024-01-29
fix: Later Alerts Didn't Overwrite Earlier Ones

asharonbaltazar:
## 📑 Description

- use `Map` to use only the last alerts
- change spread order


talboren (approved):
LGTM!

PR #766, opened 2024-01-31
chore(api): refactor user endpoints

nmbarel:
Closes # 

## 📑 Description

refactor users endpoint to sit on /users instead of /settings, in order to fit with all conventions on other configs of keep (providers, workflows, etc)

## ✅ Checks


keep/api/api.py
```diff
@@ -159,25 +160,25 @@ def get_app(
...
-            if not user:
-                return JSONResponse(
+            if not user:                return JSONResponse(
                     status_code=401,
                     content={"message": "Invalid username or password"},
                 )
             # generate a JWT secret
             jwt_secret = os.environ.get("KEEP_JWT_SECRET")
             if not jwt_secret:
+                logger.info("invalid jwt token")
```
talboren, line 181: probably better:
```suggestion
                logger.info("missing KEEP_JWT_SECRET environment variable")
```

keep/api/api.py
```diff
@@ -159,25 +160,25 @@ def get_app(
...
-
+    if AUTH_TYPE == "SINGLE_TENANT":
         @app.post("/signin")
         def signin(body: dict):
             # validate the user/password
             user = get_user(body.get("username"), body.get("password"))
 
-            if not user:
-                return JSONResponse(
+            if not user:                return JSONResponse(
```
talboren, line 174: This looks like a bug

keep/api/api.py
```diff
@@ -159,25 +160,25 @@ def get_app(
...
     app.include_router(preset.router, prefix="/preset", tags=["preset"])
     app.include_router(groups.router, prefix="/groups", tags=["groups"])
+    app.include_router(users.router, prefix="/users", tags=["users"])
 
     # if its single tenant with authentication, add signin endpoint
     logger.info(f"Starting Keep with authentication type: {AUTH_TYPE}")
     # If we run Keep with SINGLE_TENANT auth type, we want to add the signin endpoint
-    if AUTH_TYPE == AuthenticationType.SINGLE_TENANT.value:
-
+    if AUTH_TYPE == "SINGLE_TENANT":
```
talboren, line 168: Why not use the enum?

talboren (changes requested):
Left 3 comments and also tests seems to be broken (probably due to the change in endpoints)

keep/api/api.py
nmbarel: was debugging the missing jwt and tried to see if it was due to enum failure... (before i remembered to add the env)
forgot to restore...

keep/api/api.py
nmbarel: yep

talboren (approved):
LGTM, @shahargl?

PR #795, opened 2024-02-06
fix(ui): new cols are always hidden

asharonbaltazar:
- address bug which showed new user-defined cols by default
- set `columnOrder` when selecting new cols
- col order and col visibility are now constants. they must be set declaratively.
- address bug which prevented users from moving user-defined properties


keep-ui/app/alerts/alert-table-utils.tsx
```diff
@@ -17,6 +17,23 @@ import AlertAssignee from "./alert-assignee";
...
+  "severity",
+  "name",
+  "description",
+  "status",
+  "lastReceived",
+  "source",
+  "assignee",
+  "extraPayload",
+  "alertMenu",
+];
```
asharonbaltazar, line 31: Please check whether these are all the _default_ (read: static) columns. I know there are user-defined columns, but those are generated automatically inside `useAlertsTableCols`

talboren (commented):
LGTM! thanks for this @asharonbaltazar

PR #798, opened 2024-02-06
fix: try to solve high cpu consumption when receiving tons of alerts …

talboren:


keep-ui/app/alerts/alert-assignee.tsx
```diff
@@ -8,7 +8,7 @@ interface Props {
 
 export default function AlertAssignee({ assignee }: Props) {
   const [imageError, setImageError] = useState(false);
-  const { data: users = [] } = useUsers({ revalidateOnFocus: false });
+  const { data: users = [] } = useUsers();
```
shahargl, line 11: why this?

PR #802, opened 2024-02-07
fix(ui): Column Width & History Modal

### LINKED ISSUE #690: Name column does not scale when columns are removed
**Describe the bug**
columns do not scale up properly when other columns are removed.
A lot of whitespace is left
**To Reproduce**
remove all columns except name (for example)

**Expected behavior**
columns scale to fill the whole widget

asharonbaltazar:
closes #690 

- fix skeleton shifting the table layout. the amount of skeleton cells sometimes superseded the visible columns, and when loading was finished, the whole table would shift as it removed the skeleton cells
- remove set width from the `name` col
- move menu to _right_ (typo in the commit 😬)
- history modal should stay open whenever a fingerprint alert is added. changed to use `fingerprint` as an identifier. the problem was that I used to select the alerts to view their history by their ID. but whenever a new fingerprint alert was pushed, `getFormatAndMergePusherWithEndpointAlerts` replaced the last alert with the most current. so the ID essentially changed in the `alerts` and thus the history modal closed since the ID in params was `undefined`. 
- removed the `loading` component since it was disruptive whenever a new fingerprint alert was pushed. perhaps it's worth exploring showing some loading screen _only_ while waiting for long responses.

### DIFF
--- a/keep-ui/app/alerts/alert-history.tsx
+++ b/keep-ui/app/alerts/alert-history.tsx
@@ -9,6 +9,7 @@ import { useAlerts } from "utils/hooks/useAlerts";
 import { PaginationState } from "@tanstack/react-table";
 import { useRouter, useSearchParams } from "next/navigation";
 import { toDateObjectWithFallback } from "utils/helpers";
+import Image from "next/image";
 
 interface AlertHistoryPanelProps {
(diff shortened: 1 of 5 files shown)


talboren (approved):
LGTM! thanks! 🧡

PR #806, opened 2024-02-08
fix(ui): joined state for alerts based on api and pusher

talboren:


asharonbaltazar (approved):
Maybe it makes sense to remove `base64` as well?

talboren:
> base64

Isn't that built in :O

PR #825, opened 2024-02-12
refactor(ui): Navigation Bar

asharonbaltazar:
- redesign the layout. move everything into a two column grid. left side is the nav, right side the content
- clean up logic for posthog
- create nav with a brand new search experience and menus
- adjust `/alerts` for new pathname logic
- fix infinite state setting bug in `useAllAlerts`
- cleanup `package.json`. will do this occasionally. removed an unused `.css` file


talboren:
> redesign the layout. move everything into a two column grid. left side is the nav, right side the content

@asharonbaltazar I feel like this is going to be a problem (correct me if I am wrong): https://www.figma.com/file/6RR7SHKOJZVlMzv4iJ1C3S/Product?type=design&node-id=29-9577&mode=design&t=oXuOn6frBUP0U7ME-0

keep-ui/components/navbar/ConfigureLinks.tsx
```diff
@@ -0,0 +1,41 @@
...
+          </LinkWithIcon>
+        </li>
+        <li>
+          <LinkWithIcon href="/rules" icon={MdOutlineEngineering}>
+            Alert Groups
+          </LinkWithIcon>
+        </li>
+        <li>
+          <LinkWithIcon href="/workflows" icon={LuWorkflow}>
+            Worksflows
```
talboren, line 35:
```suggestion
            Workflows
```

keep-ui/app/alerts/alerts.tsx
```diff
@@ -16,12 +15,16 @@ import { AlertMethodModal } from "./alert-method-modal";
 import AlertRunWorkflowModal from "./alert-run-workflow-modal";
 
 const defaultPresets: Preset[] = [
-  { name: "Feed", options: [] },
-  { name: "Deleted", options: [] },
-  { name: "Groups", options: [] },
+  { name: "feed", options: [] },
+  { name: "deleted", options: [] },
+  { name: "groups", options: [] },
 ];
```
talboren, line 21: Perhaps we should think about moving this to the "usePreset" hook?

keep-ui/app/layout.tsx
```diff
@@ -14,26 +14,27 @@ const mulish = Mulish({
...
-    <html
-      lang="en"
-      className={`h-full bg-gray-50 ${mulish.className}`}
-      suppressHydrationWarning={true}
-    >
-      <body className="h-full">
+    <html lang="en" className={`bg-gray-50 ${mulish.className}`}>
+      <body className="h-screen grid grid-cols-[250px_auto]">
         <NextAuthProvider>
+          {/* @ts-expect-error Server Component */}
```
talboren, line 28: can you leave a comment on why do we need this?

talboren (changes requested):
Left a few small comments but this looks AMAZING! I love it.

talboren:
We do also need to take care of small screens (navbar + provider tiles)
CleanShot 2024-02-18 at 10 32 41

talboren:
Just throwing in some more thoughts in here:
1. I'd add the preset name to "Alert - {presetName}" (because when moving between different presets, if not filters is configured, you can't really know what changed...)
2. I'd move the "Live" badge up, it also doesn't "change" between presets, it's something more general in the alerts page.
3. I'd remove the subtitle (and perhaps make a "generic" way to decide whether we want a subtitle in a page or not)
4. I'd move the "Alerts" up, and add a more "generic" way to decide if we want a title for some page -- if we do, i'd "continue" the divider from the keep logo to the end of the page

Let me know if that makes sense or I'm talking rubbish 🤣

talboren:
I don't know if it's a problem or not but this keeps on sending errors in the server-side console:
`⨯ Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.
    at useLocalStorage (./utils/hooks/useLocalStorage.ts:25:90)
    at DarkModeToggle (./app/dark-mode-toggle.tsx:18:113)`

asharonbaltazar:
> Just throwing in some more thoughts in here:
> ...

I think you're right with all of these suggestions. We do have the Figma detailing the Alerts page, which addresses all of your complaints, so I think the right move is to implement that mock

keep-ui/app/alerts/alerts.tsx
asharonbaltazar: I want to make a constants variable in alerts that'll be shared between the components and pages. I was planning on doing it with the `/alerts/` refactor

keep-ui/app/layout.tsx
asharonbaltazar: Resolved with https://github.com/keephq/keep/pull/825/commits/32793c29f6c58ae7037c0b9bf99ce786e1a0222a
Apparently, either the React or NextJS team pushed an update to allow JSX promises (for React Server Components)

asharonbaltazar:
> I don't know if it's a problem or not but this keeps on sending errors in the server-side 
> ... 

That's the DarkModeToggle acting up. I've copy/pasted your logic here, but I'm planning on refactoring the behavior. It should remove that error as well as improve load times

talboren:
CleanShot 2024-02-19 at 10 46 11
There seems to be some problem with the build, besides that, LGTM.

keep-ui/app/layout.tsx
```diff
@@ -14,26 +14,26 @@ const mulish = Mulish({
 });
 
 import { ToastContainer } from "react-toastify";
-import Navbar from "./navbar";
+import Navbar from "components/navbar/Navbar";
```
talboren, line 17:
```suggestion
import Navbar from "components/navbar/navbar";
```

talboren:
CleanShot 2024-02-19 at 11 59 55
Now it seems to be smth else.

talboren (approved):
LGTM

PR #860, opened 2024-02-26
chore: control pool size from environment variable

shahargl:


keep/api/alert_deduplicator/alert_deduplicator.py
```diff
@@ -38,6 +38,9 @@ def is_deduplicated(self, alert: AlertDto) -> bool:
         alert_deduplicate = (
             True if get_alert_by_hash(self.tenant_id, alert_hash) else False
         )
+        if alert_deduplicate:
+            self.logger.info(f"Alert {alert.id} is deduplicated {alert.source}")
```
talboren, line 42:
```suggestion
            self.logger.debug(f"Alert {alert.id} is deduplicated {alert.source}")
```

talboren (approved):
LGTM! I think you can control the pool size from the connection string but I don't know what's best practice (e.g., https://stackoverflow.com/questions/7784100/should-i-set-max-pool-size-in-database-connection-string-what-happens-if-i-don)

keep/api/alert_deduplicator/alert_deduplicator.py
shahargl: I've put "info" on purpose since this is the way to count how many deduplication are there, the long term should be add some other metric and then move it to debug

PR #870, opened 2024-02-29
feat(ui): Native Dark Mode

asharonbaltazar:
- clean `package.json`
- add `next-themes` package 
- adjust styling to adjust for dark theme


talboren:
There are still a few things that are not covered by this change (that were affected by the "non-native" dark mode).
Once we can fix them as well, I think this looks good -- overall, @asharonbaltazar, can you explain why this is better than the non-native mechanism?

CleanShot 2024-03-04 at 10 14 06
CleanShot 2024-03-04 at 10 14 26
CleanShot 2024-03-04 at 10 14 54
CleanShot 2024-03-04 at 10 15 31
CleanShot 2024-03-04 at 10 15 19

talboren (changes requested):
Left a small comment with some stuff that still require adjustment 🧡

PR #875, opened 2024-02-29
fix(helm): fix ingress backend service names

pehlicd:
## 📑 Description
This PR aims to fix ingress templates backend service names because in current ingress templates they are directed to non-existent services.


shahargl:
hey @pehlicd thanks for the PR!

can you share exactly what are you seeing? what's the fullName in each ingress and how adding `-backend` / `-frontend` in the templates solves that?

PR #884, opened 2024-03-03
feat(ui): Filtering by Date Range

### LINKED ISSUE #858: feat: filter by time
for columns that represent time, add some way to filter by timezone

asharonbaltazar:
closes #858 

- add `isDateWithinRange` function and add to alerts table init object
- add select range

https://github.com/keephq/keep/assets/58940073/c794ef84-85af-4612-bf10-21939a91826e

### DIFF
--- a/keep-ui/app/alerts/alert-columns-select.tsx
+++ b/keep-ui/app/alerts/alert-columns-select.tsx
@@ -1,9 +1,40 @@
 import { Table } from "@tanstack/table-core";
-import { Subtitle, MultiSelect, MultiSelectItem } from "@tremor/react";
+import {
+  Subtitle,
+  MultiSelect,
+  MultiSelectItem,
+  Select,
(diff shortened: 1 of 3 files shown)


talboren:
@asharonbaltazar you king! Dunno if it’s worth anything now but perhaps we should consider using this: https://www.tremor.so/docs/ui/date-range-picker

asharonbaltazar:
Shouldn't be a hard switch. But you lose the simplicity of a dropdown

talboren:
> Shouldn't be a hard switch. But you lose the simplicity of a dropdown


@asharonbaltazar 
they have that:
image

asharonbaltazar:
Sweet! I'll get to it

asharonbaltazar:
New changes:

Uploading Screen Recording 2024-03-03 at 3.33.36 PM.mov…

PR #886, opened 2024-03-04
feat(doc): add docs for mappings cli/api-ref

pehlicd:
## 📑 Description
updated mappings cli help messages
added docs cli commands for mappings command
added docs to api-ref for mapping


shahargl (approved):
lgtm

PR #902, opened 2024-03-06
feat(wip): aws incident manger

### LINKED ISSUE #804: New Provider - AWS Incident Manager
AWS incident manager is a native AWS Service that is very similar to pager duty. Every incident will have its own ARN(amazon resource name).

When an alert triggers, we can create an incident in the alert manager. While sending the resolved we need to run the API call to the incident manager and pass the ARN of that incident. So somewhere the Keep should maintain the track of ARN id so it can send the resolved status to the incident manager.

MayorChristopher:
Closes #804  

## ✅ Checks


MayorChristopher:
> this is empty implementation

I'm on it. My next push will have the full implementation, just need to make sure of somethings.

shahargl:
closing as its not getting progress and it is not the implementation we need (we need to get the alerts)

PR #921, opened 2024-03-07
feat(providers): add self hosted sentry support

### LINKED ISSUE #920: add self hosted sentry support
In sentry provider `SENTRY_API` variable set hard coded it, doesn't support self hosted sentry instances.

pehlicd:
Closes #920

## 📑 Description
 add self hosted sentry support

## ✅ Checks

### DIFF
--- a/keep/providers/sentry_provider/sentry_provider.py
+++ b/keep/providers/sentry_provider/sentry_provider.py
@@ -20,7 +20,14 @@ from keep.providers.providers_factory import ProvidersFactory
 @pydantic.dataclasses.dataclass
 class SentryProviderAuthConfig:
     """Sentry authentication configuration."""
-
+    api_url: str = dataclasses.field(
+        metadata={
+            "required": False,
+            "description": "Sentry API URL",
+            "hint": "(default: https://sentry.io/api/0) https://docs.sentry.io/api/",
+            "sensitive": False,
+        }
+    )
     api_key: str = dataclasses.field(
         metadata={
             "required": True,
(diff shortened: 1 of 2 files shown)


keep/providers/sentry_provider/sentry_provider.py
```diff
@@ -88,6 +95,7 @@ def __init__(
         super().__init__(context_manager, provider_id, config)
         self.sentry_org_slug = self.config.authentication.get("organization_slug")
         self.project_slug = self.config.authentication.get("project_slug")
+        self.SENTRY_API = self.config.authentication.get("api_url", self.SENTRY_API)
```
talboren, line 98: This is problematic as it will change API_URL constantly (for example, someone might change this to self hosted and override in general provider)
```suggestion
        self.SENTRY_API = self.config.authentication.get("api_url", self.SENTRY_API)
```

talboren (changes requested):
There’s some critical stuff to fix. I’ll review more in depth tomorrow.

keep/providers/sentry_provider/sentry_provider.py
pehlicd: So do you mean it can affect other connected sentry providers?

talboren: That’s right. `__init__` should declare self.sentry_api and adjust functions to use it.

talboren: Would probably rename the const to `SENTRY_MANAGED_API`

PR #942, opened 2024-03-11
chore(providers): add new field for squadcast provider

pehlicd:
## 📑 Description
added additional_json field to utilize it on squadcast

## ✅ Checks

### DIFF
--- a/keep/providers/squadcast_provider/squadcast_provider.py
+++ b/keep/providers/squadcast_provider/squadcast_provider.py
@@ -67,9 +67,6 @@ class SquadcastProvider(BaseProvider):
         """
         Validates that the user has the required scopes to use the provider.
         """
-        return {
-            "authenticated": True,
-        }
         refresh_headers = {
             "content-type": "application/json",
             "X-Refresh-Token": f"{self.authentication_config.refresh_token}",
@@ -120,6 +117,7 @@ class SquadcastProvider(BaseProvider):
         priority: str = "",
         status: str = "",
         event_id: str = "",
+        additional_json: str = "",
     ):
         body = json.dumps(
             {
@@ -131,6 +129,20 @@ class SquadcastProvider(BaseProvider):
             }
         )
 
+        if additional_json != "":
+            additional_json = additional_json.replace("'", "\"")
+            j = eval(additional_json)
+            # assign body to the json object to prevent core field from being overwritten
+            j["message"] = message
+            j["description"] = description
+            j["priority"] = priority
+            j["status"] = status
+            j["event_id"] = event_id
+            try:
+                body = json.dumps(j)
+            except Exception:
+                pass
+        
         return requests.post(
             self.authentication_config.webhook_url, data=body, headers=headers
         )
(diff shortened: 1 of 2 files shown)


keep/providers/squadcast_provider/squadcast_provider.py
```diff
@@ -131,6 +129,20 @@ def _create_incidents(
             }
         )
 
+        if additional_json != "":
+            additional_json = additional_json.replace("'", "\"")
+            j = eval(additional_json)
```
talboren, line 134: eval is quite risky and could allow remote code execution
```suggestion
            j = json.loads(additional_json)
```

talboren (changes requested):
1 critical comment

keep/providers/squadcast_provider/squadcast_provider.py
shahargl: I told him to do that since the `additional_json` is not `json` but string representation of python dict. we should investigate why that's the case and then fix it here @talboren

shahargl: that's what he gets inside `additional_json` (which raise JSONDecode error because `False` and `None` are not json but python)
```
{"id": "high_cpu_usage", "name": "high_cpu_usage", "status": "resolved", "severity": "critical", "lastReceived": "2024-03-10T18:57:28.107592+00:00", "environment": "unknown", "isDuplicate": False, "duplicateReason": None, "service": "sdlc-monitoring", "source": ["prometheus"], "apiKeyRef": None, "message": None, "description": "CPU usage is over 90%", "pushed": True, "event_id": None, "url": None, "labels": {"instance": "instance3", "job": "example2", "workfload": "somecoolworkload", "severity": "critical", "host": "host3", "alertname": "high_cpu_usage"}, "fingerprint": "d99cc6f50856dded98fd3caaabfb74d6", "deleted": False, "dismissUntil": None, "dismissed": False, "assignee": None, "providerId": None, "group": False, "note": None, "summary": "CPU usage is over 90%", "workfload": "somecoolworkload", "startsAt": "2024-03-10T18:57:28.064792+00:00", "alertname": "high_cpu_usage", "cluster": "sdlc", "job": "example2", "payload": {"summary": "CPU usage is over 90%", "startsAt": "2024-03-10T18:57:28.064792+00:00", "endsAt": "0001-01-01T00:00:00Z", "generatorURL": "http://example.com/graph?g0.expr=high_cpu_usage"}, "endsAt": "0001-01-01T00:00:00Z", "annotations": {"summary": "CPU usage is over 90%"}, "instance": "instance3", "alert_hash": "d36b066a05247b2453192d4abfce26824d76d1ad66587846f9373043c75434af", "generatorURL": "http://example.com/graph?g0.expr=high_cpu_usage", "host": "host3", "time": 1710148297.6059468}
```

talboren: @shahargl this means that `Squadcast` provider would have to be banned from being use in the managed platform.

Another suggestion I may have (and I tested it already), as a mid-term solution for this situation could be adding a new Keep function `to_valid_json` which will just:
`return unvalid_json.replace("'", '"').replace('None', '""').replace('False', 'false').replace('True', 'true')`
and then you can use
`json.loads(additional_json)`
it's probably not 100% but it'll work for your case @pehlicd 

Let me know if I should go further and push that in

pehlicd: That would definitely work for that case but it is hard foresee other cases, like alerts coming from sentry, cloudwatch etc. instead of replacing things one by one I would prefer to continue with the eval and fix that issue completely

talboren (approved):
LGTM!

PR #944, opened 2024-03-11
feat: Simplistic AI provider to enrich alerts with severity & service name (WIP)

Matvey-Kuk:
Two simplistic providers to add severity and service name based on free-form text input and ollama.

WIP

Docs are TBD.

By default disabled not to confuse cloud users, require:  "AI_PROVIDERS_ENABLED": "true" env vaiable.


keep/contextmanager/contextmanager.py
```diff
@@ -106,6 +107,9 @@ def get_full_context(
             "alert": self.event_context,  # this is an alias so workflows will be able to use alert.source
         }
 
+        if isinstance(full_context['alert'], dict):
+            full_context['alert'] = AlertDto(**full_context['alert'])
```
Matvey-Kuk, line 111: Wonder if that's a correct place to cast. Casting is required because sometimes it's an AlertDto, sometimes a dict (mainual/automatic triggering) which complicate provider code.

keep/providers/providers_factory.py
```diff
@@ -225,11 +225,16 @@ def get_all_providers() -> list[Provider]:
...
+            "base_ai_provider",
             "mock_provider",
             "file_provider",
             "github_workflows_provider",
         ]
 
+        if not os.environ.get("AI_PROVIDERS_ENABLED", False):
+            blacklisted_providers.append("ai_service_assignation_provider")
+            blacklisted_providers.append("ai_severity_provider")
+
```
Matvey-Kuk, line 237: What's the best place to document this?

keep/contextmanager/contextmanager.py
talboren: @shahargl fixed this bug a few hours ago~ (it should always be `AlertDto` now)

talboren:
Let's reopen it when it's relevant

PR #989, opened 2024-03-20
feat(ui): CEL Filter

### LINKED ISSUE #941: Presets filter should support CEL or CEL equivalent
**Current state**
Keep currently implement four different ways to filter alerts
1. On alert groups, we have CEL **(backend filtering)**
2. On preset filters, we have simple filtering (just ==) **(frontend filtering)**
3. On mapping filters, we have simple filtering **(backend filtering)**
4. On workflow's trigger we have simple AND + regex **(backend filtering)**

**Wanted state**
Keep we have one, unified, language to filter alerts. 

**Possible implementation**
Although CEL is the most generic and expandable, it doesn't have good react support so (2) will need a different treatment or to move to the backend somehow

asharonbaltazar:
closes #941


talboren (approved):
LGTM! build pass

PR #1001, opened 2024-03-23
feat: add option to edit mappings

### LINKED ISSUE #986: Add an option to update CSV in mapping
Today, when you create a mapping, to update the CSV (or any data source), you have to delete the mapping and re-upload it once again.
We should support "Edit" for existing mapping where you can replace the existing data and also edit matchers, etc.

35C4n0r:
Closes #986
/claim #986

## 📑 Description

- added `PUT` endpoint in backend for editing the mappings.
-  Added logic on the frontend for editing the mappings.

## ✅ Checks


35C4n0r:
@shahargl @talboren PR is ready for review.

keep-ui/app/mapping/create-new-mapping.tsx
```diff
@@ -16,31 +16,56 @@ import {
...
-import { ChangeEvent, FormEvent, useMemo, useState } from "react";
+import {ChangeEvent, FormEvent, useEffect, useMemo, useRef, useState} from "react";
 import { usePapaParse } from "react-papaparse";
 import { toast } from "react-toastify";
 import { getApiURL } from "utils/apiUrl";
 import { useMappings } from "utils/hooks/useMappingRules";
+import {MappingRule} from "./models";
 
-export default function CreateNewMapping() {
+export default function CreateNewMapping( {editRule, editCallback}: {editRule: MappingRule | null; editCallback: (rule: MappingRule | null) => void } ) {
```
talboren, line 26: Move to Props

keep-ui/app/mapping/create-new-mapping.tsx
```diff
@@ -16,31 +16,56 @@ import {
...
+    if (editRule !== null) {
+      handleFileReset();
+      setEditMode(true);
+      setMapName(editRule.name);
+      setFileName(editRule.file_name? editRule.file_name : "");
+      setMapDescription(editRule.description ? editRule.description : "");
+      setSelectedAttributes(editRule.attributes ? editRule.attributes : []);
+      setPriority(editRule.priority);
+    }
+  }, [editRule]);
```
talboren, line 47: Why is that needed? Add documentation

talboren (changes requested):
I feel like we maybe need to create a dedicated "EditMapping" component and simplify the code inside CreateNewMapping, wdyt?

keep-ui/app/mapping/create-new-mapping.tsx
35C4n0r: Once the user clicks the edit button on the right hand side table, the form on the left hand side gets populated with all the information, the above snippet populates those fields.

35C4n0r:
> I feel like we maybe need to create a dedicated "EditMapping" component and simplify the code inside CreateNewMapping, wdyt?

ig we can do that, I'll push the changes.

talboren:
> > I feel like we maybe need to create a dedicated "EditMapping" component and simplify the code inside CreateNewMapping, wdyt?
> 
> @talboren we can do that, we want it to work in the same way it does now right?

Basically I feel like this code complicates what used to be a rather "simple" component (`CreateNewMapping`).
What needs to be done is to get the "general" logic (basically the form itself) into some "general" component and then use it in both `CreateNewMapping` and `EditExistingMapping` or something (just a general thinking way).

LMK what you do you think

35C4n0r:
> > > I feel like we maybe need to create a dedicated "EditMapping" component and simplify the code inside CreateNewMapping, wdyt?
> > 
> > 
> > @talboren we can do that, we want it to work in the same way it does now right?
> 
> Basically I feel like this code complicates what used to be a rather "simple" component (`CreateNewMapping`). What needs to be done is to get the "general" logic (basically the form itself) into some "general" component and then use it in both `CreateNewMapping` and `EditExistingMapping` or something (just a general thinking way).
> 
> LMK what you do you think

@talboren I'll try to divide it into 3 components

35C4n0r:
Ok so this will be the `MappigForm` the common component

```
interface Props {
  mapName: string,
  setMapName: Dispatch<SetStateAction<string>>,
  mapDescription: string,
  setMapDescription: Dispatch<SetStateAction<string>>,
  priority: number,
  setPriority: Dispatch<SetStateAction<number>>
  selectedAttributes: string[],
  setSelectedAttributes: Dispatch<SetStateAction<string[]>>,
  handleSubmit: (e: FormEvent) => Promise<void>,
  attributes: string[],
  children: ReactNode,
}

export default function MappingForm(props: Props) {}
```

keep-ui/app/mapping/create-new-mapping.tsx
```diff
@@ -16,31 +16,56 @@ import {
...
-export default function CreateNewMapping() {
+export default function CreateNewMapping( {editRule, editCallback}: {editRule: MappingRule | null; editCallback: (rule: MappingRule | null) => void } ) {
   const { data: session } = useSession();
   const { mutate } = useMappings();
   const [mapName, setMapName] = useState<string>("");
   const [fileName, setFileName] = useState<string>("");
   const [mapDescription, setMapDescription] = useState<string>("");
   const [selectedAttributes, setSelectedAttributes] = useState<string[]>([]);
   const [priority, setPriority] = useState<number>(0);
+  const [editMode, setEditMode] = useState(false);
```
talboren, line 34: Can't you just
```suggestion
  const editMode = editRule !== null;
```
?

keep-ui/app/mapping/create-new-mapping.tsx
talboren, line 26:
```suggestion
interface Props {
editRule: MappingRule | null; 
editCallback: (rule: MappingRule | null) => void 
}

export default function CreateNewMapping( {editRule, editCallback}: Props ) {
```

keep/api/routes/mapping.py
```diff
@@ -79,3 +79,28 @@ def delete_rule(
...
+
+@router.put("", description="Update an existing rule")
+def update_rule(rule: MappingRuleDtoUpdate,
+                authenticated_entity: AuthenticatedEntity = Depends(AuthVerifier(["write:rules"])),
+                session: Session = Depends(get_session)) -> MappingRuleDtoOut:
+    logger.info("Updating a mapping rule")
+    existing_rule: MappingRule = session.query(MappingRule).filter(
+        MappingRule.tenant_id == authenticated_entity.tenant_id and MappingRule.id == rule.id).first()
+
+    existing_rule.name = rule.name
```
talboren, line 92: Check if existing_rule is not None, raise proper exception if it is

talboren (changes requested):
Please properly document and added a few notes

keep/api/routes/mapping.py
35C4n0r: I think if rule.name is somehow None, pydantic will throw a 422 UnprocessableIdentity Error.

35C4n0r: Oh got it. I misunderstood it with the edited name :)

35C4n0r:
@talboren, thanks for the review, updated as needed.

talboren (approved):
LGTM

PR #1002, opened 2024-03-23
feat: update database conn envs

pehlicd:
## 📑 Description
Right now while creating the db engine backend getting the DATABASE_CONNECTION_STRING env var and use it. It is really nice but when the time comes to deployment in helm chart I wouldn't want to declare this env variable inside of my values.yaml because I am keeping all of my helm charts in a gitops repo so it would cause a security problem if I declare my db connection string as clear text. So in order fix that I am proposing to declare

```
DATABASE_HOST
DATABASE_PORT
DATABASE_USER
DATABASE_PASSWORD
DATABASE_NAME 
```

env variables separately and combine them in the backend afterwards. This will let us to use secrets for DATABASE_PASSWORD and get this env from k8s secrets

## ✅ Checks

## ℹ Additional Information
I will also update helm chart as following.

### DIFF
--- a/docker-compose.common.yml
+++ b/docker-compose.common.yml
@@ -21,7 +21,11 @@ services:
       - PORT=8080
       - SECRET_MANAGER_TYPE=FILE
       - SECRET_MANAGER_DIRECTORY=/state
-      - DATABASE_CONNECTION_STRING=sqlite:///./db.sqlite3?check_same_thread=False
+      - DATABASE_HOST=""
+      - DATABASE_PORT="3306"
+      - DATABASE_USER="root"
+      - DATABASE_PASSWORD=""
+      - DATABASE_NAME="keep"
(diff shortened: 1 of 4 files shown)


shahargl:
Few comments:
1. We need backward compatibility - meaning if a user DID specify a database connection string, it should be used (maybe extract the other variables from it)
2. We need to support DB's other than MySQL

Why not just storing DATABASE_CONNECTION_STRING as a secret?

pehlicd:
> Few comments:
> 
> 1. We need backward compatibility - meaning if a user DID specify a database connection string, it should be used (maybe extract the other variables from it)
> 2. We need to support DB's other than MySQL
> 
> Why not just storing DATABASE_CONNECTION_STRING as a secret?

make sense I am closing this PR and update helm charts PR to make `DATABASE_CONNECTION_STRING` as secret

PR #1092, opened 2024-04-11
feat: added event_id to alertDTO to utilize as unique identifier in workflows and rest

pehlicd:
## 📑 Description
In current stack, there is nothing evaluate alerts with unique id, so I've updated `event_id` to utilize in workflows.

## ✅ Checks


keep/api/core/db.py
```diff
@@ -996,6 +996,7 @@ def get_last_alerts(tenant_id, provider_id=None, limit=1000) -> list[Alert]:
         alerts = []
         for alert, startedAt in alerts_with_start:
             alert.event["startedAt"] = str(startedAt)
+            alert.event["event_id"] = str(alert.id)
```
talboren, line 999: I'd suggest the following so we don't override something that may be provided by the provider itself
```suggestion
            alert.event["keep_event_id"] = str(alert.id)
```

pehlicd: done!

keep/api/routes/alerts.py
```diff
@@ -523,8 +523,10 @@ def handle_formatted_events(
                 alert_hash=formatted_event.alert_hash,
             )
             session.add(alert)
-            formatted_event.event_id = alert.id
-            alert_dto = AlertDto(**alert.event)
+            session.flush()
+            session.refresh(alert)
```
talboren, line 527: why not before?

talboren (approved):
LGTM

PR #1136, opened 2024-04-23
fix: fixed event_id logic and some typos

pehlicd:
## 📑 Description
We have added event_id field in event lately but we missed to update it on alert tables event column so every alert supposed to have event_id as the id of the alert on table but since we missed it, it was coming as null in as the value of the event_id field. This is problematic because the way we check throttling we need that unique event_id and while it is null we get the error below.

```
Failed to run action xxx: one of the hex, bytes, bytes_le, fields, or int arguments must be given
```

## ✅ Checks


talboren:
@pehlicd I'm closing this PR for now, let's discuss it over Slack.

PR #1189, opened 2024-05-21
feat(ui): Alert Correlation Rule Badges

asharonbaltazar:


keep-ui/app/rules/CorrelationSidebar/CorrelationSidebarBody.tsx
```diff
@@ -67,6 +73,7 @@ export const CorrelationSidebarBody = ({
             ruleName: name,
             celQuery: formatQuery(query, "cel"),
             timeframeInSeconds,
+            grouping_criteria: alertsFound.length ? groupedAttributes : [],
```
asharonbaltazar, line 76: `grouping_criteria` won't be sent to the backend if there are no alerts found. This ternary is a simple check to avoid clearing the GroupBy `MultiSelect` whenever `alertsFound.length === 0`, if that makes sense.

keep-ui/utils/hooks/useSearchAlerts.ts
```diff
@@ -0,0 +1,36 @@
...
+import { RuleGroupType, formatQuery } from "react-querybuilder";
+
+export const useSearchAlerts = (
+  args: { query: RuleGroupType; timeframe: number },
+  options?: SWRConfiguration
+) => {
+  const apiUrl = getApiURL();
+  const { data: session } = useSession();
+
+  const [debouncedArgs] = useDebouncedValue(args, 2000);
```
asharonbaltazar, line 16: Is a 2 seconds delay enough?

keep-ui/app/rules/CorrelationSidebar/CorrelationSidebarBody.tsx
talboren: It makes sense but looks like it sends that also when it's found.

keep-ui/utils/hooks/useSearchAlerts.ts
talboren: think so, let's adjust when needed if not

talboren (changes requested):
Shared some notes in Slack :) ❤️

PR #1208, opened 2024-06-05
fix: updated provider db model to make name unique

pehlicd:
## 📑 Description
In provider table name column is not unique and it is kind of problematic this PR aims to fix that problem.

## ✅ Checks

### DIFF
--- a/keep/api/models/db/provider.py
+++ b/keep/api/models/db/provider.py
@@ -7,7 +7,7 @@ from sqlmodel import JSON, Column, Field, SQLModel
 class Provider(SQLModel, table=True):
     id: str = Field(default=None, primary_key=True)
     tenant_id: str = Field(foreign_key="tenant.id")
-    name: str
+    name: str = Field(unique=True)
     description: Optional[str]
     type: str
     installed_by: str
--- a/keep/api/models/db/provider.py
+++ b/keep/api/models/db/provider.py
@@ -10,7 +10,7 @@ class Provider(SQLModel, table=True):
 
     id: str = Field(default=None, primary_key=True)
     tenant_id: str = Field(foreign_key="tenant.id")
-    name: str = Field(unique=True)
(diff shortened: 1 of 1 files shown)


talboren (approved):
LGTM!

keep/api/models/db/provider.py
```diff
@@ -7,7 +7,7 @@
 class Provider(SQLModel, table=True):
     id: str = Field(default=None, primary_key=True)
     tenant_id: str = Field(foreign_key="tenant.id")
-    name: str
+    name: str = Field(unique=True)
```
talboren, line 10: @pehlicd this could be problematic because we need it to be unique per tenant id.

Please see `db/preset.py` (__table_args__ = (UniqueConstraint("tenant_id", "name"),))

talboren (changes requested):
1 comment

keep/api/models/db/provider.py
pehlicd: @talboren I added table args do you think I should remove `Field(unique=True)` this one. In db/preset.py name is like that as well so I left this one

talboren: you can remove it :)

keep/api/models/db/provider.py
```diff
@@ -1,13 +1,16 @@
...
 from sqlmodel import JSON, Column, Field, SQLModel
 
 
 class Provider(SQLModel, table=True):
+    __table_args__ = (UniqueConstraint("tenant_id", "name"),)
+
     id: str = Field(default=None, primary_key=True)
     tenant_id: str = Field(foreign_key="tenant.id")
-    name: str
+    name: str = Field(unique=True)
```
talboren, line 13:
```suggestion
    name: str
```

talboren (approved):
LGTM!

PR #1227, opened 2024-06-11
[DO NOT MERGE] feat: enrich workflow (workflow state)

talboren:
WIP workflow state mechanism


keep/providers/slack_provider/slack_provider.py
```diff
@@ -104,6 +104,7 @@ def _notify(self, message="", blocks=[], channel="", **kwargs: dict):
         Args:
             kwargs (dict): The providers with context
         """
+        return
```
talboren, line 107: Needs to be removed (was added for testing)

PR #1132, opened 2024-04-20
fix: extraction create error

pehlicd:
## 📑 Description
While creating extraction rule it was throwing sql error since we were not supply `updated_at` and `updated_by` infos. This PR fixes it.

```json
{
    "message": "An internal server error occurred.",
    "trace_id": "x",
    "error_msg": "(sqlite3.OperationalError) no such column: updated_at\n[SQL: INSERT INTO extractionrule (tenant_id, priority, name, description, created_by, created_at, updated_by, updated_at, disabled, pre, condition, attribute, regex) VALUES (?, ?, ?, ?, ?, ?, ?, updated_at, ?, ?, ?, ?, ?) /*db_driver='pysqlite',db_framework='sqlalchemy%%3A0.41b0',traceparent='x'*/]\n[parameters: ('keep', 0, 'asd', 'asd', 'admin@keephq', '2024-04-20 12:30:39.005900', None, 0, 0, '', 'asd', '(?P<groupname>)')]\n(Background on this error at: https://sqlalche.me/e/14/e3q8)"
}
```

## ✅ Checks
- [x] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [x] All the tests have passed

PR #1284, opened 2024-06-25
fix: "no such column updated_at" during extraction creating

VladimirFilonov:
## 📑 Description

While creating extraction rule it was throwing SQL error:

```json
{
    "message": "An internal server error occurred.",
    "trace_id": "x",
    "error_msg": "(sqlite3.OperationalError) no such column: updated_at\n[SQL: INSERT INTO extractionrule (tenant_id, priority, name, description, created_by, created_at, updated_by, updated_at, disabled, pre, condition, attribute, regex) VALUES (?, ?, ?, ?, ?, ?, ?, updated_at, ?, ?, ?, ?, ?) /*db_driver='pysqlite',db_framework='sqlalchemy%%3A0.41b0',traceparent='x'*/]\n[parameters: ('keep', 0, 'asd', 'asd', 'admin@keephq', '2024-04-20 12:30:39.005900', None, 0, 0, '', 'asd', '(?P<groupname>)')]\n(Background on this error at: https://sqlalche.me/e/14/e3q8)"
}
```

The issue was already mentioned in PR https://github.com/keephq/keep/pull/1132, but the solution was just bypassing consequences.

The problem here was in using sa.Column instance directly, without wrapping it with Field. As a result, when no value for updated_at was provided during the SQL query string generation, the engine treated this Column as an expression. Like if we were doing INSERT INTO SELECT. 

`VALUES (?, ?, ?, ?, ?, ?, ?, updated_at, ?, ?, ?, ?, ?)` <- SQL engine is trying to find column `updated_at` in our INSERT statement (not in target table), and of course can't. That's why we have `no such column: updated_at` - this is not about the field in the table but the virtual field in the request.

## ✅ Checks

## ℹ Additional Information

1. I extracted some functions to a separate module to re-use test_app and client, created in test_auth.
2. The original goal of using Column for update_at was to make this field automatically updated. But I see some inconsistency here with [MappingRule.last_updated_at](https://github.com/keephq/keep/blob/main/keep/api/models/db/mapping.py#L29). Sould we align then to have the same behavior?

### DIFF
--- a/keep/api/models/db/extraction.py
+++ b/keep/api/models/db/extraction.py
@@ -2,6 +2,7 @@ from datetime import datetime, timezone
 from typing import Optional
 
 from pydantic import BaseModel
+from sqlalchemy import DateTime
 from sqlalchemy.sql import func
 from sqlmodel import Column, Field, SQLModel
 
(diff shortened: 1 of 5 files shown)


shahargl:
@VladimirFilonov ruff fails because of redundant import datetime (now when its not being used), can you remove it so e2e tests will run?

shahargl (changes requested):
see comment

VladimirFilonov:
@shahargl Done, please check

shahargl (approved):
lgtm

PR #1356, opened 2024-07-10
Call provider_class.format_alert in process_event only when an event is a dict

VladimirFilonov:
## 📑 Description

Fixes issue for some providers

```
Traceback (most recent call last):
  File "/app/keep/api/tasks/process_event_task.py", line 373, in process_event
    event = provider_class.format_alert(event, None)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/keep/providers/base/base_provider.py", line 263, in format_alert
    formatted_alert = cls._format_alert(event, provider_instance)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/keep/providers/grafana_provider/grafana_provider.py", line 199, in _format_alert
    alerts = event.get("alerts", [])
             ^^^^^^^^^
AttributeError: 'list' object has no attribute 'get'
```

## ✅ Checks

### DIFF
--- a/keep/api/tasks/process_event_task.py
+++ b/keep/api/tasks/process_event_task.py
@@ -368,16 +368,18 @@ def process_event(
         except Exception:
             logger.exception("Failed to run pre-formatting extraction rules")
 
-        if provider_type is not None:
+        if provider_type is not None and type(event) is dict:
             provider_class = ProvidersFactory.get_provider_class(provider_type)
             event = provider_class.format_alert(event, None)
+
+        # In case when provider_type is not set
+        if isinstance(event, dict):
+            event = [AlertDto(**event)]
 
         # Prepare the event for the digest
         if isinstance(event, AlertDto):
             event = [event]
 
-        if isinstance(event, dict):
-            event = [AlertDto(**event)]
 
         __internal_prepartion(event, fingerprint, api_key_name)
         __handle_formatted_events(


keep/api/tasks/process_event_task.py
```diff
@@ -368,16 +368,18 @@ def process_event(
         except Exception:
             logger.exception("Failed to run pre-formatting extraction rules")
 
-        if provider_type is not None:
+        if provider_type is not None and type(event) is dict:
```
talboren, line 371:
```suggestion
        if provider_type is not None and isinstance(event, dict):
```

keep/api/tasks/process_event_task.py
```diff
@@ -368,16 +368,18 @@ def process_event(
...
             provider_class = ProvidersFactory.get_provider_class(provider_type)
             event = provider_class.format_alert(event, None)
 
+        # In case when provider_type is not set
+        if isinstance(event, dict):
+            event = [AlertDto(**event)]
```
talboren, line 377: why did you move it?

talboren (dismissed):
1 small comment and lgtm.
still not sure what happens when it's a list[dict]

talboren:
> 1 small comment and lgtm. still not sure what happens when it's a list[dict]

I feel like the provider should handle it(?)

keep/api/tasks/process_event_task.py
VladimirFilonov: Do make it right after handling the dict with the provider. Looks better for me: Dict with provider -> dict without provider -> AlertDto...

VladimirFilonov:
> > 1 small comment and lgtm. still not sure what happens when it's a list[dict]
> 
> I feel like the provider should handle it(?)

Well, technically, we should not have list[dict]:

```
event: (
        AlertDto | list[AlertDto] | dict
    ), 
```

But if some providers can create such data, let handle this inside provider driver

Matvey-Kuk (approved):

Issue #1283, opened 2024-06-25
[🐛 Bug] Dashboard overflow is not taken into consideration


talboren:
**Describe the bug**
If you mistakenly move dashboard widget too much down, it disappears from screen

**To Reproduce**
1. See the attached video

**Expected behavior**
User should not be able to move widgets outside of the scope of the screen or the screen should grow accordingly.

**Screenshots**

https://github.com/keephq/keep/assets/68807791/2d0d9dc1-af45-40ac-ac43-1cbca07d0b0d

ArtArtem:
@talboren Hi, I want to try, I think I have an idea why this bug is happening

ArtArtem:
https://github.com/keephq/keep/assets/45171843/c9e75ccc-b8dc-4aa1-8844-1a91fd13619a

@talboren @Matvey-Kuk 

The problem is solved, I think it's better for UX to choose a scrollbar solution, because clients may need to place a large number of widgets of different shapes, so a height equal to the height of the screen may not be enough, I want to hear your thoughts

talboren:
@ArtArtem sorry for not replying earlier! Somehow I missed it. Looks awesome! Do you have a PR for it?

cc @shahargl

ArtArtem:
@talboren, sure https://github.com/keephq/keep/pull/1365

talboren:
@ArtArtem thank you for the contribution!

PR #1364, opened 2024-07-11
fix: mobile mode (#1283)

ArtArtem:
On large screens, there is a MinimizeMenuButton to hide and display the menu, and for small screens, a Popover.Button.

However, both buttons are displayed on small screens by mistake, and due to the change of grid to flex MinimizeMenuButton takes up too much space.

Problem is solved.


talboren (approved):
LGTM

PR #1365, opened 2024-07-11
fix: dashboard overflow (#1283)

ArtArtem:
If you mistakenly move dashboard widget too much down, it disappears from screen. 

Solved. The screen grows accordingly.


talboren (approved):
LGTM!

PR #1368, opened 2024-07-11
feat: Hello Migrations 👋

Matvey-Kuk:
Closing #980 
1) Alembic - driven
2) Autogenerator eats model schema 👍
3) It will auto-migrate on a server startup
4) It's tolerating environments which were created before the migrator was introduced.


talboren (approved):
LGTM!

Issue #1255, opened 2024-06-19
[🐛 Bug] `workflowexecution` result fails when result contains datetime


talboren:
**Describe the bug**
When the result of a workflow execution might include datetime object, we're unable to save it in to the database because of a serialization problem.

**To Reproduce**
1. TBD

**Expected behavior**
Workflow result should be saved in the DB properly.

**Screenshots**
![CleanShot 2024-06-19 at 12 04 33](https://github.com/keephq/keep/assets/68807791/2dd64160-18df-472e-8d35-4324c488462d)

**Additional context**
-

ArtArtem:
@talboren @Matvey-Kuk Hi, I probably found the cause of the bug, check it out https://github.com/keephq/keep/pull/1382

ArtArtem:
@talboren After further exploring the capabilities of SQLAlchemy, I found a more correct way to solve the problem, I'm sure it's better now
https://github.com/keephq/keep/pull/1392

Matvey-Kuk:
Labeling this one as a Good First Issue.

@ArtArtem already made a good progress with this issue, his PR is a good starting point to pick this up :)

If you want to work on this, join Slack and we will help: https://slack.keephq.dev/ !

talboren:
This should already been solved, closing as it's probably just obsolete

PR #1236, opened 2024-06-13
Feat: Support for mongodb without database and binary data storage in Keep's DB

HNitzanH:
Closes # 

## ✅ Checks

- [ ] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [ ] All the tests have passed

PR #1382, opened 2024-07-15
fix: converting workflow results datetime objects (#1255)

ArtArtem:
Describe the bug
When the result of a workflow execution might include datetime object, we're unable to save it in to the database because of a serialization problem.

Solution:
So now, according to mock testing data:
e.g.
mock_action1 = Mock(name="action1")
mock_action1.name = "action1"
mock_action 1.provider.results = {"result": "value1"}

results dict's fields will be iterated and datetime objects will be converted to be properly saved in DB

image

Needs to be double checked with relevant workflowexecution data

### DIFF
--- a/keep/api/core/db.py
+++ b/keep/api/core/db.py
@@ -1009,6 +1009,14 @@ def create_user(tenant_id, username, password, role):
         session.refresh(user)
     return user
 
+def serialize_workflow_results(results):
+    # Converting datetime objects to strings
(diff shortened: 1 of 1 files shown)


talboren:
Hey @ArtArtem 
Thank you for this contribution! Unfortunately, this doesn't seem to be the right way to solve this issue. Please see: https://github.com/keephq/keep/pull/1236 for more details.

PR #1392, opened 2024-07-17
fix: workflowexecution result fail 1255

ArtArtem:
Describe the bug
When the result of a workflow execution might include datetime object, we're unable to save it in to the database because of a serialization problem.

Solution:
Now, when you save data in the results field, the datetime objects will be automatically serialized to a JSON-compatible format, and when the data is retrieved, they will be deserialized back.


ArtArtem:
@talboren Please share your thoughts

talboren:
@ArtArtem i think it makes sense.
@shahargl wdyt?

shahargl:
LGTM, @ArtArtem are you up to add e2e test for this scenario?

ArtArtem:
@shahargl 
I have written a draft function test_save_workflow_results and I need your advice, the main thing we have to test is the _save_workflow_results function call.

You can use it directly, as I have in my example, or you can use _run_workflow, this option is used in test_handle_workflow_test, but it uses “mock_workflow_manager = Mock()” instead of workflow_manager, so no real processing is going on.

Maybe you can suggest your commit with the most correct approach?

shahargl:
Why not using the workflow manager without mocking it?

Issue #1391, opened 2024-07-17
[🐛 Bug]: if condition in workflow visual editor is broken


talboren:
When adding a condition to some step, it gets indented in to the provider section while it should be out

**not good:**
![CleanShot 2024-07-17 at 11 49 07@2x](https://github.com/user-attachments/assets/08f57b41-503f-4112-8027-161c85e6f8e0)

good:
![CleanShot 2024-07-17 at 11 50 08@2x](https://github.com/user-attachments/assets/48d2f72d-449c-48a2-b3dc-8faa63fdca61)

PR #1426, opened 2024-07-21
fix(ui): Workflow builder move the workflow 'if' condition provider nesting

shahargl:
close https://github.com/keephq/keep/issues/1391


talboren (changes requested):
don't know how backend will take
`"if": None`
so let's add it only when ifParam is not undefined

talboren (approved):
k

PR #1467, opened 2024-07-25
fix: error image fallback in trigger button

rajeshj11:
fixed the error image fallback issue


rajeshj11:
@talboren please review

PR #1473, opened 2024-07-25
fix: Add alerts_count, affected_services and sources fields to the Incident

### LINKED ISSUE #1472: [🐛 Bug]: Poor performance of Incidents list with large amount of alerts
**Describe the bug**
Data preparation for the incidents list page takes a lot of time when we have multiple incidents with thousands of alerts in them

**To Reproduce**
1. Create 5-10 incidents
2. Assign 5-10k alerts with each
3. Go to Incidents page

**Expected behavior**
The incidents list page should be loaded fast

VladimirFilonov:
Add alerts_count, affected_services and sources fields to the Incident and pre-population logic for them on adding alerts to the incident

Closes #1472

## 📑 Description

The incidents list page performed poorly when there were a large number of alerts in the incidents. That happened because alerts count, affected services and sources were calculated in runtime.

This PR changes this behavior to pre-calculated data for incidents

## ✅ Checks


keep/api/core/db.py
```diff
@@ -2044,6 +2043,42 @@ def get_incident_alerts_by_incident_id(tenant_id: str, incident_id: str) -> List
...
+
+    def inner(db_session: Session):
+        alerts_data = db_session.exec(
+            select(
+                func.json_extract(Alert.event, '$.service').label("service"),
+                func.json_extract(Alert.event, '$.source').label("source")
+            ).where(
+                col(Alert.id).in_(alert_ids),
+            )
+        ).all()
```
talboren, line 2059: This is a bit problematic because of different sql dialects (looks for `session.bind.dialect.name` across the code and you'll see)

keep/api/core/db.py
```diff
@@ -2044,6 +2043,42 @@ def get_incident_alerts_by_incident_id(tenant_id: str, incident_id: str) -> List
     return query.all()
 
 
+def get_alerts_data_for_incident(
```
talboren, line 2046: Is this performant now? can you add some documentation on what exactly is being done here? why is the inner function needed?

keep/api/core/db.py
```diff
@@ -2102,8 +2150,44 @@ def remove_alerts_to_incident_by_incident_id(
             )
             .delete()
         )
+        session.commit()
 
+        alerts_data_for_incident = get_alerts_data_for_incident(alert_ids, session)
+
+        source_json_each_alias = aliased(func.json_each(Alert.event, '$.source').table_valued("value"))
```
talboren, line 2157: again, I think this is dialect specific and might cause problems

talboren (changes requested):
I have a few comments, mostly around dialect specific SQL functions that could be problematic in the future (or even now).
Maybe the ideal case here is to have some worker doing those calcs in the background? just thinking out loud, idk.

talboren (approved):
LGTM!

PR #1486, opened 2024-07-29
fix: arq by adding an async_process_event

Matvey-Kuk:
Not sure if it's the best decision, but that's the only way I found to make Arq work :)


talboren (approved):
lgtm

PR #1514, opened 2024-08-01
fix: VS code quickstart

### LINKED ISSUE #1513: [🐛 Bug]: Fresh launch via VS code doesn't work
Following docs doesn't help, the issue is with .env file

IMAGE 2024-08-01 14:53:25

Matvey-Kuk:
closes #1513

### DIFF
--- a/keep-ui/.env.local.example
+++ b/keep-ui/.env.local.example
@@ -1,6 +1,6 @@
 NEXTAUTH_URL=http://localhost:3000
 
-# Requered:
+# Required:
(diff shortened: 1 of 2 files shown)


keep-ui/.env.local.example
```diff
@@ -1,5 +1,8 @@
 NEXTAUTH_URL=http://localhost:3000
-NEXTAUTH_SECRET= # Linux: `openssl rand -hex 32` or go to https://generate-secret.now.sh/32
+
+# Requered:
```
talboren, line 3:
```suggestion
# Required:
```

talboren (approved):
LGTM

PR #1526, opened 2024-08-03
fix: when preset is private it needs to filtered form the asset links

### LINKED ISSUE #1163: Private presets are being pushed as part of `async-presets` message
Even if the preset is private, it's still being added to users who doesn't own the preset

rajeshj11:
/claim #1163 

Closes #1163

## 📑 Description

If there is a private preset. we will check preset-created_by. if it matches the current session user then we will allow the preset on the UI or else we will ignore it.

## ✅ Checks


rajeshj11:
@talboren please review

talboren (approved):
LGTM!

PR #1559, opened 2024-08-06
feat: creating incidents from the alert feed

### LINKED ISSUE #1540: [➕ Feature]: Creating incidents from the alert Feed
Use-case: As a user, I noticed that some alerts in the feed should be associated with a new incident. I choose those alerts, click "+Associate with an Incident," and can't assign them to the newly created incident; I can only assign them to an existing one. Even if I create a new incident in a new tab, the list of incidents in the popup won't update. If I refresh the Feed, I lose my page 😅 (covered here https://github.com/keephq/keep/issues/1541)

1) Make it possible to create a new incident from the popup
2) Make the list of incidents dynamic, so if an incident was created in the background, I should be able to choose it without updating a page.

Screenshot 2024-08-05 at 13 13 42

sergtl:
Closes #1540 
/claim #1540 

## 📑 Description
This PR introduces two new features:

1. You can now create a new incident directly from the "Associate with an incident" popup. Simply click on "Add a new incident," and the incident creation form will appear without navigating away from the current page.
2. The list of incidents is now dynamic. If a user creates new incidents in another tab, they will automatically appear in both the popup list and the sidebar navigation.

## ✅ Checks

## ℹ Additional Information

https://www.loom.com/share/1d29c1258c644781b1d9349e9fe2a920


Matvey-Kuk (changes requested):
I tested the PR, and noticed that:
1) Alerts are not getting assigned to the freshly created incident.
2) I think there should be a link to the new incident available so I will be able to jump there.
3) Incident list is not getting updated.

sergtl:
Thanks @Matvey-Kuk 

Just retested:

1. Seems to be working
2. Sure, I'll add a link to newly created incident
3. Also seems to be working as expected

Matvey-Kuk:
@sergtl re-tested, my bad about 3.

About 1, I checked your video, alerts are not getting attached to the freshly created incident, you make an additional click "Associate alert" before you close the popup.

sergtl:
Got it! I'll automatically associate the new incident and close the popup. In this case we pass num 2?

Matvey-Kuk:
@sergtl yep!

sergtl:
@Matvey-Kuk Done

Matvey-Kuk:
Works like a charm! I didn't find the link to the incident, but merging. Thank you for the contribution!

PR #1560, opened 2024-08-06
fix: core/db - remove unused function. closes #1561

### LINKED ISSUE #1561: Two definitions of get_incident_by_id in the same file
**Describe the bug**
In the file https://github.com/keephq/keep/blob/main/keep/api/core/db.py we have 2 definitions of the same function:
```
def get_incident_by_id(incident_id: UUID) -> Incident:
        incident = session.exec(
            select(Incident)
            .options(selectinload(Incident.alerts))
            .where(Incident.id == incident_id)
        ).first()
    return incident
```

```
def get_incident_by_id(tenant_id: str, incident_id: str) -> Optional[Incident]:
        query = session.query(
            Incident,
        ).filter(
            Incident.tenant_id == tenant_id,
            Incident.id == incident_id,
        )

    return query.first()
```

First of them is not used (and can't be used) by any other code and I guess could/should be removed

**To Reproduce**
Observe it in https://github.com/keephq/keep/blob/main/keep/api/core/db.py

achivkun:
Closes #1561

## 📑 Description
Now we have 2 definitions of the same function:
```
def get_incident_by_id(incident_id: UUID) -> Incident:
def get_incident_by_id(tenant_id: str, incident_id: str) -> Optional[Incident]:
```
Only 2nd definition is used. I suppose there are no much reasons to keep first one.

## ✅ Checks

### DIFF
--- a/keep/api/core/db.py
+++ b/keep/api/core/db.py
@@ -1840,16 +1840,6 @@ def update_preset_options(tenant_id: str, preset_id: str, options: dict) -> Pres
     return preset
 
 
-def get_incident_by_id(incident_id: UUID) -> Incident:
-    with Session(engine) as session:
-        incident = session.exec(
-            select(Incident)
-            .options(selectinload(Incident.alerts))
-            .where(Incident.id == incident_id)
-        ).first()
-    return incident
-
-
 def assign_alert_to_incident(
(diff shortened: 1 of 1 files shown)


Matvey-Kuk (approved):

Matvey-Kuk:
Thank you!

PR #1594, opened 2024-08-12
chore: Keep provider to support different operators when filtering

talboren:


shahargl:
you can now use Common Expression Logic to filter alerts (https://github.com/keephq/keep/pull/1617)

example:
https://github.com/keephq/keep/blob/main/examples/workflows/resolve_old_alerts.yml#L15

Issue #1539, opened 2024-08-05
[➕ Feature]: Merging of Incidents


Matvey-Kuk:
If there are 2 incidents which are accidentally raised for the same reason, there should be an ability to merge them. 

- All associated objects should be also linked to the new incidents
- There should be a record in the DB that incidents are merged to the new one.
- The old Incident link should work since users send incident links to each other.

achivkun:
@Matvey-Kuk, could you please asign it to me?

Matvey-Kuk:
With pleasure! @achivkun

achivkun:
It looks like for now I have less time than expected for this activity. I started to look at this yesterday, even did something, but then discovered that I most probably went wrong direction. @Matvey-Kuk please let me ask to clarify few things before I'll go further:
1) When we do merge - should we remove original incidents automatically (I guess this was the idea)? Alternative is to keep them, but make linked to new incident as children. Later user may decide if he/she wants to remove them or no. Second alrernative requires to add new field to Incident model to setup relation between incidents. I went second way initially, but then realized that first one is most probably those what was initially supposed here.
But at the same time second alternative (when we keep original incidents) makes it easier to implement following requirement:

> The old Incident link should work since users send incident links to each other.

2) Regarding this requirement: 

> There should be a record in the DB that incidents are merged to the new one.

Is there some specific table for this purpose, which is already exists? Or we have to add kind of "Audit/Timeline" table, which will store different events and their relation to certain object(s)?
In UI I see `Timeline` section, which is marked as `Coming Soon...` and I feel that this is the good place where this info can be placed, but I didn't find corresponding table on DB level. We have `AlertAudit`, but not `IncidentAudit`.

frwl404:
My vacation has ended last week, I have no time to work on this ticket anymore. Please feel free to re-assign it.

Kiryous:
Closed by #2274

PR #1614, opened 2024-08-13
feat: discuss DB-level changes, related to incidents merging. refs #1539

frwl404:
Related #1539 

## 📑 Description
@Matvey-Kuk @talboren 

I want to use this PR (which is probably not needed) to discuss questions, which I raised [here](https://github.com/keephq/keep/issues/1539#issuecomment-2280709812) some time back.
I see 2 alternatives how to implement this feature:
- remove initial incidents, when we do merge (in this case current PR is not needed, but it is still to be useful, because I learned something about non-trivial alembic migrations from it).
- keep all initial incidents, but make them children of new one (merge root). Users will have option to remove children later if they will need. In this alternative current PR is needed. With this alternative it seems to be easy to fulfill following requrement:

> The old Incident link should work since users send incident links to each other

And 2nd question, which I have, related to this requirement:

>  There should be a record in the DB that incidents are merged to the new one.

Is there some specific table for this purpose, which is already exists? Or we have to add kind of "Audit/Timeline" table, which will store different events and their relation to certain object(s)?
In UI I see `Timeline` section, which is marked as `Coming Soon...` and I feel that this is the good place where this info can be placed, but I didn't find corresponding table on DB level. We have `AlertAudit`, but not `IncidentAudit`.

## ✅ Checks


Matvey-Kuk:
@frwl404 sorry for long reply!

The challenging part of this Issue is actually to make those architectural decisions, so I hope you don't mind if I won't make them, but will share the context which could help you.

Both options sound valid for me, in general we try to keep all changes minimalistic and not to introduce extensive data structures. Also I could comment that I'm not sure that making a user to remove left-overs after incidents are merged is a way to go :)

> Is there some specific table for this purpose, which is already exists? Or we have to add kind of "Audit/Timeline" table, which will store different events and their relation to certain object(s)?

There is not, Incidents table is pretty basic.

> In UI I see Timeline section, which is marked as Coming Soon... and I feel that this is the good place where this info can be placed, but I didn't find corresponding table on DB level. We have AlertAudit, but not IncidentAudit.

Data structure for this is not in the codebase yet.

talboren:
@Matvey-Kuk @frwl404 closing this for now as this is stale.

PR #1615, opened 2024-08-14
feat: alert groups create incidents instead of alerts

### LINKED ISSUE #1501: [🔨 Enhancement]: Alert Groups should create incidents
As a user, when configuring Alert Groups rules, I want an **incident** created when the rule matches, **instead of the new alert** that is created as of today.

This work is mostly in the `RulesEngine` (rulesengine.py) at `run_rules` function. Happy to answer any question in our [Slack](https://slack.keephq.dev)

Extra requirements:

* Created incidents are not predictions, so `Incident.is_predicted` should be `False`
* Add flag "Create Incidents as Candidates" to rule config. If it's `True` set `Incident.is_confirmed` to `False`

Optional:
* Send an update to the UI using a [pusher](https://github.com/keephq/keep/blob/7060d721650511bfd39de7ecd010c414417713f6/keep/api/core/dependencies.py#L420) after creating new incidents

Savokr:
Closes #1501 

## 📑 Description

This PR is changing the behaviour for triggers of alert groups, which are now creating incidents instead of new alert.

## ✅ Checks


keep/api/tasks/process_event_task.py
```diff
@@ -302,7 +302,8 @@ def __handle_formatted_events(
         # if new grouped alerts were created, we need to push them to the client
         if grouped_alerts:
             logger.info("Adding group alerts to the workflow manager queue")
-            workflow_manager.insert_events(tenant_id, grouped_alerts)
+            #workflow_manager.insert_events(tenant_id, grouped_alerts)
+            #Inserting of the incident into workflow manager
```
Savokr, line 306: Is there a standard way for incidents to be added to the workflow manager or there should be a new method similar to `insert_events`?

keep/rulesengine/rulesengine.py
```diff
@@ -132,40 +132,65 @@ def run_rules(self, events: list[AlertDto]):
...
+            # )
+            # elastic_client = ElasticClient(self.tenant_id)
+            # elastic_client.index_alert(
+            #     alert=AlertDto(event_id=str(group_alert.id), **event),
+            # )
+            # grouped_alerts.append(group_alert)
+            # self.logger.info(f"Created alert {group_alert.id} for group {group.id}")
+            
+            # Create the incident
+            incident_data = {
```
Savokr, line 168: Not sure whether this is a proper way to add additional information for an incident, let me know what should be included.

keep/rulesengine/rulesengine.py
```diff
@@ -132,40 +132,65 @@ def run_rules(self, events: list[AlertDto]):
...
-                "status": group_status,
-                "pushed": True,
-                "group": True,
-                # "groupPayload": group_payload,
-                "fingerprint": group_fingerprint,
-                **group_attributes,
+                "is_predicted": True,
             }
-            group_alert = create_alert_db(
+            incident = create_incident_from_dict(
```
Savokr, line 177: One question I had here, I suppose we don't want to create a new incident for each new trigger of alert group, rather we want to add new alerts to the same incident we created for the initial trigger. Is my understanding correct, i.e. there should be some new logic for checking whether incident with specific fingerprint already exists, or is it something else expected here?

VladimirFilonov:
Hey, [Savokr](https://github.com/Savokr)! Thank you for the pull request. Switching from AlertGroups to incidents requires more profound changes, including database changes, logic updates, and code removal. 

Unfortunately, changes you provided were not complete.  For example, just replacing call create_alert_db with create_incident_from_dict will not actually create an incident with content equal to Group Alert. It will just create an empty Incident without any alerts and relations to the rule. 

As a result, they are absorbed by another PR (https://github.com/keephq/keep/pull/1625)

PR #1625, opened 2024-08-15
feat: ex-Alert Groups are now creating incidents

### LINKED ISSUE #1501: [🔨 Enhancement]: Alert Groups should create incidents

VladimirFilonov:
Closes #1501 

## 📑 Description
Alert Groups are no longer helpful after introducing Incident functionality but can still be useful as an incident automation tool. This PR adjusts Corellation rules to create Incidents instead of Alert Groups.

## ✅ Checks


keep-ui/components/navbar/NoiseReductionLinks.tsx
```diff
@@ -40,7 +40,7 @@ export const NoiseReductionLinks = ({ session }: NoiseReductionLinksProps) => {
       <Disclosure.Panel as="ul" className="space-y-2 p-2 pr-4">
         <li>
           <LinkWithIcon href="/rules" icon={Rules}>
-            <Subtitle>Alert Groups</Subtitle>
+            <Subtitle>Correlations</Subtitle>
```
Matvey-Kuk, line 43: ❤️ !!!!!

keep/api/models/db/migrations/versions/2024-08-14-18-30_87594ea6d308.py
```diff
@@ -0,0 +1,57 @@
+"""add rules-related fields to the incident
```
shahargl, line 1: wouldn't it make sense to have one migration per PR?

shahargl:
update docs - mainly 
- https://docs.keephq.dev/overview/ruleengine
- https://docs.keephq.dev/api-ref/groups/get-groups

keep-ui/app/rules/CorrelationTable.tsx
```diff
@@ -94,14 +95,6 @@ export const CorrelationTable = ({ rules }: CorrelationTableProps) => {
       columnHelper.accessor("name", {
         header: "Correlation Name",
       }),
-      columnHelper.display({
```
shahargl, line 97: why? how a user would have indication for if this rule works or not?

keep/api/models/db/migrations/versions/2024-08-14-18-30_87594ea6d308.py
VladimirFilonov: Going to squash them

keep-ui/app/rules/CorrelationTable.tsx
VladimirFilonov: These columns were always empty. I'm going to improve this UI in one of next steps

VladimirFilonov:
> update docs - mainly
> 
> * https://docs.keephq.dev/overview/ruleengine
> * https://docs.keephq.dev/api-ref/groups/get-groups

Will do in the next iteration: https://github.com/keephq/keep/issues/1662

keep-ui/app/rules/CorrelationTable.tsx
VladimirFilonov: https://github.com/keephq/keep/issues/1664

Matvey-Kuk:
A few notes from my testing:

- [ ] I think we should highlight that those "Correlations" are manual
Screenshot 2024-08-20 at 12 47 28

- [ ] In the incident page, now it's predicted by AI & Correlation Rules
Screenshot 2024-08-20 at 12 48 10

- [ ] If I set Correlation's "scan every" to "once per 5 seconds", it's slicing my alerts to batches of 5 seconds. I like the fact that it's appending to the same incident if I send alerts every 4 seconds, but I guess the config name is a bit misleading. Let's call it "Append to the same Incident if delay between alerts is below": 
Screenshot 2024-08-20 at 12 52 26

Screenshot 2024-08-20 at 12 49 01

- [ ] I was a bit confused by "34 alert(s) were found matching this condition", made me feel like correlation will work retrospectively, but it won't. Maybe nice to add some text that correlation will work for new alerts only, not for historical ones?

VladimirFilonov:
> I was a bit confused by "34 alert(s) were found matching this condition", made me feel like correlation will work retrospectively, but it won't. Maybe worth adding some text that correlation will work for new alerts only, not for historical ones?

Maybe not the best UI solution, but visible =) Will think about better one, but out of this PR

image

Matvey-Kuk (approved):

Matvey-Kuk (approved):

Issue #1632, opened 2024-08-17
[🔨 Enhancement]: Performance issues on > 500k alerts


shahargl:

talboren:
This is not relevant anymore. We should be able to digest this amount, especially with recent changes and elastic.

PR #1633, opened 2024-08-17
feat: performance

shahargl:
close https://github.com/keephq/keep/issues/1632

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# Start the REST API with 4 workers
+echo "Starting REST mode..."
+gunicorn "keep.api.api:get_app(mode='rest')" \
+  --bind 0.0.0.0:8080 \
+  --workers 4 \
+  -k uvicorn.workers.UvicornWorker \
+  -c /venv/lib/python3.11/site-packages/keep/api/config.py &
+
+# Check if REDIS is true, and start the worker mode if it is
+# I'm using random port 1234 because it doesn't matter
+if [ "$REDIS" = "true" ]; then
+  echo "Starting worker mode..."
+  gunicorn "keep.api.api:get_app(mode='worker')" \
+    --workers 4 \
+    -b 0.0.0.0:1234 \
+    -k uvicorn.workers.UvicornWorker &
+fi
+
+# Wait for all background processes to complete
+wait -n
+
+# Exit with the status of the first process to exit
+exit $?
--- a/keep/api/api.py
+++ b/keep/api/api.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import os
+import time
 from importlib import metadata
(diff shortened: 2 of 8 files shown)


docker/entrypoint.sh
```diff
@@ -0,0 +1,25 @@
...
+gunicorn "keep.api.api:get_app(mode='rest')" \
+  --bind 0.0.0.0:8080 \
+  --workers 4 \
+  -k uvicorn.workers.UvicornWorker \
+  -c /venv/lib/python3.11/site-packages/keep/api/config.py &
+
+# Check if REDIS is true, and start the worker mode if it is
+# I'm using random port 1234 because it doesn't matter
+if [ "$REDIS" = "true" ]; then
+  echo "Starting worker mode..."
```
talboren, line 14: Dunno why this change but small comment: arq is a cli tool and can be executed independently. Guincorn is redundant in this case ❤️

shahargl: Yea I know I didn't to keep it a minimal change.

The full context is wide by this specific change is because if there are lot of alerts in the queue, the HTTP requests are throttled and not served, which I wanted to change by separating between workers and REST processes.

talboren (approved):
LGTM

talboren (changes requested):
1 thing before lgtming: looks like there's a migration needed here, no?

keep/api/api.py
```diff
@@ -159,84 +161,100 @@ def get_app(
...
-                "role": user.role,
-            }
+                }
+
+    # WORKER MODE
+    elif mode == "worker":
+        # Initialize only the worker-related components
+        @app.on_event("startup")
+        async def start_worker():
+            if REDIS and WORKER_ENABLED:
```
Matvey-Kuk, line 252: Looks like this PR is making WORKER_ENABLED env var obsolete, it used to serve the same purpose:
- REDIS & WORKER_ENABLED for worker
- REDIS & !WORKER_ENABLED for api only

docker/entrypoint.sh
```diff
@@ -0,0 +1,25 @@
...
+echo "Starting REST mode..."
+gunicorn "keep.api.api:get_app(mode='rest')" \
+  --bind 0.0.0.0:8080 \
+  --workers 4 \
+  -k uvicorn.workers.UvicornWorker \
+  -c /venv/lib/python3.11/site-packages/keep/api/config.py &
+
+# Check if REDIS is true, and start the worker mode if it is
+# I'm using random port 1234 because it doesn't matter
+if [ "$REDIS" = "true" ]; then
```
Matvey-Kuk, line 13: I don't think REDIS could be definitive for worker, shouldn't api be able to schedule tasks for worker via redis?

Matvey-Kuk: Other words, API should be able to access REDIS

PR #1721, opened 2024-08-25
fix: rule_fingerprint stores not hashed value, so can be bigger than VARCHAR

### LINKED ISSUE #1718: [🐛 Bug]: Data too long for column 'rule_fingerprint' at row 1
**Describe the bug**
Column length is not enough to store value0

VladimirFilonov:
Closes #1718


talboren (approved):
lgtm

PR #1734, opened 2024-08-26
feat: improve correlation rules form

### LINKED ISSUE #1692: [🔨 Enhancement]: Improve correlation rules form
image

- [ ] Push final result callout somewhere higher
- [ ] Move “Rules will be applied” to the bottom
- [ ] Do not show errors on the new empty form
- [ ] Add a correlation image to the top
- [ ] For each group inside the rule, request to search for intersections with other rules and show a notification if any exist.
- [ ] Change default timeframe to 24 hours

VladimirFilonov:
Closes #1692

## 📑 Description

image

## ✅ Checks

### DIFF
--- a/keep-ui/app/rules/CorrelationSidebar/AlertsFoundBadge.tsx
+++ b/keep-ui/app/rules/CorrelationSidebar/AlertsFoundBadge.tsx
@@ -5,7 +5,7 @@ import { AlertDto } from "app/alerts/models";
 type AlertsFoundBadgeProps = {
   alertsFound: AlertDto[];
   isLoading: boolean;
-  vertical: boolean;
+  vertical?: boolean;
 };
 
 export const AlertsFoundBadge = ({
@@ -30,10 +30,10 @@ export const AlertsFoundBadge = ({
 
   return (
     <Badge className="mt-3 w-full" color="teal">
-      <span className={`flex items-center justify-center flex-wrap ${vertical ? "mt-2 mb-2" : ""}`}>
+      <span className={`flex items-center justify-center flex-wrap ${vertical ? "mt-2 mb-2 gap-y-3 gap-x-2" : ""}`}>
         {images.map((source, index) => (
           <Image
-            className={`inline-block ${index == 0 && !vertical ? "ml-2" : "-ml-2"}`}
+            className={`inline-block ${index == 0 || vertical ? "" : "-ml-2"}`}
             key={source}
             alt={source}
             height={24}
@@ -44,7 +44,7 @@ export const AlertsFoundBadge = ({
         ))}
         {vertical && <span className="basis-full"></span>}
         <span className="ml-4">
-          {alertsFound.length} alert(s) were found matching this condition
+          {alertsFound.length} alert(s) were found{vertical && <br/>}matching this condition
         </span>
       </span>
     </Badge>
--- a/keep-ui/app/rules/CorrelationSidebar/AlertsFoundBadge.tsx
+++ b/keep-ui/app/rules/CorrelationSidebar/AlertsFoundBadge.tsx
@@ -5,11 +5,13 @@ import { AlertDto } from "app/alerts/models";
 type AlertsFoundBadgeProps = {
   alertsFound: AlertDto[];
   isLoading: boolean;
+  vertical: boolean;
 };
 
 export const AlertsFoundBadge = ({
   alertsFound,
   isLoading,
+  vertical = false,
 }: AlertsFoundBadgeProps) => {
   if (alertsFound.length === 0) {
     return (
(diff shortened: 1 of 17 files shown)


shahargl:
(before even reviewing)
Looks cool :)

keep-ui/app/rules/CorrelationSidebar/index.tsx
```diff
@@ -7,8 +7,8 @@ import { RuleGroupType } from "react-querybuilder";
 export const DEFAULT_CORRELATION_FORM_VALUES: CorrelationForm = {
   name: "",
   description: "",
-  timeAmount: 5,
-  timeUnit: "minutes",
+  timeAmount: 24,
+  timeUnit: "hours",
```
shahargl, line 11: its still show it in seconds (86400) which looks bit odd

VladimirFilonov: The problem here is that the backend stores seconds. I decided not to add a field right now to store the unit right now, but if you think it's important - I'll do

keep-ui/app/rules/CorrelationSidebar/AlertsFoundBadge.tsx
```diff
@@ -40,8 +42,9 @@ export const AlertsFoundBadge = ({
             src={`/icons/${source}-icon.png`}
           />
         ))}
+        {vertical && <span className="basis-full"></span>}
         <span className="ml-4">
-          {alertsFound.length} alert(s) were found matching this condition
+          {alertsFound.length} alert(s) were found{vertical && <br/>}matching this condition
```
shahargl, line 47: So - for the "specific group" it makes sense for me to say "1000 alerts were found", but for the whole rule (x groups) - what does it mean "1000 alerts were found matching this condition"? its not a condition, and its not the outcome I would think about. I would think "This rule correlate 1500 alerts into 5 incidents" which is WAY more valuable.

(also - can we write alert if alerts == 1 else alerts?)

shahargl:
Not on the original ticket but maybe:
1. On the time input, explain what it means. Maybe a callout that will say "when the first alert arrives, Keep will start to calculate the timespan, any new alert within the timeframe will correlate into the same incident")
2. Same for attributes - "Keep will use these attributes to split between incidents. For example,  with attribute `host`, Keep will correlate alert with `hostX` and alert with host `hostY` to different incidents.

keep-ui/app/rules/CorrelationSidebar/AlertsFoundBadge.tsx
VladimirFilonov: But we don't actually correlate historical data. It can be confusing

keep-ui/app/rules/CorrelationSidebar/index.tsx
VladimirFilonov: Done

keep-ui/app/rules/CorrelationSidebar/AlertsFoundBadge.tsx
VladimirFilonov: I Will do it in the next iteration - it will require quite a bit of work at the backend

VladimirFilonov: https://github.com/keephq/keep/issues/1737

keep/api/models/db/migrations/versions/2024-08-26-17-01_1aacee84447e.py
```diff
@@ -0,0 +1,29 @@
...
+revision = "1aacee84447e"
+down_revision = "1c650a429672"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("rule", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("timeunit", sqlmodel.sql.sqltypes.AutoString(), nullable=False, server_default="seconds")
```
shahargl, line 23: Wouldn't it be easier just to store it as seconds and always translate it on the front end? (e.g. if it's divided by 3600 - show errors if its divided by 84400, show days?)

keep-ui/app/rules/CorrelationTable.tsx
```diff
@@ -28,6 +28,14 @@ import { useRouter, useSearchParams } from "next/navigation";
 import { FormattedQueryCell } from "./FormattedQueryCell";
 import { DeleteRuleCell } from "./CorrelationSidebar/DeleteRule";
 
+
+const TIMEFRAME_UNITS_FROM_SECONDS= {
```
shahargl, line 32: if you are already doing this calculation - why saving the "timeUnit"? isn't it redundant?

keep/api/models/db/migrations/versions/2024-08-26-17-01_1aacee84447e.py
VladimirFilonov: What if I want to avoid seeing minutes and like to see 300 seconds - i'd like to think that I've entered them like this for a reason

keep-ui/app/rules/CorrelationTable.tsx
VladimirFilonov: I'm doing calculations to user-defined units

shahargl (dismissed):
lgtm

talboren (approved):
lgtm

Issue #1203, opened 2024-05-30
feat: support oauth2-proxy authentication


shahargl:
full context - https://getkeep.slack.com/archives/C04PT44MATS/p1717007052868319

in short - add a new authentication mode that let the user delegate the authentication from Keep to oauth2-proxy.

Technically, with `AUTH_TYPE=oauth2-proxy` the following should happen:
1. Keep will trust `HTTP_X_FORWARDED_EMAIL` and `HTTP_X_FORWARDED_ROLE` headers and will skip any authentication methods
2. Keep will allow auto-provisioning 

References:
- https://jysk.tech/unleashing-netbox-in-kubernetes-with-fluxcd-a-journey-of-automation-and-high-uptime-cc02cdc3f84b
- https://github.com/healthchecks/healthchecks/discussions/752#discussioncomment-9587451

lukasmrtvy:
@shahargl hi, any update on this? Thanks

shahargl:
hey @lukasmrtvy - I'm ready to work on it.
A few quick questions:
1. Is this still relevant to you?
2. Wdym with "auto-provisioning"? From my understanding - all user management will be done outside of Keep - Keep will just use `HTTP_X_FORWARDED_EMAIL` as the username and `HTTP_X_FORWARDED_ROLE` as the role, so I'm not sure what provisioning is even needed?

lukasmrtvy:
hi @shahargl 

yes to both. I will reach You on Slack :)

shahargl:
@lukasmrtvy once you give a last "ok" this will be merged - https://github.com/keephq/keep/pull/2014

PR #1801, opened 2024-09-02
[WIP] feat(identity): oauth2-proxy support

shahargl:
close https://github.com/keephq/keep/issues/1203


keep/api/config.py
```diff
@@ -16,15 +16,22 @@
...
     logger.info("Keep server starting")
-    
+
     migrate_db()
 
     # Create single tenant if it doesn't exist
     if AUTH_TYPE in [
-        AuthenticationType.SINGLE_TENANT.value,
-        AuthenticationType.NO_AUTH.value,
+        IdentityManagerTypes.DB.value,
```
talboren, line 24: is this right? "DB" is the new single tenant mode?

keep/api/core/db_on_start.py
```diff
@@ -42,7 +42,7 @@
 engine = create_db_engine()
 
 
-def try_create_single_tenant(tenant_id: str) -> None:
+def try_create_single_tenant(tenant_id: str, create_default_user=True) -> None:
```
talboren, line 45: why is this needed?

talboren (dismissed):
lgtm, elegant

keep/api/config.py
shahargl: yep - with single_tenant the users are being stored on the db (also authentication) so the "provider" here is db

keep/api/core/db_on_start.py
shahargl: cuz with oauth2proxy we don't want default user

talboren:
Closing for now as this is not ready yet.

Issue #1878, opened 2024-09-09
[🐛 Bug]: Documentation typo and grammer requests


mirrormystic:
There's a bunch of grammer and typo mistakes in the docs

PR #1879, opened 2024-09-09
fix: First revision of docs edit

mirrormystic:
close https://github.com/keephq/keep/issues/1878


mirrormystic:
Renamed file is missing

PR #1910, opened 2024-09-12
feat: inject env vars into provider provision json

shahargl:
close https://github.com/keephq/keep/issues/1911


VladimirFilonov (approved):
LGTM

PR #1913, opened 2024-09-12
feat: introduce iterative alert batch processing

### LINKED ISSUE #1912: feat: introduce iterative alert processing

GlebBerjoskin:
Closes #1912


keep/api/models/db/tenant.py
```diff
@@ -37,3 +37,10 @@ class TenantInstallation(SQLModel, table=True):
     bot_id: str
     installed: bool = False
     tenant: Optional[Tenant] = Relationship(back_populates="installations")
+
+
+class TenantAIConfig(SQLModel, table=True):
```
shahargl, line 42: ? 
why you need a new table for that? use configuration column on the existing tenant table

keep/api/core/db_on_start.py
```diff
@@ -143,6 +143,12 @@ def try_create_single_tenant(tenant_id: str) -> None:
                     )
                     logger.info(f"Api key {api_key_name} provisioned")
                 logger.info("Api keys provisioned")
+                
+            tenant_ai_config = session.exec(select(TenantAIConfig).where(TenantAIConfig.tenant_id == tenant_id)).first()
+            if not tenant_ai_config:
+                logger.info("Creating tenant AI config")
+                session.add(TenantAIConfig(tenant_id=tenant_id, config={'last_correlated_batch_start': ''}))
```
shahargl, line 150: last_batch_start is not a configuration, its a metadata on current status so I'm not sure whats happening here

keep/api/arq_worker.py
```diff
@@ -84,7 +83,7 @@ def get_arq_worker(queue_name: str) -> Worker:
         "ARQ_KEEP_RESULT", cast=int, default=3600
     )  # duration to keep job results for
     expires = config(
-        "ARQ_EXPIRES", cast=int, default=3600
+        "ARQ_EXPIRES", cast=int, default=3600*1000
```
shahargl, line 86: I'll need @talboren opinion on that, it affects also other tasks not only AI

keep/api/models/db/tenant.py
GlebBerjoskin: There may be multiple ai configs per tenant, I think we need a separate table for that

keep/api/arq_worker.py
GlebBerjoskin: That's fair, I updated the code and made it conditional (bigger only for ai-related tasks) @shahargl

keep/api/core/db.py
```diff
@@ -14,7 +14,6 @@
 from uuid import uuid4
 
 import numpy as np
-import pandas as pd
```
shahargl, line 17: ? we are not using pandas anymore?

keep/api/core/db.py
```diff
@@ -2107,13 +2111,16 @@ def get_last_incidents(
             session.query(
                 Incident,
             )
-            # .options(joinedload(Incident.alerts))
+            .options(joinedload(Incident.alerts))
```
shahargl, line 2114: ?

keep/api/models/db/tenant.py
```diff
@@ -36,4 +36,4 @@ class TenantInstallation(SQLModel, table=True):
     tenant_id: str = Field(foreign_key="tenant.id")
     bot_id: str
     installed: bool = False
-    tenant: Optional[Tenant] = Relationship(back_populates="installations")
+    tenant: Optional[Tenant] = Relationship(back_populates="installations")
```
shahargl, line 39: ?

keep/api/arq_worker.py
```diff
@@ -84,7 +83,7 @@ def get_arq_worker(queue_name: str) -> Worker:
         "ARQ_KEEP_RESULT", cast=int, default=3600
     )  # duration to keep job results for
     expires = config(
-        "ARQ_EXPIRES", cast=int, default=3600
+        "ARQ_EXPIRES", cast=int, default=3600*1000 if KEEP_ARQ_TASK_POOL == KEEP_ARQ_TASK_POOL_AI else 3600
```
shahargl, line 86: when someone runs Keep with `ARQ_EXPIRES` it overrides the default, what will happen to your AI? if your AI needs more time? you need to warn somehow that expiration is too low. Also, its not clear to someone who will read the code why you did that.

keep/api/core/db_on_start.py
```diff
@@ -143,6 +143,7 @@ def try_create_single_tenant(tenant_id: str) -> None:
                     )
                     logger.info(f"Api key {api_key_name} provisioned")
                 logger.info("Api keys provisioned")
+
```
shahargl, line 146: ?

keep/api/core/db.py
```diff
@@ -2806,3 +2752,4 @@ def get_provider_by_name(tenant_id: str, provider_name: str) -> Provider:
             .where(Provider.name == provider_name)
         ).first()
     return provider
+
```
shahargl, line 2755: ?

shahargl (changes requested):
see comments

shahargl:
also - no tests for this change?

keep/api/core/db.py
GlebBerjoskin: We do, but not in this module.

keep/api/core/db.py
GlebBerjoskin: Someone's recent commit introduced this line as commented. I need joinedload uncommented.

If this is a necessary optimization, I can introduce a flag to function params

keep/api/models/db/tenant.py
GlebBerjoskin: @shahargl Please clarify what is the question here

keep/api/arq_worker.py
GlebBerjoskin: It's a copy of the condition introduced by @Matvey-Kuk [here](https://github.com/keephq/keep/blob/9f3895985e78373dfda54d9a8395e4b4566e538c/keep/api/arq_worker.py#L122).

I'm not sure I get the point, do you propose to have extra logs here, introduce a better config for ai-related operations, add documentation?

GlebBerjoskin:
> also - no tests for this change?

I can add some, but it may be bit of tricky. Please, lmk if it's an important priority @shahargl

keep/api/arq_worker.py
shahargl: 1. dont you see the difference? what you sent from @Matvey-Kuk is a hardcoded value. here its a value from environment value. you changed only the default value when environment value is not supplied. can you tell which user deploys Keep with `ARQ_EXPIRES`?
2. regarding logs and better explainability - that's a goes without saying

keep/api/core/db.py
shahargl: who introduced it and why? if he fixed a bug you don't know about - you are just adding this bug again

keep/api/models/db/tenant.py
shahargl: why this line changed?

GlebBerjoskin: Idk, it seems to be the same. I wasn't changing anything in this section.

keep/api/core/db.py
GlebBerjoskin: Introduced by @VladimirFilonov [here](https://github.com/keephq/keep/commit/51905691d6f576f10e1605dca16cd52558467f6d). According to our discussion, may be uncommented for now.

keep/api/arq_worker.py
GlebBerjoskin: But .ee is not supposed to be used by any external users or am I missing something?

Re extended expiration: I think in this case we need to go back to discussion about possible caveats of an extended expiration time with @talboren

GlebBerjoskin: @shahargl I added extra logs ofc

Matvey-Kuk: I dig a bit into `expires` which is getting assigned to `expires_extra_ms` (check here https://arq-docs.helpmanual.io/). That this value is responsible for task discharging in case old tasks are kept in the queue for too long. It's a time = time_task_scheduled_for - time_task_started. Since it's `time_task_started`, it shouldn't be different for long-lasting or short-lasting tasks.

I propose to keep this value the same for AI or non-AI tasks :)

GlebBerjoskin: Problem resolved with the introduction of ARQ_EXPIRES_AI variable

Matvey-Kuk (approved):

PR #1916, opened 2024-09-13
feat: don't allow update/edit provisioned resources

shahargl:
close https://github.com/keephq/keep/issues/1915


talboren (approved):
LGTM

PR #1923, opened 2024-09-15
fix: don't fail the server if api key already in the secret manager

shahargl:
close https://github.com/keephq/keep/issues/1922


talboren (approved):
lgtm

PR #1925, opened 2024-09-15
feat: add the option to run workflow manually but with alert payload

shahargl:
CleanShot 2024-09-15 at 15 12 29

close https://github.com/keephq/keep/issues/1926
close https://github.com/keephq/keep/issues/1927


talboren (approved):
lgtm

PR #1951, opened 2024-09-17
fix: add minor sleep to setup_alerts to avoid lastRec 

shahargl:
close https://github.com/keephq/keep/issues/1950


VladimirFilonov (approved):
LGTM

Issue #2001, opened 2024-09-25
[➕ Feature]: Redirect to incidents by default


Matvey-Kuk:
Came up on a customer call, for a daily usage it makes more sense to redirect to /incident at start.

PR #2002, opened 2024-09-25
chore(ui): Main redirect to incidents

Matvey-Kuk:
close https://github.com/keephq/keep/issues/2001


keep-ui/next.config.js
```diff
@@ -41,8 +41,8 @@ const nextConfig = {
       : [
           {
             source: "/",
-            destination: "/providers",
-            permanent: true,
+            destination: "/incidents",
+            permanent: process.env.ENV === "production",
```
talboren, line 45: why? 👀

Matvey-Kuk: It's breaking local dev env if you run multiple things at http://localhost:3000 one after another. Once you touch keep, it will always redirect to http://localhost:3000/providers :)

talboren (dismissed):
LGTM but it breaks e2e test so need to adjust.

talboren (approved):
LGTM

PR #2013, opened 2024-09-27
feat: backend workflow validation

### LINKED ISSUE #1885: [➕ Feature]: Improve workflow validation
1. Today "id" is being used as an name. This behavior should be deprecated. If an id is supplied, use it, and use name as name https://github.com/keephq/keep/blob/main/keep/parser/parser.py#L37
2. Deleting an workflow and creating another workflow with the same name shows again all old data since deleting a workflow make it is_deleted=1 and not really deletes the workflow (which is ok). Uploading a workflow without an id should generated different uuid.
3. Improve validation - mark workflow with some flag that says it won't run (e.g. configuration error) until something will fix it

35C4n0r:
Closes #1885
/claim #1885

## 📑 Description
- Today "id" is being used as an name. This behavior has been deprecated. 
- If an id is supplied we use that, else we create a new one and add it to the workflow.
- Names are not used for db queries anymore, except for in here https://github.com/keephq/keep/blob/02f219deb0368955bc69ab5fc36e666423fcc3b8/keep/parser/parser.py#L42, which handles the backwards compatibility.
- `parse` method of `Parser` class is used for workflow validation.
- Added a field called is_valid which checks the validity of the workflow.
- UI changes: 
image

## ✅ Checks


35C4n0r:
fixing tests.

35C4n0r:
All tests pass PR ready for review

talboren:
@shahargl is this ready for merge?

shahargl:
closing until it will be needed again @35C4n0r

PR #2034, opened 2024-09-29
feat: Test if all providers are documented

Matvey-Kuk:
Noticed that ServiceNow provider is not documented or mentioned in our docs. This test will fail if one will add a provider without docs.


talboren (commented):
Would separate the new docs thingy in to a different workflow that only triggers when providers/** directory changes. It makes sense not to run it every time and in general I’m paranoid about all these scripts :( (bad past experience 😅)

Matvey-Kuk:
@talboren good idea, moved!

talboren (approved):
LGTM

PR #2065, opened 2024-10-01
feat(mailgun): receive alerts/events via emails

### LINKED ISSUE #2018: [🔌 Provider]: Get events from email
As a user, I would like to be able to send events/alerts to Keep through an email.

talboren:
close #2018


shahargl:
Looking 💣 !
Do you see a difference between a self-hosted (where it makes sense to bring your own mail provider, mailgun) a managed version (where we can use Keep's mail somehow?

shahargl:
Ok then!

shahargl (approved):
lgtm

PR #2102, opened 2024-10-06
feat(provisioning): allow dashboard provisioning

talboren:


keep/api/routes/dashboard.py
```diff
@@ -31,6 +34,41 @@ class DashboardResponseDTO(BaseModel):
 
 
 router = APIRouter()
+logger = logging.getLogger(__name__)
+
+
+def provision_dashboards(tenant_id: str):
+    dashboards_raw = json.loads(os.environ.get("KEEP_DASHBOARDS", "[]"))
```
shahargl, line 41: could raise exception and Keep won't start

keep/api/routes/dashboard.py
```diff
@@ -31,6 +34,41 @@ class DashboardResponseDTO(BaseModel):
...
+    dashboards_to_provision = [
+        DashboardCreateDTO.parse_obj(dashboard) for dashboard in dashboards_raw
+    ]
+    for dashboard in dashboards_to_provision:
+        logger.info(
+            "Provisioning Dashboard",
+            extra={"dashboard_name": dashboard.dashboard_name},
+        )
+        try:
+            create_dashboard_db(
```
shahargl, line 54: if dashboard already exists? it will create over and over?

tests/test_provisioning.py
```diff
@@ -221,3 +221,21 @@ def test_reprovision_provider(monkeypatch, db_session, client, test_app):
     ]
     assert len(provisioned_providers) == 1
     assert provisioned_providers[0]["type"] == "prometheus"
+
+
+@pytest.mark.parametrize(
```
shahargl, line 226: - maybe add a test where `KEEP_DASHBOARDS` is not a valid json?
- maybe add a test where app reloads more than once?

keep/api/api.py
```diff
@@ -263,6 +264,8 @@ async def on_startup():
             logger.info("Providers loaded successfully")
             WorkflowStore.provision_workflows_from_directory(SINGLE_TENANT_UUID)
             logger.info("Workflows provisioned successfully")
+            provision_dashboards(SINGLE_TENANT_UUID)
```
shahargl, line 267: on workflows, removing workflow from the env var will de-provision it from Keep.

shahargl: can dashboard be de-provisioned?

talboren: IDK. I guess we can do it if the user asks for it. For now, no.

keep/api/routes/dashboard.py
talboren: No. There's a unique constraint for the name of the dashboard.
`__table_args__ = (
        UniqueConstraint(
            "tenant_id", "dashboard_name", name="unique_dashboard_name_per_tenant"
        ),
    )`

shahargl (approved):
lgtm

PR #2111, opened 2024-10-07
feat: Implement auto resolution rules for incident correlations

### LINKED ISSUE #2087: [➕ Feature]: Ability to resolve Incidents from alerts
As a user I have a well-tuned monitoring. It's sending "firing" and "resolve" alerts to Keep. I want keep's Incidents to represent only Firing incidents at any moment of time.

Proposed solution: update correlation mechanism so it will open and resolve Incidents based on inbound alerts.

VladimirFilonov:
Closes #2087

## 📑 Description

image

Now, when provider sends alert with the same fingerprint multiple times with different statuses, user can configure when automatically resolve the Incident: when `all` associated alerts are resolved, when `first` or `last` alert is resolved or `never`. `Never` by default

## ✅ Checks

### DIFF
--- a/keep-ui/app/rules/CorrelationSidebar/CorrelationForm.tsx
+++ b/keep-ui/app/rules/CorrelationSidebar/CorrelationForm.tsx
@@ -4,7 +4,9 @@ import {
   MultiSelectItem,
   NumberInput,
   Select,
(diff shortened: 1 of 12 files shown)


talboren (approved):
LGTM!

PR #2138, opened 2024-10-09
fix: enrichments regex operation

### LINKED ISSUE #2139: [🐛 Bug]: Enrichments use `re.match()` instead of `re.search()`

mfroembgen:
close #2139
## 📑 Description
```
re.match() checks for a match only at the beginning of the string
re.search() checks for a match anywhere in the string (this is what Perl does by default)
```

## ✅ Checks

## ℹ Additional Information
https://docs.python.org/3/library/re.html#re.match
https://docs.python.org/3/library/re.html#re.search

### DIFF
--- a/keep/api/bl/enrichments_bl.py
+++ b/keep/api/bl/enrichments_bl.py
@@ -343,7 +343,7 @@ class EnrichmentsBl:
     def _is_match(value, pattern):
         if value is None or pattern is None:
             return False
-        return re.match(pattern, value) is not None
+        return re.search(pattern, value) is not None
 
     def _check_matcher(self, alert: AlertDto, row: dict, matcher: str) -> bool:
         """
--- a/keep/api/bl/enrichments_bl.py
+++ b/keep/api/bl/enrichments_bl.py
@@ -126,7 +126,7 @@ class EnrichmentsBl:
                         extra={"rule_id": rule.id},
                     )
                     continue
-            match_result = re.match(rule.regex, attribute_value)
+            match_result = re.search(rule.regex, attribute_value)
             if match_result:
                 match_dict = match_result.groupdict()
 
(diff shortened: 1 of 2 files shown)


mfroembgen:
FYI: Still need to test/verify it.

keep/api/bl/enrichments_bl.py
```diff
@@ -343,7 +343,7 @@ def _check_alert_matches_rule(self, alert: AlertDto, rule: MappingRule) -> bool:
     def _is_match(value, pattern):
         if value is None or pattern is None:
             return False
-        return re.match(pattern, value) is not None
+        return re.search(pattern, value) is not None
```
mfroembgen, line 346: Or should it be `re.fullmatch` ? What is the intentional usage?

talboren: Asked GPT to summarize the difference for us :)

	•	re.match(): Checks for a match only at the start of the string.
	•	re.search(): Searches for a match anywhere in the string.
	•	re.fullmatch(): Ensures the pattern matches the entire string.

I think `re.search` is our winner as we want to search for that pattern anywhere in the string.

talboren (approved):
LGTM

keep/api/bl/enrichments_bl.py
mfroembgen: I think it should be `fullmatch` as it otherwise could cause problems with mapping and would maybe defeat the propose of extraction?

talboren: I'm not sure @mfroembgen because I think fullmatch expects the patterns to match the *whole* string.
In some example, where somebody wants to extract severity and name from an alert message:
`INFO: The name of the alert - <TIMESTAMP> - <SOMETHING ELSE>`
it might not work, wdyt?

mfroembgen: @talboren For extraction I totally agree but should it behave the same for the mapping as well. Because I would have thought to extract for example the severity from the alert and utilise the extracted severity then for the mapping. Lets say to add P1-5 if severity is INFO and so on.

talboren: I think it makes sense for mapping as well - your use case would still work.

PR #2159, opened 2024-10-10
feat(api): don't update the key every time but every interval

shahargl:
close https://github.com/keephq/keep/issues/2160


talboren (approved):
lgtm

PR #2226, opened 2024-10-17
feat: Possibility to run workflows for incidents manually

### LINKED ISSUE #2197: [➕ Feature]: Manually execute workflows on incidents
## Problem statement

It is cool that we are able to get bunch of different alerts to be correlated and batched into single incident automatically. But it will be even more useful if it would be possible to execute workflow on this incident. 

Exact user story may looks like this:
* bunch of alerts generated and batched into incident (lets say by a "server" field)
* user can create jira ticket about this incident by executing workflow
* workflow have access to all alerts within incident, so all of them listed in ticket (richer context available)
* incident have backlink to created ticket

Basically it is how it works right now with individual alerts but on incident level. For now if something happens with "server" we got separate alerts for "cpu", "disk", "network" etc and for each of them individually it is possible to create tickets but essentially this is single "Incident" and to have one jira ticket per one incident is great.

But may be I am missing something and this is already possible (we use 0.24.5)?

Thanks

VladimirFilonov:
Closes #2197 

## 📑 Description

https://github.com/user-attachments/assets/70b49746-fa88-4b73-bced-462d7c64a28a

## ✅ Checks

### DIFF
--- a/keep-ui/app/alerts/alerts.tsx
+++ b/keep-ui/app/alerts/alerts.tsx
@@ -9,7 +9,7 @@ import AlertNoteModal from "./alert-note-modal";
 import { useProviders } from "utils/hooks/useProviders";
 import { AlertDto } from "./models";
 import { AlertMethodModal } from "./alert-method-modal";
-import AlertRunWorkflowModal from "./alert-run-workflow-modal";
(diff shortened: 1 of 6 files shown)


talboren (approved):
LGTM

Issue #2302, opened 2024-10-25
[🐛 Bug]: `deleted_at` default is wrong (at least for sqlite) and migration breaks incidents alerts


shahargl:

shahargl:
some context - 

1. On the migration:
```server_default="1000-01-01 00:00:00",```

2. But on SQLite, this isn't the default:
```
print(query.statement.compile(dialect=sqlite.dialect(), compile_kwargs={"literal_binds": True}))
SELECT alert.timestamp, alert.event, alert.id, alert.tenant_id, alert.provider_type, alert.provider_id, alert.fingerprint, alert.alert_hash, alerttoincident.incident_id, alerttoincident.tenant_id AS tenant_id_1, alerttoincident.timestamp AS timestamp_1, alerttoincident.alert_id, alerttoincident.is_created_by_ai, alerttoincident.deleted_at 
FROM alert JOIN alerttoincident ON alerttoincident.alert_id = alert.id JOIN incident ON alerttoincident.incident_id = incident.id 
WHERE alerttoincident.tenant_id = 'keep' AND incident.id = 'bf430dddec1642e7b1c9327c79bbf2cd' AND alerttoincident.deleted_at = '1000-01-01 00:00:00.000000' ORDER BY alert.timestamp DESC
```

3. Local workaround:
```
UPDATE alerttoincident 
SET deleted_at = '1000-01-01 00:00:00.000000' 
WHERE deleted_at = '1000-01-0   ...> SET deleted_at = '1000-01-01 00:00:00.000000' 
```

PR #2338, opened 2024-10-29
fix: migration messing link between alerts and incidents

Matvey-Kuk:
Close: https://github.com/keephq/keep/issues/2302

Some context:
- Using pure SQL to make sure future changes of the python model won't break it.
- I don't fix original migration, but create a new one. It should fix the issue for those who already executed the first one.


VladimirFilonov (approved):
LGTM

PR #2349, opened 2024-10-30
fix: Remove unnecessary Relations between Alerts to Incidents

Matvey-Kuk:
We don't need those relations & excessive complications.

Closes https://github.com/keephq/keep/issues/2350


Matvey-Kuk:
Whoops, we use those relations in tests... Ok.

PR #631, opened 2023-12-15
fix(startup): posthog issue when enabled and no connectivity

vijay-jangir:
Dockerfile enhancement, posthog bugfix

posthog doesn't let application start if there is any connection issue, temporarily fixed it to atleast get the application in started mode.
Updated dockerfile for ui and api so that it can run seemlessly on openshift update dockerfile for api for caching layers more efficiently

Closes # 

## ✅ Checks

- [ ] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [ ] All the tests have passed

PR #2366, opened 2024-10-31
feat: measure amount of running instances

Matvey-Kuk:
- Reverting https://github.com/keephq/keep/pull/631
- Not reporting to posthog if there is no connectivity to it.
- Tested with air-gapped keep.
- Removing duplicate env var: ENABLE_POSTHOG_API because we have DISABLE_POSTHOG
- Reporting backend uptime to PostHog every 10 minutes.
- PostHog is messing up with unnittests causing infinite recursion. Spent a pile of time trying to find the reason, just disabling it for Unittests.

Closes https://github.com/keephq/keep/issues/2367


keep/posthog/posthog.py
```diff
@@ -1,16 +1,47 @@
...
+if DISABLE_POSTHOG:
+    posthog.disabled = True
+
+def get_posthog_client(sync_mode=False):
     posthog_api_key = (
         os.getenv("POSTHOG_API_KEY")
         or "phc_muk9qE3TfZsX3SZ9XxX52kCGJBclrjhkP9JxAQcm1PZ"
     )
-    posthog_client = Posthog(api_key=posthog_api_key, host="https://app.posthog.com")
+    posthog_client = Posthog(api_key=posthog_api_key, host="https://app.posthog.com", sync_mode=sync_mode)
```
talboren, line 24: This will add delay to every request received by Keep

talboren: if I understand it correctly

Matvey-Kuk: It's only for `is_posthog_reachable` to raise Exception if posthog is not reachable. False by default.

keep-ui/.env.local.example
```diff
@@ -14,4 +14,6 @@ PUSHER_HOST=localhost
 PUSHER_PORT=6001
 PUSHER_APP_KEY=keepappkey
 # Logging
-LOG_FORMAT=dev_terminal
\ No newline at end of file
+LOG_FORMAT=dev_terminal
+
+DISABLE_POSTHOG=true
```
talboren, line 19:
```suggestion
```

keep/api/api.py
```diff
@@ -121,47 +120,44 @@ def _extract_identity(request: Request, attribute="email") -> str:
         return "anonymous"
 
 
-class EventCaptureMiddleware(BaseHTTPMiddleware):
+class PostHogEventCaptureMiddleware(BaseHTTPMiddleware):
```
talboren, line 123: can we move to a dedicated file?

keep/posthog/posthog.py
```diff
@@ -1,16 +1,56 @@
...
     return posthog_client
+
+def is_posthog_reachable():
+    posthog_client = get_posthog_client(sync_mode=True)
+    try:
+        posthog_client.capture(
```
talboren, line 36: what's the default timeout? can we put something here? I'm worried that in some air-gapped envs this will make startup "stuck"

keep/posthog/posthog.py
```diff
@@ -1,16 +1,56 @@
...
+            RANDOM_TENANT_ID_PERSISTENT_WITHIN_LAUNCH,
+            "backend_status",
+            properties={
+                "status": "up",
+                "keep_version": KEEP_VERSION,
+            },
+        )
+        logger.info("Uptime reported to PostHog.")
+        time.sleep(UPTIME_REPORTING_CADENCE)
```
talboren, line 56: just add some doc string

talboren (changes requested):
some small comments

keep-ui/.env.local.example
Matvey-Kuk: Why do you suggest this? I added it to disable Posthog for dev envs. The same with the previous change.

keep/posthog/posthog.py
Matvey-Kuk: By default it's 3 sec, made explicit here: https://github.com/keephq/keep/pull/2366/files#diff-80e770595b2d185d5c97ed37847dd7566565fcf981ef6b25d2f88ba23642728bR34

tests/fixtures/client.py
```diff
@@ -12,6 +12,8 @@
 
 @pytest.fixture
 def test_app(monkeypatch, request):
+    monkeypatch.setenv("POSTHOG_DISABLED", "true")
```
talboren, line 15: hmhmhmhm only thing that frightens me

talboren (approved):
LGTM

PR #2382, opened 2024-11-03
fix: don't call read_secret in pull_data_from_providers

talboren:


tests/test_iohandler.py
```diff
@@ -767,63 +767,63 @@ def test_add_time_to_date_function(context_manager):
...
-    )
-    context_manager.alert = alert
-    context_manager.event_context = context_manager.alert
-    iohandler = IOHandler(context_manager)
-
-    # it should be greater than 60 minutes and less than 90 minutes
-    s = iohandler.render(template)
-    # the alert is not really added to the DB so the firing time is 0.00
-    assert s == "0.00 >= 30 and 0.00 < 90"
+# def test_openobserve_rows_bug(db_session, context_manager):
```
VladimirFilonov, line 770: Why comment instead of deleting?

talboren: i'll fix it in future pr

PR #2387, opened 2024-11-04
fix: One more attempt to return posthog

Matvey-Kuk:
Past history:

Check: https://github.com/keephq/keep/pull/2366
Reverted: #2384

Closes https://github.com/keephq/keep/issues/2388


Matvey-Kuk:
Server didn't die last 3 hours at the Cloud Run after the async fix. I assume the issue is resolved now and we could proceed.

PR #2419, opened 2024-11-08
feat(api): keep the failed alerts

shahargl:
close https://github.com/keephq/keep/issues/2418


talboren (approved):
lgtm

keep/api/models/db/migrations/versions/2024-11-08-20-58_895fe80117aa.py
```diff
@@ -0,0 +1,41 @@
...
+    with op.batch_alter_table("alertraw", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column(
+                "timestamp",
+                sa.DateTime(),
+                nullable=False,
+                server_default=sa.text("CURRENT_TIMESTAMP"),
+            )
+        )
+        batch_op.add_column(sa.Column("provider_type", sa.String(), nullable=True))
```
talboren, line 31: in MySQL it fails over:
`VARCHAR requires a length on dialect mysql`
```suggestion
        batch_op.add_column(sa.Column("provider_type", sa.String(255), nullable=True))
```

talboren (changes requested):

VladimirFilonov (approved):
LGTM

PR #2253, opened 2024-10-21
refactor: use nginx instead of rewrite rules

shahargl:
close https://github.com/keephq/keep/issues/2153

PR #2484, opened 2024-11-14
fix: Local pusher & alert feed

Matvey-Kuk:
1) Seems like https://github.com/keephq/keep/pull/2253/files#diff-008fb887e8276618c51329f01feb3092bbc660c780709c1c8beb9d1ebf831c49R28 prevented local & dev envs from connecting to Pusher
2) We didn't trigger poll-alerts so alert feed was not live

Closes https://github.com/keephq/keep/issues/2487


VladimirFilonov (approved):
LGTM

Issue #2518, opened 2024-11-18
[🐛 Bug]: GCPMonitoring documentation is broken


talboren:
We expect documentation to exist in the policy, when it does not exist - the alert fails (name extraction fails)

PR #2519, opened 2024-11-18
fix: minor fix in gcp monitoring

shahargl:
close https://github.com/keephq/keep/issues/2518


talboren (changes requested):
description can be empty too, it seems :(

talboren (approved):
LGTM

PR #2523, opened 2024-11-18
feat: Demo mode

Matvey-Kuk:
LIVE_DEMO_MODE will launch what uset to be simulate_alert.py in the background

Closes https://github.com/keephq/keep/issues/2524

✅ Closed in favor of https://github.com/keephq/keep/pull/2561

## Summary by CodeRabbit

- **New Features**
	- Introduced a new environment variable `KEEP_LIVE_DEMO_MODE` for simulating alerts in demo mode.
	- Added functionality for simulating alerts through the enhanced `demo_mode_runner` module, including structured correlation rules and a robust service topology.

- **Bug Fixes**
	- Enhanced alert configurations across multiple providers to improve specificity and organization.

- **Documentation**
	- Updated deployment documentation to include the new `KEEP_LIVE_DEMO_MODE` environment variable.

- **Chores**
	- Simplified alert simulation logic by integrating with the new demo mode functionality.

PR #2561, opened 2024-11-20
feat: Demo mode v2 & moving background tasks outside of the server

Matvey-Kuk:
Demo mode will fill the instance with some data:

https://github.com/user-attachments/assets/8c91e09c-aaaf-479f-b24b-e872e946ae15

It's a refinement of https://github.com/keephq/keep/pull/2523
- Bringing KEEP_LIVE_DEMO_MODE
- Updates simulate_alert.py with --full-demo
- Moved background tasks to the separate python module, **process separate from server**.
- Launching background tasks using docker's ENTRYPOINT.
- POSTHOG_DISABLED=true for dev server.
- PostHog ID now is persistent in the DB.

Tested:
- simulate_alert.py is backwards compatible.
- If background tasks will fail, server will continue working.

Closes https://github.com/keephq/keep/issues/2562


keep/api/core/report_uptime.py
```diff
@@ -18,19 +18,25 @@ async def report_uptime_to_posthog():
     Should be lunched in a separate thread.
     """
     while True:
+        start_time = asyncio.get_event_loop().time()
```
talboren, line 21: ?

keep/server_jobs_bg.py
```diff
@@ -0,0 +1,45 @@
...
+
+    threads = []
+    threads.append(launch_demo_mode_thread(keep_api_url))
+    threads.append(launch_uptime_reporting_thread())
+    
+    for thread in threads:
+        if thread is not None:
+            thread.join()
+
+    logger.info("Background server jobs started.")
```
talboren, line 37: ?

talboren (dismissed):
👌🏼

talboren (approved):

PR #2685, opened 2024-11-28
fix: workflow id and name usage

vishwamartur:
Related to #1885

Update workflow parsing to use `id` as id and `name` as name, generate different UUID if `id` is not provided, and add validation for configuration errors.

* **keep/parser/parser.py**
  - Update `_get_workflow_id` to use `workflow.get("id")` for id and `workflow.get("name")` for name.
  - Add logic to generate a different UUID if `workflow.get("id")` is not provided.
  - Update `_parse_workflow` to use `workflow.get("name")` for name.

* **keep/api/models/db/workflow.py**
  - Add a new field `invalid` to mark workflows with configuration errors.
  - Update the `Workflow` class to include the `invalid` field.

* **keep/api/routes/workflows.py**
  - Update `create_workflow` and `create_workflow_from_body` to generate a different UUID if `workflow.get("id")` is not provided.
  - Add validation logic to mark workflows with configuration errors as `invalid`.

* **tests/test_parser.py**
  - Add test cases to verify that a different UUID is generated if `workflow.get("id")` is not provided.
  - Add test cases to verify that workflows with configuration errors are marked as `invalid`.


keep/api/models/db/workflow.py
```diff
@@ -21,6 +21,7 @@ class Workflow(SQLModel, table=True):
     last_updated: datetime = Field(default_factory=datetime.utcnow)
     provisioned: bool = Field(default=False)
     provisioned_file: Optional[str] = None
+    invalid: bool = Field(default=False)  # P22f5
```
Matvey-Kuk, line 24: This change should be reflected in a migration. Also what's P22f5? :)

keep/api/routes/workflows.py
```diff
@@ -359,6 +361,8 @@ async def create_workflow_from_body(
     workflowstore = WorkflowStore()
     # Create the workflow
     try:
+        if not workflow.get("id"):
+            workflow["id"] = str(uuid.uuid4())
```
Matvey-Kuk, line 365: May be added to __get_workflow_raw_data?

Matvey-Kuk (changes requested):

talboren:
This is open for more than a month. Closing for now, please re-open if it's relevant
cc: @vishwamartur @Matvey-Kuk

PR #2715, opened 2024-12-01
fix: OnCall provider and docs

Matvey-Kuk:
OnCall provider was actually a duplicate of the Grafana Incident provider. Fixed.

Closes https://github.com/keephq/keep/issues/2716


talboren (approved):
LGTM

PR #2720, opened 2024-12-01
feat(provider): vector dev

talboren:
Still `coming_soon` because it only allows to format events now. It will evolve to also interact with Vector's GraphQL


shahargl:
Champ

shahargl (approved):
lgtm

PR #2727, opened 2024-12-02
feat: multiple alerts matching rules PoC

### LINKED ISSUE #2776: [➕ Feature]: incidents to be created only when multiple alerts matches rule
We need to create a correlation rule, which will also work with historical data. 
 
 Use case: As a User, I want incidents to be created only when multiple alerts are firing per the grouping rule.

Example:
* server fires CPU alert -> Not an incident yet
* server fires Disk Usage alert -> Not an incident yet
* server fires Temperature alert -> Incident1 is created

VladimirFilonov:
Closes #2776

## ✅ Checks


Matvey-Kuk (dismissed):

Matvey-Kuk (dismissed):

Kiryous (approved):
LGTM!

PR #2765, opened 2024-12-05
fix: revert validation for Slack provider to work with oauth

talboren:
Slack does not work with OAuth anymore because of the new validation method, reverting https://github.com/keephq/keep/pull/2430 for slack


shahargl (approved):
lgtm

Matvey-Kuk (approved):

Issue #2827, opened 2024-12-15
[➕ Feature]: Refactor vectordev provider to a wrapper to support multiple providers


sainad2222:
Creating an issue based on our offline discussion about refactoring vectordev provider as a wrapper to support multiple providers
[Slack Ref](https://getkeep.slack.com/archives/C04PT44MATS/p1734253285124639?thread_ts=1734189305.713829&cid=C04PT44MATS)

PR #2828, opened 2024-12-15
Feature: refactor vectordev provider to support wrapping multiple sou…

sainad2222:
Closes https://github.com/keephq/keep/issues/2827
To start with adding support for prometheus and grafana. If this is looking good then I'll add for all other providers too
Screenshot 2024-12-15 at 15 33 32


keep/providers/vectordev_provider/vectordev_provider.py
```diff
@@ -32,33 +40,22 @@ def validate_config(self):
...
-                    name="",
-                    host=e.get("host"),
-                    message=e.get("message"),
-                    description=e.get("message"),
-                    lastReceived=e.get("timestamp"),
-                    source_type=e.get("source_type"),
-                    source=["vectordev"],
-                    original_event=event_json,
-                )
-            )
```
talboren, line 57: @sainad2222 would leave this in if 
`provider_class = ProvidersFactory.get_provider_class(provider)` didn't find the provider (by `event_type`).

talboren:
@sainad2222 thank you so much for the pr! left 1 comment and lgtm :)

keep/providers/vectordev_provider/vectordev_provider.py
```diff
@@ -32,18 +42,25 @@ def validate_config(self):
...
-
-            events.append(
+        try:
+            event_type = event["source_type"]
+            provider_class = ProvidersFactory.get_provider_class(VectordevProvider.SOURCE_TO_PROVIDER_MAP[event_type])
+            return provider_class._format_alert(event["event"],provider_instance)
+        except Exception as e:
+            alert_dtos = []
+            if isinstance(event, list):
+                return event
```
talboren, line 54: Is this correct?

keep/providers/vectordev_provider/vectordev_provider.py
```diff
@@ -32,18 +42,25 @@ def validate_config(self):
...
+            else:
+                alerts = event.get("alerts", [event])
```
talboren, line 56: this is something to expect from vector?

keep/providers/vectordev_provider/vectordev_provider.py
sainad2222: My bad. Let me fix it

sainad2222:
@talboren You can check now. Meanwhile I will try to test the source not found cases

keep/providers/vectordev_provider/vectordev_provider.py
```diff
@@ -32,18 +42,27 @@ def validate_config(self):
...
+            return provider_class._format_alert(event["event"],provider_instance)
+        except Exception as e:
+            alert_dtos = []
+            events = []
+            if isinstance(event, list):
+                events = event
+            else:
+                events = [event]
+            for event in events:
+                e = event["event"]
```
talboren, line 59: I think this may fail if the event doesn't have the "event" key

talboren:
@sainad2222 i left 1 comment, other than that looks good!

sainad2222:
Tested with both `prometheus` as source and some `random` value as source
Screenshot 2024-12-18 at 22 06 04

This is the vector config(which we need currently as long as Keep is not a sink in Vector)
```
sources:
  alertmanager_http:
    type: http
    address: 0.0.0.0:8686 # Configure this address as a receiver webhook in alertmanager

transforms:
  restructure_event:
    inputs:
      - alertmanager_http
    source: |-
      .message = parse_json!(string!(.message))
      .keep_source_type = "random"
    type: "remap"

sinks:
  keep:
    type: "http"
    inputs: ["restructure_event"]
    uri: "http://localhost:8080/alerts/event/vectordev?provider_id=test"
    encoding:
      codec: "json"
    request:
      headers:
        x-api-key: "keepappkey"
        content-type: "application/json"

```

keep/providers/vectordev_provider/vectordev_provider.py
```diff
@@ -32,33 +44,47 @@ def validate_config(self):
...
-            except json.JSONDecodeError:
-                pass
-
-            events.append(
+        if isinstance(event, list):
+            events = event
+        else:
+            events = [event]
+        try:
+            alert_dtos = []
```
talboren, line 55: can define that 1 time in the beginning

keep/providers/vectordev_provider/vectordev_provider.py
```diff
@@ -32,33 +44,47 @@ def validate_config(self):
...
+            for e in events:
+                event_type = e["keep_source_type"]
```
talboren, line 57: I think we can make this more elegant.

Split it 2 to different foramtting functions and check the first event in events. If it has the event_type attribute (i.e., keep_source_type), we can assume that it's a provider we can ingest, otherwise, just add the event like it used to before.

wdyt?

talboren (approved):
LGTM!

PR #2835, opened 2024-12-15
fix: workflow executions in alert name

### LINKED ISSUE #2829: [🐛 Bug]: Alert to workflow execution is not working
As a user, I would like to see which alerts had workflow executions from the alerts feed. It used to show a gear icon but it doesn't work anymore

image

talboren:
close #2829


shahargl (approved):
lgtm

PR #2922, opened 2024-12-29
feat: Add soft-delete for incidents through the status

### LINKED ISSUE #2909: [➕ Feature]: See deleted/resolved incidents
As a user, I would like to see deleted/resolved incidents in a different panel than the default incident page, or at least filter them by default.

As a user, right now, I cannot see incidents that were deleted.

VladimirFilonov:
Closes #2909

## ✅ Checks

### DIFF
--- a/keep/api/core/db.py
+++ b/keep/api/core/db.py
@@ -37,7 +37,7 @@ from sqlalchemy.dialects.mysql import insert as mysql_insert
 from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.dialects.sqlite import insert as sqlite_insert
 from sqlalchemy.exc import IntegrityError, OperationalError
-from sqlalchemy.orm import joinedload, selectinload, subqueryload
+from sqlalchemy.orm import joinedload, subqueryload
 from sqlalchemy.sql import exists, expression
 from sqlmodel import Session, SQLModel, col, or_, select, text
 
--- a/keep/api/core/db.py
+++ b/keep/api/core/db.py
@@ -3481,7 +3481,6 @@ def get_incident_by_fingerprint(tenant_id: str, fingerprint: str) -> Optional[In
 def delete_incident_by_id(
     tenant_id: str,
     incident_id: UUID,
-    deleted_by: str | None = None,
 ) -> bool:
     if isinstance(incident_id, str):
         incident_id = __convert_to_uuid(incident_id)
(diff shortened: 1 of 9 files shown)


keep/api/models/alert.py
```diff
@@ -113,6 +113,8 @@ class IncidentStatus(Enum):
     ACKNOWLEDGED = "acknowledged"
     # Incident was merged with another incident
     MERGED = "merged"
+    # Incident was merged with another incident
```
talboren, line 116: fix comment

tests/test_incidents.py
```diff
@@ -227,7 +227,7 @@ def test_add_remove_alert_to_incidents(db_session, setup_stress_alerts_no_elasti
 def test_get_last_incidents(db_session, create_alert):
```
talboren, line 227: can we add a test for deleting an incident and making sure that it gets the right value?

keep/api/core/db.py
```diff
@@ -3489,31 +3481,30 @@ def get_incident_by_fingerprint(tenant_id: str, fingerprint: str) -> Optional[In
 def delete_incident_by_id(
     tenant_id: str,
     incident_id: UUID,
+    deleted_by: str | None = None,
```
talboren, line 3484: use deleted_by when you update?

keep/api/core/db.py
```diff
@@ -2921,25 +2922,16 @@ def is_alert_assigned_to_incident(
     with Session(engine) as session:
         assigned = session.exec(
             select(LastAlertToIncident)
+            .join(Incident, LastAlertToIncident.incident_id == Incident.id)
```
talboren, line 2925: is this a bug now?

talboren (changes requested):
left a few minor comments

keep/api/core/db.py
VladimirFilonov: Not actually. If incident is deleted we treat it as Alert is not assigned to the incident anymore

keep/api/core/db.py
VladimirFilonov: It's leftover from the first try when I added new fields. Now we need to have it in Audit as status change. BTW, do we track status changes in IncidentAudit?

tests/test_incidents.py
VladimirFilonov: We have `test_incident_bl_delete_incident` where we checking the status (reversed, but still )

talboren (approved):
LGTM

PR #2961, opened 2025-01-01
feat(api): topology incidents

### LINKED ISSUE #2984: [➕ Feature]: Incident improvements
- Topology based incident correlation
- Incident enrichments
- Chat capable of executing provider methods
- Cilium to create applications automatically
- More...

shahargl:
close #2984


keep/api/core/tracer.py
```diff
@@ -0,0 +1,46 @@
...
+from opentelemetry.trace import Link, SpanKind
+from opentelemetry.trace.span import TraceState
+from opentelemetry.util.types import Attributes
+
+
+class KeepSampler(sampling.Sampler):
+    def __init__(self, parent_sampler=None):
+        self.parent_sampler = parent_sampler or sampling.ParentBased(sampling.ALWAYS_ON)
+        # Operations we want to exclude from tracing
+        self.excluded_operations = {
```
Matvey-Kuk, line 15: <3

keep/api/core/db.py
```diff
@@ -138,10 +138,12 @@ def get_session_sync() -> Session:
     return Session(engine)
 
 
-def __convert_to_uuid(value: str) -> UUID | None:
+def __convert_to_uuid(value: str, should_raise: bool = False) -> UUID | None:
```
Matvey-Kuk, line 141: Looks for me that it should always raise since it's working with internally-generated ID's only. 🤔

Screenshot 2025-01-13 at 13 36 00

Matvey-Kuk (approved):

keep/api/core/db.py
talboren: > Looks for me that it should always raise since it's working with internally-generated ID's only. 🤔
> 
> Screenshot 2025-01-13 at 13 36 00

The problem was introduced when we moved to SQLAlchemy 2.0, which doesn't allow "string" uuids in queries anymore (but constraints the use of UUID object), and sometimes, there are endpoints where we accept uuid as string.

PR #2966, opened 2025-01-02
fix: event as list[AlertDTO]  not included in the condition

### LINKED ISSUE #2974: [🐛 Bug]: _format_alert function in provider called with incorrect params
## 📑 Description
in one of the conditions of the process_event function there is a flaw in processing event as a list. 2 months ago there was no such problem, but now when I upgraded the version of keep I found that zabbix provider started getting an error. while studying the issue (2 days) I finally understood what was the problem. The point is that in one of the conditions there is an attempt to convert dict into some formatted model, but the check condition checks list as well, without even trying to understand what is inside. The base_provider function requires implementing _format_alert, where event=dict. This means that you cannot pass list, or even list[alertDTO]. 

# base_provider
```python
@staticmethod
    def _format_alert(
        event: dict | list[dict], provider_instance: "BaseProvider" = None
    ) -> AlertDto | list[AlertDto]:
```

00041275:
close #2974 

## 📑 Description
in one of the conditions of the process_event function there is a flaw in processing event as a list. 2 months ago there was no such problem, but now when I upgraded the version of keep I found that zabbix provider started getting an error. while studying the issue (2 days) I finally understood what was the problem. The point is that in one of the conditions there is an attempt to convert dict into some formatted model, but the check condition checks list as well, without even trying to understand what is inside. The base_provider function requires implementing _format_alert, where event=dict. This means that you cannot pass list, or even list[alertDTO]. The change ensures that the _format_alert call is excluded if it is not dict.

# base_provider
```python
@staticmethod
    def _format_alert(
        event: dict | list[dict], provider_instance: "BaseProvider" = None
    ) -> AlertDto | list[AlertDto]:
```

## ✅ Checks
- [х] All the tests have passed

### DIFF
--- a/keep/api/tasks/process_event_task.py
+++ b/keep/api/tasks/process_event_task.py
@@ -597,12 +597,26 @@ def process_event(
                 except Exception:
                     provider_class = ProvidersFactory.get_provider_class("keep")
 
-                event = provider_class.format_alert(
-                    tenant_id=tenant_id,
-                    event=event,
-                    provider_id=provider_id,
-                    provider_type=provider_type,
-                )
+                if isinstance(event, list):
+                    event_list = []
+                    for event_item in event:
+                        if not isinstance(event_item, AlertDto):
+                            event_list.append(provider_class.format_alert(
+                                tenant_id=tenant_id,
+                                event=event_item,
+                                provider_id=provider_id,
+                                provider_type=provider_type,
+                            ))
+                        else:
+                            event_list.append(event_item)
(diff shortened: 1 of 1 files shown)


talboren:
Thought we took care of it! Thank you. 
I’ll push that in first thing in the morning

talboren:
@00041275 may I ask you to create an issue for this PR too?

talboren (approved):
LGTM

00041275:
created #2974

PR #2994, opened 2025-01-07
docs: Update howdoeskeepgetmyalerts.mdx

### LINKED ISSUE #2975: [➕ Feature]: freedom for variable KEEP_PULL_INTERVAL
## 📑 Description

At the moment, keep prefers to work with webhook from source systems. But my company prefers to initiate a request from keep (for example, Oracle Enterprise Management). In view of this, it is problematic to have a limitation in a variable "KEEP_PULL_INTERVAL", which represents days. As more flexibility, it would be better to give users the ability to configure the pulling time themselves (for example, in minutes). Of course, ideally for the pull data option, it is best to have such a variable in the provider itself, but as an intermediate flexible solution, I suggest considering giving users the freedom to configure data collection from sources themselves.

00041275:
Close #2975

missed to update doc for issue #2975


shahargl (approved):
Lgtm

PR #3002, opened 2025-01-09
feat: Amazon SQS Provider

### LINKED ISSUE #2963: [🔌 Provider]: AWS SQS
**Describe the provider you want to add**
As a developer, I want to receive and send alerts from and to SQS.

**Describe your use case**
This will enable us to use and integrate keep into our cloud architecture.
With that we can call a lambda from that queue and execute custom code for integration in our platform.
It will also be a lot cheaper for us to use keep, because instead of keep querying our database all the time, we can simply check critical values during preprocessing and notify keep.

**Are you already using Keep?**
No

**Additional context**
Add any other context or screenshots about the provider request here.

35C4n0r:
Closes #2963

## ✅ Checks


docs/providers/documentation/amazonsqs-provider.mdx
```diff
@@ -0,0 +1,45 @@
+---
```
shahargl, line 1: can you add some example for input/output payload? for example - how event payload should look like for Keep to process it? what are mandatory fields?

keep/providers/amazonsqs_provider/amazonsqs_provider.py
```diff
@@ -0,0 +1,279 @@
...
+    @staticmethod
+    def get_status_or_default(status_value):
+        try:
+            # Check if status_value is a valid member of AlertStatus
+            return AlertStatus(status_value)
+        except ValueError:
+            # If not, return the default AlertStatus.FIRING
+            return AlertStatus.FIRING
+
+    def _get_alerts(self) -> list[AlertDto]:
```
shahargl, line 236: see https://github.com/keephq/keep/blob/main/keep/providers/kafka_provider/kafka_provider.py#L227

shahargl (changes requested):
sqs provider should be "queue type" like Kafka - https://github.com/keephq/keep/blob/main/keep/providers/kafka_provider/kafka_provider.py#L227

PR #3042, opened 2025-01-16
fix: correlation rule `is_deleted`

talboren:
Instead of deleting the correlation rule we will mark it as deleted and show incident that were created by that rule as orphans:

CleanShot 2025-01-16 at 14 28 21@2x


shahargl (approved):
lgtm

PR #3047, opened 2025-01-16
feat(docs): add sound.mdx

### LINKED ISSUE #3018: [📃 Docs]: sound notification mechanism work
How does the sound notification mechanism work? Who will hear it? Is it correct to launch a notification for everyone who has entered keep, maybe it is worth launching it from the dashboards with set audiofile to play? Is there a poll for new alerts and its frequency or is it expected that the webhook will deliver information and the sound will go? Why I can't change the sound? 1.5 minutes of audio is something cool at first, but disgusting after 5 minutes?
If I was heard from the dashboards by sounds, then I would like to add an interval polling function. The current dashboards are more like reports than operational dashboards.

image

00041275:
add doc for sound #3018 

close #3018


shahargl:
@00041275 thanks! you should add it also in `mint.json` so it will show on docs

00041275:
> @00041275 thanks! you should add it also in `mint.json` so it will show on docs

done

shahargl (approved):
lgtm 👑

Issue #3093, opened 2025-01-22
[➕ Feature]: Support Basic Auth while adding victoria metrics providers


sushovan23:
**Is your feature request related to a problem? Please describe.**
Right now victoria metrics provider form doesn't have option to provide basic authentication details
**Describe the solution you'd like**
Update form to include basic auth options

**Describe alternatives you've considered**

**Additional context**

PR #3098, opened 2025-01-23
feat(victoriametrics): support both vmalert and vmmetrics

shahargl:
close https://github.com/keephq/keep/issues/3093


talboren (approved):
lgtm

PR #3110, opened 2025-01-24
docs(fix): error of missing highline code mark

congto:
Closes https://github.com/keephq/keep/issues/3112

## 📑 Description
Fix error of missing highline code mark


talboren:
@congto we just need a respective issue for this and it’s good to go

PR #3299, opened 2025-02-03
feat: clickhouse and vllm

Matvey-Kuk:
Added http and https to clickhouse, and vLLM provider with an example workflow.

Closes https://github.com/keephq/keep/issues/3300


docs/providers/documentation/vllm-provider.mdx
```diff
@@ -0,0 +1,58 @@
...
+        enum:
+          - production
+          - debug
+          - pre-prod
+      impacted_customer_name:
+        type: string
+    required:
+      - environment
+      - impacted_customer_name
+```
```
talboren, line 58: This is a bit vague. Explain that it's part of workflow? otherwise, there's nothing I can really do with it as a user

VladimirFilonov (approved):
Nice!

PR #3326, opened 2025-02-05
feat: manual topology manipulation

### LINKED ISSUE #2904: [➕ Feature]: Manual creation of topology
As a user, I would like to manually be able to add services and create connection with Keep's topology feature

We imagine the Topology map to be dynamic, allowing users to add new components and drawing the connections between them.
If the Topology item is fetched from 3rd party provider, it should be stated on object itself and it should not be able to edit it but only to create new connections.

### LINKED ISSUE #3328: [➕ Feature]: Topology (services + application) export import from YAML
As a user, I would like to export/import the topology (application/services) as YAML file.

35C4n0r:
Closes #2904
Closes #3328

## ✅ Checks


shahargl (commented):
are there any migrations needed? if you have service topology map exists, do I need to change from int to str or smth? did you check ti?

keep-ui/app/(keep)/topology/topology-client.tsx
```diff
@@ -42,47 +42,49 @@ export function TopologyPageClient({
...
-            topologyServices={topologyServices}
-            isVisible={tabIndex === 0}
-          />
-        </TabPanel>
-        <TabPanel className="flex-1">
-          <ApplicationsList applications={applications} />
-        </TabPanel>
-      </TabPanels>
-    </TabGroup>
+    <>
```
talboren, line 45: is that needed?

talboren (changes requested):
@35C4n0r can we add unit tests and perhaps an e2e test to all the functionality you added? also - you mentioned import/export, is it in this PR?

talboren (approved):
LGTM!

I didn't review in depth but I played with it and the experience was great. I fixed a small issue not related with this PR and pushed it.

shahargl (approved):
lgtm

PR #3336, opened 2025-02-06
fix(ui): saving workflow doesn't refresh cache

shahargl:
close https://github.com/keephq/keep/issues/3335

### DIFF
--- a/keep-ui/entities/workflows/model/useWorkflowActions.ts
+++ b/keep-ui/entities/workflows/model/useWorkflowActions.ts
@@ -1,11 +1,11 @@
 import { useApi } from "@/shared/lib/hooks/useApi";
 import { showSuccessToast } from "@/shared/ui/utils/showSuccessToast";
-import { useRevalidateMultiple } from "@/shared/lib/state-utils";
 import { showErrorToast } from "@/shared/ui";
 import { Definition } from "@/app/(keep)/workflows/builder/builder-store";
 import { getWorkflowFromDefinition } from "@/app/(keep)/workflows/builder/utils";
 import { stringify } from "yaml";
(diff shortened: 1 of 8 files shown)


Matvey-Kuk (approved):

talboren (approved):

keep-ui/shared/ui/YAMLCodeblock/ui/MonacoYAMLEditor.tsx
```diff
@@ -444,7 +446,7 @@ const MonacoYAMLEditor = ({
         workflow: Record<string, unknown>;
       };
       // update workflow
-      await updateWorkflow(workflowId!, parsedYaml);
+      await updateWorkflow(workflowId, parsedYaml, { mutateWorkflowDetail });
```
Kiryous, line 449: it's better not to pass around `mutateWorkflowDetail`, but instead call mutateMultiple with appropriate key `workflows/${workflowId}` in useWorkflowActions()

keep-ui/entities/workflows/model/useWorkflowActions.ts
```diff
@@ -45,20 +49,21 @@ export function useWorkflowActions(): UseWorkflowActionsReturn {
           }
         );
         showSuccessToast("Workflow created successfully");
-        refreshWorkflows();
+        await mutateWorkflows();
```
Kiryous, line 52: no need to await here, better to return response sooner imo

Kiryous (changes requested):
a few comments

Kiryous (approved):
LGTM! Thanks for the edits

PR #3391, opened 2025-02-10
fix: timeouts for workflows

Matvey-Kuk:
If worker dies during the workflow execution, or it got stuck in "forever running state" for some other reason, record an error: 
Screenshot 2025-02-10 at 17 55 18

Closes https://github.com/keephq/keep/issues/3392

### DIFF
--- a/keep/providers/clickhouse_provider/clickhouse_provider.py
+++ b/keep/providers/clickhouse_provider/clickhouse_provider.py
@@ -14,7 +14,7 @@ from keep.providers.models.provider_config import ProviderConfig, ProviderScope
 from keep.validation.fields import NoSchemeUrl, UrlPort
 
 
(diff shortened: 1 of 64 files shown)


keep/providers/clickhouse_provider/clickhouse_provider.py
```diff
@@ -14,15 +14,17 @@
 from keep.validation.fields import NoSchemeUrl, UrlPort
 
 
+DEFAULT_TIMEOUT = 120  # Not to hang the thread forever, only for extreme cases
```
talboren, line 17: that's seconds, right?

Matvey-Kuk: Yep!

talboren (approved):
LGTM

PR #3421, opened 2025-02-12
fix: Mapping clarification

Matvey-Kuk:
None

close https://github.com/keephq/keep/issues/3422


talboren:
I don't know why it's failing, I'll take some of the stuff here - thanks!

PR #3424, opened 2025-02-13
fix: wf timeouting db load

Matvey-Kuk:


keep/api/core/db.py
```diff
@@ -240,7 +241,8 @@ def get_timeouted_workflow_exections():
                 select(WorkflowExecution)
                 .filter(WorkflowExecution.status == "in_progress")
                 .filter(
-                    WorkflowExecution.started <= datetime.utcnow() - WORKFLOWS_TIMEOUT
+                    WorkflowExecution.started <= datetime.utcnow() - WORKFLOWS_TIMEOUT,
+                    WorkflowExecution.started >= datetime.utcnow() - CUTOUT_TO_IGNORE_TOO_OLD_WORKFLOWS,
```
talboren, line 245: i don't think it'll solve the problem because it still needs to scan all statuses, that doesn't have index :(

PR #3456, opened 2025-02-14
fix(ui): Merge incidents not working in some cases

### LINKED ISSUE #3370: [🐛 Bug]: Merge incidents not working in some cases
While the UI shows success:

Image

The actual response from the backend is:

Image

nayan-bagale:
/claim #3370
Closes #3370

## ✅ Checks

## ℹ Additional Information

Before:
https://github.com/user-attachments/assets/6f564086-8e43-43b6-85e3-5247f4b0c6f8

After:
https://github.com/user-attachments/assets/4a45153c-a96d-4a04-82df-d0786fad2005

---------------
I've replaced `toast.success` with toast.info and passed `result.message` as an argument since the response contains a different message. A success toast isn't necessary in this case.
image

### DIFF
--- a/keep-ui/entities/incidents/model/useIncidentActions.tsx
+++ b/keep-ui/entities/incidents/model/useIncidentActions.tsx
@@ -166,7 +166,7 @@ export function useIncidentActions(): UseIncidentActionsValue {
           source_incident_ids: sourceIncidents.map((incident) => incident.id),
           destination_incident_id: destinationIncident.id,
         });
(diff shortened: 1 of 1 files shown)


talboren:
@nayan-bagale thank you for this contribution, but this was not the idea. I could've changed the toast to something else as well but what we want is to actually overcome the problem. There's another PR opened for that https://github.com/keephq/keep/pull/3457

PR #3472, opened 2025-02-15
fix: Restrict incident creation to firing alerts only

### LINKED ISSUE #3418: [🐛 Bug]: Correlation rule should not create incident if alert is not firing
Correlation rules should *create* the incident only if the alert (or all alerts in case all conditions met is marked, see screenshot) is firing.
**After the incident was created, all alerts, firing and resolved should be correlated to that incident** -- I'm only speaking about incident creation part.

Image

VladimirFilonov:
This update ensures that new incidents are only created for alerts with a "firing" status, preventing unnecessary incidents for resolved or non-firing alerts. A corresponding test was added to validate this behavior in the rules engine.

Closes #3418

## ✅ Checks

### DIFF
--- a/keep/rulesengine/rulesengine.py
+++ b/keep/rulesengine/rulesengine.py
@@ -127,53 +127,54 @@ class RulesEngine:
                         session=session,
                         event=event,
                     )
-                    incident = assign_alert_to_incident(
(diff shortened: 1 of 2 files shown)


talboren (approved):
LGTM

Issue #3483, opened 2025-02-16
[🐛 Bug]: checkmk different tz


shahargl:

VladimirFilonov:
Maybe worth using `dateutil.parser.parse` which tries to identify format itself?

PR #3484, opened 2025-02-16
fix: checkmk wrong format

shahargl:
CleanShot 2025-02-16 at 09 43 11@2x

close https://github.com/keephq/keep/issues/3483


talboren (approved):
LGTM

PR #3584, opened 2025-02-19
feat: better sign in error page

shahargl:
before:
CleanShot 2025-02-19 at 18 57 28@2x

after:
CleanShot 2025-02-19 at 18 57 40@2x

close https://github.com/keephq/keep/issues/3583


talboren (approved):
lgtm

PR #3594, opened 2025-02-19
feat: AWS EC2 IAM

Matvey-Kuk:
Leverage EC2 attached IAM for AWS integration. Following this instruction:

First, create an IAM policy:

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:GetObject",
                "s3:PutObject",
                "s3:ListBucket",
                "s3:DeleteObject"
            ],
            "Resource": [
                "arn:aws:s3:::non-public-keep",
                "arn:aws:s3:::non-public-keep/*"
            ]
        }
    ]
}
```
Steps to create and attach the role:

Go to IAM in AWS Console
Click "Roles" → "Create role"
Select "AWS service" as trusted entity
Choose "EC2" as use case
Attach the policy above (you'll need to create it first)
Name the role (e.g., "EC2-S3-NonPublicKeep-Access")

To attach the role to your EC2 instance:

Go to EC2 in AWS Console
Select your instance
Click "Actions" → "Security" → "Modify IAM role"
Select the role you created
Click "Save"

To verify it works, you can test with this Python code on your EC2:
```
import boto3

s3 = boto3.client('s3', region_name='eu-north-1')

# List objects in bucket
response = s3.list_objects_v2(Bucket='non-public-keep')
for obj in response.get('Contents', []):
    print(obj['Key'])
```

### DIFF
--- a/keep/providers/amazonsqs_provider/amazonsqs_provider.py
+++ b/keep/providers/amazonsqs_provider/amazonsqs_provider.py
@@ -27,15 +27,15 @@ class AmazonsqsProviderAuthConfig:
 
     access_key_id: str = dataclasses.field(
         metadata={
-            "required": True,
-            "description": "Access Key Id",
+            "required": False,
+            "description": "Access Key Id (Leave empty if using IAM role at EC2)",
             "hint": "Access Key ID",
         },
     )
     secret_access_key: str = dataclasses.field(
         metadata={
-            "required": True,
-            "description": "Secret access key",
+            "required": False,
+            "description": "Secret access key (Leave empty if using IAM role at EC2)",
             "hint": "Secret access key",
             "sensitive": True,
         },
--- a/keep/providers/cloudwatch_provider/cloudwatch_provider.py
+++ b/keep/providers/cloudwatch_provider/cloudwatch_provider.py
@@ -24,12 +24,16 @@ from keep.providers.models.provider_config import ProviderConfig, ProviderScope
 @pydantic.dataclasses.dataclass
 class CloudwatchProviderAuthConfig:
     access_key: str = dataclasses.field(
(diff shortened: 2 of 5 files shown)


keep/providers/amazonsqs_provider/amazonsqs_provider.py
```diff
@@ -27,15 +27,15 @@ class AmazonsqsProviderAuthConfig:
 
     access_key_id: str = dataclasses.field(
         metadata={
-            "required": True,
-            "description": "Access Key Id",
+            "required": False,
+            "description": "Access Key Id (Leave empty if using IAM role at EC2)",
```
talboren, line 31: Doesn’t it require default because the annotation is str?

talboren (approved):
Nice! That was everything required?
A note worth mentioning: if we ever deploy Keep managed on AWS, this will be a security issue

Matvey-Kuk:
@talboren that shouldn't be a sec issue since the IAM should be explicitly attached to the container. Actually any app on the EC2 could make a request to the internal AWS API and get current container's IAM :)

Matvey-Kuk:
Screenshot 2025-02-20 at 9 58 30
like this

talboren:
> @talboren that shouldn't be a sec issue since the IAM should be explicitly attached to the instance. Actually any app on the EC2 could make a request to the internal AWS API and get current instance's IAM :)

In a multi tenant app, this will be an issue

keep/providers/amazonsqs_provider/amazonsqs_provider.py
Matvey-Kuk: fixed!

skynetigor (approved):
LGTM

PR #3633, opened 2025-02-23
fix: boto3 auth

Matvey-Kuk:
close https://github.com/keephq/keep/issues/3634


shahargl:
If all these providers needs the same code for authentication why doesn’t they inherit from some AWSProvider?

Matvey-Kuk:
I wonder if validate_scopes could be unified somehow. Btw this PR is an issue fix @shahargl

shahargl:
Yea but a simple fix that requires changing the same code at 5 different places is prune to errors. I guess we will add eventually more AWS services so it will be even more

skynetigor (approved):
LGTM

Issue #3048, opened 2025-01-17
[🐛 Bug]: deduplication rules


00041275:
#3017  not competed.
for UI page /deduplication, I haven't button "Create Deduplication Rule", but in playground I can see it.

https://platform.keephq.dev/
![Image](https://github.com/user-attachments/assets/e67137ad-8380-455d-a534-f9b4720f73a9)

my view
![Image](https://github.com/user-attachments/assets/a0b6dcd6-b23d-401f-9639-698246aff58e)

@shahargl @Matvey-Kuk

Matvey-Kuk:
Your version of Keep is not up-2-date, on the current version there is a text "This page will become active once the first alerts are registered."

https://github.com/keephq/keep/pull/3044/files#diff-68d38de3340592b1d67c3df027877456ac5461a1b579f5563815a30329b47216R21

00041275:
> Your version of Keep is not up-2-date, on the current version there is a text "This page will become active once the first alerts are registered."
> 
> https://github.com/keephq/keep/pull/3044/files#diff-68d38de3340592b1d67c3df027877456ac5461a1b579f5563815a30329b47216R21

but I have thousands alerts and few incidents, please look to screen

00041275:
![Image](https://github.com/user-attachments/assets/ac7d1398-fd90-442c-8f66-a0e3ee55e0b1)

shahargl:
@00041275 can you send the output of your console? any errors there? Also can you show the request/response to /deduplications please?

00041275:
> [@00041275](https://github.com/00041275) can you send the output of your console? any errors there? Also can you show the request/response to /deduplications please?

Find it. Problem was havent files keep/.env and /.keep.yaml. 
I used OKD and I can set environment from kubernetes env yaml.
I think this files not required for Docker, Kubernetes deployment.

shahargl:
Amazing! Closing the issue as “solved”

00041275:
> Amazing! Closing the issue as “solved”

why they need for cloud /.env and /.keep.yaml? 
may be we can add warning for required filed to deployment in there https://docs.keephq.dev/deployment/configuration?

PR #3795, opened 2025-03-03
Update configuration.mdx

00041275:
#3048 added warning for config files


talboren:
@00041275 can you open the respective issue for this PR? closing it for now until we have something.
Also, you can use `<Tip></Tip>` which will look better

00041275:
> <Tip>

done

PR #3839, opened 2025-03-04
fix: Workflow manual run modal improvements

### LINKED ISSUE #3619: [🐛 Bug]: Run Workflow Modal Problems
1. If you have multiple workflows with the same name, there's no way to identify the difference between the two. (For example a workflow called X and a workflow called X will show 2 X in the modal select box).
2. Select box is not searchable, which makes it hard to find a specific workflow if you have many workflows
3. _Disabled workflows may appear in the select list (although they shouldn't)_ (think I fixed it)

VladimirFilonov:
Replaced ID-based selection with a more intuitive object-based approach using custom React-Select components. Added support for improved filtering, display of workflow details, and trigger badges in the dropdown. Simplifies interaction and enhances user experience.

Closes #3619 

## 📑 Description
image

## ✅ Checks

### DIFF
--- a/keep-ui/app/(keep)/workflows/manual-run-workflow-modal.tsx
+++ b/keep-ui/app/(keep)/workflows/manual-run-workflow-modal.tsx
@@ -1,4 +1,4 @@
-import { Badge, Button, Text, Title } from "@tremor/react";
+import { Button, Text, Title } from "@tremor/react";
 
 import Modal from "@/components/ui/Modal";
 import { useWorkflows } from "utils/hooks/useWorkflows";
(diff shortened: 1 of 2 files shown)


shahargl:
nice!

shahargl (dismissed):
lgtm

keep-ui/app/(keep)/workflows/manual-run-workflow-modal.tsx
```diff
@@ -8,7 +8,11 @@ import { useRouter } from "next/navigation";
 import { IncidentDto } from "@/entities/incidents/model";
 import { AlertDto } from "@/entities/alerts/model";
 import { useApi } from "@/shared/lib/hooks/useApi";
-import { showErrorToast } from "@/shared/ui";
+import { Select, showErrorToast } from "@/shared/ui";
+import { Trigger, Workflow } from "@/shared/api/workflows";
+import { components, OptionProps } from "react-select";
```
shahargl, line 13: I think we already have Select in ui/components?

VladimirFilonov: Yes, we have already used this one in the incidents. In fact, this is a wrapper around react-select.

shahargl (approved):
lgtm

PR #3845, opened 2025-03-04
fix: alert menu

shahargl:
close https://github.com/keephq/keep/issues/3844

before:
CleanShot 2025-03-04 at 15 51 13

after:
CleanShot 2025-03-04 at 15 51 38


talboren (approved):
LGTM

PR #3854, opened 2025-03-05
fix: provider docs links

### LINKED ISSUE #3768: [🐛 Bug]: Dead end when users struggle with the provider installation
A few times I noticed that users don't proceed to docs while struggling with the provider installation. I think it's a UX bug and docs link should be more understandable.

Image

Matvey-Kuk:
close #3768 

Before:
Screenshot 2025-03-05 at 9 40 54

After:
Screenshot 2025-03-05 at 9 40 40


skynetigor (approved):
LGTM

PR #3900, opened 2025-03-06
feat: incident link from alert table

talboren:
Not sure how much will it impact performance (added `with_incident=true`) back.

https://github.com/user-attachments/assets/db9ba4ef-ee04-47c9-b400-454a4f952a30


shahargl (approved):
lgtm

PR #3901, opened 2025-03-06
fix: frontend-side variables' validation

### LINKED ISSUE #3783: [🐛 Bug]: validate "actions.<action-id>.results" can be accessed
**Describe the bug**
Validate that actions are in order (for example, if I use {{actions.slack-action.results}} in ntfy-action, it'll order properly. (i think today it break the wf)

Kiryous:
Closes #3783  

https://github.com/user-attachments/assets/623c27bd-71a6-4815-87b3-30342202552a


VladimirFilonov (approved):
LGTM

PR #3910, opened 2025-03-07
fix(ui): correlation rule fields

### LINKED ISSUE #3909: [🐛 Bug]: correlation rule fields not showing

talboren:
close #3909 

Before (only default fields):

CleanShot 2025-03-07 at 09 45 05@2x

After:

CleanShot 2025-03-07 at 09 44 38@2x


shahargl (approved):
Lgtm

PR #3949, opened 2025-03-10
fix: cli installation instructions

### LINKED ISSUE #3495: [🐛 Bug]: Keep CLI installation instruction is missing important nuances
1) Were helping a prospect and realized that the fresh mac (even with python) doesn't have pip, we're missing venv part of the instruction etc. Prospect who is not familiar with python felt lost.

2) Docker instruction doesn't link the local config file so the usability is also pretty tricky.

https://docs.keephq.dev/cli/installation

Going to update, taking this task on me.

Matvey-Kuk:
Tested on fresh mac:

Screenshot 2025-03-10 at 13 26 08

close https://github.com/keephq/keep/issues/3950
close #3495

### DIFF
--- a/docs/cli/installation.mdx
+++ b/docs/cli/installation.mdx
@@ -22,7 +22,10 @@ git clone https://github.com/keephq/keep.git && cd keep
 Install Keep CLI with `pip`:
 
 ```shell
-pip install .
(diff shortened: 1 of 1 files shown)


skynetigor (approved):
LGTM

PR #3954, opened 2025-03-10
fix: better explanation for provider installation with missing scopes

### LINKED ISSUE #2379: [🐛 Bug]: Provider should not be installed if scope validation fails
If scopes validation failed, the provider should not be installed.

Matvey-Kuk:
Noticed a customer who installed a provider with tons of scopes, some of those scopes were valid, others were not. Customer got confused and thought that provider is not installed. Added a panel.

Screenshot 2025-03-10 at 18 09 38

outdated but related:
close #2379


skynetigor (approved):
LGTM

PR #4058, opened 2025-03-16
fix: incident merging

Matvey-Kuk:
close [#3370](https://github.com/keephq/keep/issues/3370)


keep/api/routes/incidents.py
```diff
@@ -517,14 +517,12 @@ def merge_incidents(
         else:
             message = f"{pluralize(len(merged_ids), 'incident')} merged into {command.destination_incident_id} successfully"
 
-        if skipped_ids:
-            message += f", {pluralize(len(skipped_ids), 'incident')} were skipped"
         if failed_ids:
             message += f", {pluralize(len(failed_ids), 'incident')} failed to merge"
+            raise HTTPException(f"Some incidents failed to merge. {message}")
```
talboren, line 522: what will be with `skipped_incident_ids`?

talboren (commented):
What's the outcome of this PR, I don't understand what will happen with those skipped incidents?

Matvey-Kuk:
@talboren I removed skipped incidents, it will fail only on the case of exception. This PR will allow merging incidents without alerts & will fix the case when UI says incidents are merged but they are not.

talboren (approved):
LGTM

Issue #3699, opened 2025-02-25
[🐛 Bug]: Markup oddities in the incident page


Matvey-Kuk:
A few places where markup is odd:

<img width="1170" alt="Image" src="https://github.com/user-attachments/assets/ed363d70-5d2e-4e26-80b8-1fbd9484a300" />

PR #4063, opened 2025-03-16
fix: minor oddities on the incident page

Matvey-Kuk:
close [#3699](https://github.com/keephq/keep/issues/3699)


talboren:
@Matvey-Kuk can you add before/after?

Issue #3960, opened 2025-03-11
[🔌 Provider]: Nagios Provider


talboren:
https://www.nagios.org/

Matvey-Kuk:
@Ayush9026 it's yours :)

Matvey-Kuk:
Hey @Ayush9026, just checking if you have something to show

Matvey-Kuk:
@Ayush9026 hi! I see the PR is stale. Are you still considering working on this issue? :)

DMMutua:
Hi @Matvey-Kuk. I'd like to be assigned and /attempt #3960 this as Initial PR Remains stale. I'd raise a PR in hours cc @talboren

DMMutua:
hello @Matvey-Kuk. I'd request for review on a pr #5149 that will add the relevant provider and close this issue cc @talboren

Harmatta:
/attempt #3960

Submitted PR #5639 implementing a full Nagios XI provider with:
- Pull-based alert ingestion via Nagios XI REST API (servicestatus + hoststatus)
- Webhook-based push ingestion for Nagios notification commands
- Problem acknowledgement support
- Complete severity/status mapping for all Nagios states
- README with setup instructions

kdn-posipaka:
I've submitted a PR implementing the Nagios provider: #5654

The implementation follows the same webhook-based pattern as existing providers (Checkmk, Zabbix). It handles both host and service notifications, maps all Nagios states/notification types to Keep's severity/status model, and includes detailed webhook setup documentation.

tianzhuxhc-bit:
Hi maintainers, I'd like to work on this bounty.

Plan:
1) Reproduce and align scope quickly
2) Submit an initial PR with tests/docs
3) Iterate fast based on review feedback

I can post first progress within 12 hours.

Jujubee-LLM:
/attempt #3960

Submitted PR #5675 implementing Nagios provider support with:
- Webhook-based ingestion for Nagios notifications (Core/XI macro variants)
- Host/service state -> severity/status mapping
- Notification type handling (PROBLEM/RECOVERY/ACKNOWLEDGEMENT/downtime/flapping)
- Added docs + mock payload + unit tests

justinleeyang:
I can take bounty #3960. I will share a scoped implementation plan and first PR shortly.

justinleeyang:
Quick follow-up on #3960: I am still actively working on this. Please let me know if you want me to narrow scope to a smaller first PR.

[...]

PR #4091, opened 2025-03-18
feat: Implemented Nagios Provider (#3960)

deekshatomer:
/claim #3960
Closes #3960

## Summary  
Implemented the Nagios Provider for Keep.  

Description:
-Added a new class NagiosProvider to integrate with Nagios.
-Implemented the fetch_alerts method to retrieve alerts from Nagios using HTTP requests.
-Created a unit test file (test_nagios_provider.py) to verify the functionality of the NagiosProvider class.
-Used unittest.mock to mock the fetch_alerts method in the test to avoid real HTTP requests.
-Verified that the test passes successfully.

Testing:
Ran the test file test_nagios_provider.py using pytest, and it passed without any issues.
Confirmed that the implementation works as expected with mocked responses.
/claim #3960

### DIFF
--- a/providers/nagios_provider.py
+++ b/providers/nagios_provider.py
@@ -1,0 +1,25 @@
+# filepath: /providers/nagios_provider.py
+import requests
+
+class NagiosProvider:
+    def __init__(self, base_url, api_key):
+        """
+        Initialize the Nagios Provider with the base URL and API key.
+        """
+        self.base_url = base_url
+        self.api_key = api_key
+
+    def fetch_alerts(self):
+        """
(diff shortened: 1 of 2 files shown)


deekshatomer:
hi ! @Matvey-Kuk  Could you please review this PR? Thanks!

Matvey-Kuk (changes requested):
I see only tests in the proposed changes, not the provider itself 😢

deekshatomer:
@Matvey-Kuk I have added the missing NagiosProvider class. Please review.

PR #4114, opened 2025-03-19
fix: snow topology

Matvey-Kuk:
close https://github.com/keephq/keep-private/issues/33

close https://github.com/keephq/keep/issues/4132


skynetigor (approved):
LGTM

PR #4127, opened 2025-03-19
fix: navbar in smaller screens is broken

talboren:
fixed cmd + k and:

https://github.com/user-attachments/assets/78f99b42-14fa-4232-99b7-b109d86bd4f9


shahargl (approved):
lgtm

PR #4166, opened 2025-03-22
[draft] feat: improve performance of mapping

### LINKED ISSUE #4165: [🐛 Bug]: Mapping performance improvements

talboren:
close #4165 

Improves the performance of mapping by moving the logic to the DB side.
Since `rows` is just a simple "table" (e.g., a CSV file), we can flatten the JSON column with DB functions (i.e., JSON_TABLE, or 
```sql
WITH flattened AS (
                SELECT
                    value AS row_data
                FROM mappingrule, json_each(mappingrule.rows)
                WHERE mappingrule.id = :rule_id
            )
            SELECT row_data FROM flattened
```
in the case of SQLite, and run the respective query in the DB.

In cases where we have a lot of rows (>50K), it's much more performant then loading everything in memory on every event..

TBD:
- [ ] Write tests for MySQL
- [ ] Write tests for SQlite
- [ ] Write tests for PG 

Useful: https://docs.sqlalchemy.org/en/14/orm/loading_columns.html#deferred-column-loading


talboren:
@skynetigor i'm not sure if cel_to_sql is the right place to put it, but it kinda reminded me with what you're doing with formatting the query to work on the DB. That's why I put it over there.
Would love your review btw, because I think you have a lot of context about formatting queries ;)

VladimirFilonov:
@talboren crazy question - don't we want to convert CSV with table on upload, but not on runtime?

skynetigor:
@talboren Great idea about utilizing DB for this 👍
To help keep the folder structure clean and organized, I believe this file aligns more with the mapping rule business logic rather than being a generic utility.
Since CEL to SQL is a generic mechanism that can be reused for querying or faceting data across the DB, I’d suggest placing mapping_rule_matcher closer to the **bl** folder instead.

talboren:
Still working on it! Will try to finalize next week.

PR #4185, opened 2025-03-23
feat: add provider name query parameter in event handling

### LINKED ISSUE #4184: [➕ Feature]: Add query parameter provider_name in alert receive endpoint
Currently, we are trying to provision the Keep providers by using the environment variable `KEEP_PROVIDERS` , each provider will have its own `id` that is generated via UUID.

We need the ids of providers to configure the query parameter `provider_id` of the **Receive Event** endpoint (for referencing the correct provider, assume we have many provider instances with the same provider type): https://docs.keephq.dev/api-ref/alerts/receive-event.

But we think that we should not configure the id manual and should let the system decide it. Of course we can get the `id` after provisioning the provider but imagine the Keep service is somehow destroyed and lost all data, after re-deploying the ids will be different, we need to re-fetch the ids and change it manual.

Therefore, we are thinking of modifying the query parameter provider_id to provider_name as provider_name is unique within each tenant.

```python
class Provider(SQLModel, table=True):
    __table_args__ = (UniqueConstraint("tenant_id", "name"),)
```

This way, we can provision the provider and use its name in the **Receive Event** endpoint instead of the id.

tuantran0910:
Closes #4184

## 📑 Description

Add provider name query parameter in event handling.

## ✅ Checks

## ℹ Additional Information

I have enhanced UI when displaying provider icon along with the provider's name a little bit, just add more spacing between them :D I don't know whether I am able include this tiny changes in pull request. If not, just let me know and I will delete it and migrate into new separate pull request.

**Before:**

image

**After:**

image


shahargl (approved):
lgtm

shahargl (approved):
lgtm

Issue #4005, opened 2025-03-13
[🐛 Bug]: muted alert called workflow


00041275:
workflow started for maintenance muted alert with setting "Alerts will not show in feed" = true

![Image](https://github.com/user-attachments/assets/f48f00c1-ae39-4b1e-a997-40e14d1b55fe)
![Image](https://github.com/user-attachments/assets/a4a13b64-1063-48ac-a39a-4e76cf5b9863)
![Image](https://github.com/user-attachments/assets/7fd21797-a4ac-4011-af98-f8e00ada6a79)

Matvey-Kuk:
The user who is willing to launch workflows for alerts with a specific status could leverage the filter by status.
Preventing workflow launch for "dismissed" alerts will bring too much magic, so, for now, I guess we'll leave the feature behavior "as is".

There is some related discussion in the PR with the bugfix proposal: https://github.com/keephq/keep/pull/4188

Issue #3858, opened 2025-03-05
[🐛 Bug]: AI summary doing nothings


VladimirFilonov:
Cutout from https://github.com/keephq/keep/issues/3822

![Image](https://github.com/user-attachments/assets/32568307-efe1-4983-8012-3d532db4f91f)

Possible problem: is OpenAI key is not valid, request failed, but there is no any feedback from the system

PR #4188, opened 2025-03-23
[draft] fix: running workflows for suppressed alerts

Matvey-Kuk:
It's pretty confusing when we run Workflows for suppressed alerts: https://github.com/keephq/keep/issues/4005 

I don't like idea of just not allowing to run workflows for suppressed alerts at all, so I allowed to run workflows for them if suppressed status is mentioned in the filter.

close https://github.com/keephq/keep/issues/4005
close https://github.com/keephq/keep/issues/3858


keep/workflowmanager/workflowmanager.py
```diff
@@ -183,6 +183,19 @@ def insert_events(self, tenant_id, events: typing.List[AlertDto | IncidentDto]):
                     if not trigger.get("type") == "alert":
                         self.logger.debug("trigger type is not alert, skipping")
                         continue
+                    # Skipping suppressed alerts unless it's explicitly mentioned in the filter
+                    if event.status == "suppressed":
```
shahargl, line 187:
```
AlertStatus.SUPPRESSED
```

keep/workflowmanager/workflowmanager.py
```diff
@@ -183,6 +183,19 @@ def insert_events(self, tenant_id, events: typing.List[AlertDto | IncidentDto]):
...
+                        if trigger.get("filters", []) == [
+                            {"key": "status", "value": "suppressed"}
+                        ]:
+                            self.logger.debug(
```
shahargl, line 191: should be `info`

Matvey-Kuk:
@shahargl in general, how do you think if we should avoid running workflows for suppressed alerts? I wonder if we shouldn't run them by default also for `resolved` if it's not mentioned in the filter.

shahargl:
@Matvey-Kuk idk - is there any context for this? did someone ask for it? because if not I would probably keep things simple

Matvey-Kuk:
@shahargl it's based on https://github.com/keephq/keep/issues/4005

talboren:
> @shahargl it's based on #4005

> @shahargl in general, how do you think if we should avoid running workflows for suppressed alerts? I wonder if we shouldn't run them by default also for `resolved` if it's not mentioned in the filter.

i think that in general it should be configurable, but I would take the approach where you can specify which statuses you ignore, instead of marking those you want to execute for.
So workflows run for every event, based on filters, and you can exclude certain statuses (it's already possible, btw, by filtering out)

```yaml
  triggers:
    - type: alert
      filters:
        - key: status
          value: r"(firing|acknowledged)"
```

PR #4240, opened 2025-03-27
fix: validate agains loose workflow schema before syncing yaml editor

### LINKED ISSUE #3676: [🐛 Bug]: You can test-run AI-created workflows without saving but not YAML-created
When you create a new workflow using AI, you could test-run it. When you want to create a new workflow and test-run it via YAML, you can't. I think it's a bug.

Image

Kiryous:
Closes #3676  

## Before
Changes are not saved if the workflow missing some props

https://github.com/user-attachments/assets/8d34ed10-745d-4cef-8de4-e97a74291f9c

## After
Changes saved if something is missing but the general structure is correct

https://github.com/user-attachments/assets/bf44b756-7bb2-4093-92f9-77ad6091f397


shahargl (approved):
lgtm

PR #4242, opened 2025-03-27
fix: rm excessive "+ New" button from workflow builder

### LINKED ISSUE #3677: [🐛 Bug]: "+ New" button is odd in the workflow aditor
The first issue is that I can't understand logics: when do we show it and when we don't. If we create workflow from example, it's not visible:

Image

When we create workflow from scratch, it is:

Image

When I click it, it's prompting some meaningless validation. It's hard to understand "what will happen". Does it mean that it will drop the current unsaved changes? Or it will be saved and I will go to the new workflow page?

Image

One more popup following:

Image

In general I think this button is excessive. I propose to remove it. It's complicating the scenario and the UI.

Kiryous:
Closes #3677  

## Before

image

## After

No more confusing button and Load button renamed to more clean "Import from YAML"

Screenshot 2025-03-27 at 12 37 05

### DIFF
--- a/keep-ui/widgets/workflow-builder/workflow-builder-widget.tsx
+++ b/keep-ui/widgets/workflow-builder/workflow-builder-widget.tsx
@@ -6,7 +6,6 @@ import {
   ArrowUpOnSquareIcon,
   PencilIcon,
   PlayIcon,
-  PlusIcon,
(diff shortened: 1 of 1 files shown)


VladimirFilonov (approved):
LGTM

PR #4260, opened 2025-03-27
feat: add opensearch serverless provider

### LINKED ISSUE #4228: [🔌 Provider]: AWS Opensearch with Serverless
**Describe the provider you want to add**
AWS Opensearch provider and Opensearch Serverless provider. Opensearch Serverless can use IAM roles to authenticate.
https://docs.aws.amazon.com/opensearch-service/latest/developerguide/security-iam-serverless.html
We are already using cloudwatch provider with an IAM role attached to the keep service accounts which provides the needed access. We would like to have the same possibility with this Opensearch provider. 

**Describe your use case**
This will help us integrate Opensearch AWS into our keep which will help us correlate more alerts and logs.

**Are you already using Keep?**
Yes

**Additional context**
If we could us this provider the same way we use cloudwatch provider with an IAM role that would be great

35C4n0r:
Closes #4228 

## ✅ Checks


keep/providers/opensearchserverless_provider/opensearchserverless_provider.py
```diff
@@ -0,0 +1,209 @@
...
+            res = self.__get_doc(test_index, test_doc_id)
+            if res.status_code == 403:
+                self.logger.error("No permission for aoss:ReadDocument")
+            else:
+                scopes["aoss:ReadDocument"] = True
+                self.logger.info("Read permission validated successfully.")
+        except Exception as e:
+            self.logger.error("Error while testing aoss:ReadDocument", extra={"exception": str(e)})
+
+        return scopes
```
talboren, line 134: This reminds me of the problem we previously had with OpsGenie where we created a test alert to test scopes.
I think it's a bad practice and we should try hard to see if there's another way to validate the scopes the user has without creating/deleting anything.
For example, one may give you access to create a doc but not to delete it, and we will fail because of that, though we do have the access to create the doc.

talboren (changes requested):
@35C4n0r see my comment 🙏🏼

shahargl (changes requested):
1. why do we need a new library? you can use boto3  like other aws services - https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/opensearch.html
2. can you add a readme of how to use?
3. can you add a workflow using this?

shahargl (approved):
lgtm

VladimirFilonov (approved):

talboren (approved):

PR #4268, opened 2025-03-27
feat: show wf revision in the UI

### LINKED ISSUE #4266: [➕ Feature]: Make workflow revision visible in the UI

Kiryous:
Closes #4266  

Screenshot 2025-03-27 at 21 25 57


talboren:
> Member

@Kiryous I like that very much and I think it could've been super helpful if we could actually show previous revisions -- as far as I understand that now, it'll just show me that I edited this workflow 5 times but wouldn't really get me in to any action I can do with that.
Let's merge this (because it's beautiful ;)), and push to save older revisions in some table, where we can present them and presumably let users revert back to older revision. wdyt?

talboren (approved):
LGTM

PR #4272, opened 2025-03-28
feat: allow using latest providerId and alter UI display

### LINKED ISSUE #4271: [🐛 Bug]: ProviderId should be updated to the newest value
## Describe the Bug

When provisioning a provider and sending an alert with an incorrect `providerId` (one that does not belong to the provisioned provider), the alert gets permanently associated with the wrong provider.

If the same alert is sent again (with partial deduplication) using the correct `providerId`, it still retains the previously assigned incorrect `providerId`.

## Solution

Allow the `providerId` to be updated to the latest value.

tuantran0910:
Closes #4271

## 📑 Description

When provisioning a provider and sending an alert with an incorrect providerId (one that does not belong to the provisioned provider), the alert gets permanently associated with the wrong provider.

If the same alert is sent again (with partial deduplication) using the correct providerId, it still retains the previously assigned incorrect `providerId`. We should allow the providerId to be updated to the latest value.

## ✅ Checks

## ℹ Additional Information
Here’s a clearer and more polished version of your text:  

I have configured alerts to be sent to Keep via a generic webhook endpoint. For the provisioned provider **A** with ID `123`, I decided to add the query parameter `?provider_id=123`.  

On the **Provider** page, each provider card was incorrectly displaying the `id` field instead of the `name` field. I have updated it to display the `name` field correctly.

**Before**

image

**After**

image

Additionally, the spacing between the icon and the provider name was too tight, affecting readability. I have added some spacing for better clarity.  

**Before**

image

**After**

image

### DIFF
--- a/keep-ui/app/(keep)/alerts/alert-sidebar.tsx
+++ b/keep-ui/app/(keep)/alerts/alert-sidebar.tsx
@@ -128,9 +128,9 @@ const AlertSidebar = ({
                       providerType={alert.source![0]}
                       width={24}
                       height={24}
-                      className="inline-block w-6 h-6"
+                      className="inline-block w-6 h-6 mr-2"
                     />
-                    {providerName}
+                    <span>{providerName}</span>
                   </p>
                   <p>
                     <FieldHeader>Description</FieldHeader>
--- a/keep-ui/app/(keep)/providers/provider-tile.tsx
+++ b/keep-ui/app/(keep)/providers/provider-tile.tsx
@@ -231,7 +231,7 @@ export default function ProviderTile({ provider, onClick }: Props) {
 
             {provider.details && provider.details.name && (
               <Subtitle className="truncate">
-                id: {provider.details.name}
+                Name: {provider.details.name}
               </Subtitle>
             )}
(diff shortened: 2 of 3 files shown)


talboren:
@tuantran0910 thank you for the contribution. I have to admit that I'm missing the before/after difference. Can you point me to where can I see the changes? It seems like both are still displaying the id of the provider.

tuantran0910:
Hi @talboren, here is the difference, just rename `Id` to `Name` for clarity.

image

Actually, these are the name of the providers and the ID should be the UUID-like string like this :D

image

talboren (approved):
LGTM

tuantran0910:
Thanks @talboren for the approval :D Can you help me to see what happened with the E2E tests. It seems that I don't have permission for those tests right ?

talboren:
> Thanks @talboren for the approval :D Can you help me to see what happened with the E2E tests. It seems that I don't have permission for those tests right ?

I think it has something to do with a change in the e2e workflow we did, which caches and pulls images, and requires some other permissions now. I've re-run it and if it fails again, will fix and merge on Sunday.

Issue #4275, opened 2025-03-29
[🐛 Bug]: SQLAlchemy Warning – Implicit Coercion of SELECT Object


EnotShow:
We are encountering an **SQLAlchemy warning** in Keep:  

```
/venv/lib/python3.11/site-packages/keep/api/core/alerts.py:184: SAWarning: implicitly coercing SELECT object to scalar subquery; please use the .scalar_subquery() method to produce a scalar subquery.
```  

This suggests that a **SELECT object is being coerced into a scalar subquery implicitly**, which is deprecated or incorrect.  

#### Expected Behavior  
The code should explicitly use `.scalar_subquery()` when a scalar subquery is required, preventing this warning.  

#### Steps to Reproduce  
1. Run Keep.  
2. Observe logs for the above warning.  

#### Possible Solution  
- Modify the query at **alerts.py:184** to explicitly use `.scalar_subquery()`.  
- Identify where the SELECT object is being used incorrectly and refactor accordingly.

PR #4276, opened 2025-03-29
fix: Explicitly Use `.scalar_subquery()` to Prevent SQLAlchemy Warning  

EnotShow:
#### Description  
This pull request fixes the following **SQLAlchemy warning** in Keep:  

```
/venv/lib/python3.11/site-packages/keep/api/core/alerts.py:184: SAWarning: implicitly coercing SELECT object to scalar subquery; please use the .scalar_subquery() method to produce a scalar subquery.
```  

The issue occurs because a **SELECT object is being implicitly coerced** when a scalar subquery is expected.  

#### Fix Details  
- Updated the affected query in `alerts.py` to explicitly use `.scalar_subquery()`, ensuring compatibility with SQLAlchemy’s best practices.  
- This prevents unnecessary warnings and potential future incompatibility with newer SQLAlchemy versions.

### DIFF
--- a/keep/api/core/alerts.py
+++ b/keep/api/core/alerts.py
@@ -204,7 +204,8 @@ def get_threeshold_query(tenant_id: str):
         .where(LastAlert.tenant_id == tenant_id)
         .order_by(LastAlert.timestamp.desc())
         .limit(1)
-        .offset(alerts_hard_limit - 1),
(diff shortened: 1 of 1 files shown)


EnotShow:
This PR fix issue #4275

PR #4323, opened 2025-03-31
fix(ui): button color

talboren:
before:

https://github.com/user-attachments/assets/b1e18a67-3073-484b-bfbc-ad23fe50bf96

after:

https://github.com/user-attachments/assets/4a089284-b62b-41b9-9057-97241d655b3a


shahargl (approved):
lgtm

PR #4389, opened 2025-04-03
fix: add migration to change dashboard_config column type to JSON

### LINKED ISSUE #4387: [🐛 Bug]: Saving dashboard throws errors
**Describe the bug**
The reason is that dashboard_config column in dashboard table is VARCHAR(255) and the dashboard being saved always exceeds this limit causing dashboard saving not to work with errors.

**To Reproduce**

1. Go to Dashboard
2. Create a widget
3. Save the dashboard and see that saving errors

**Expected behavior**
Dashboard is saved with no errors.

skynetigor:
Closes #4387

## ✅ Checks


talboren:
@skynetigor maybe worth wrapping it up with try catch?

skynetigor:
> @skynetigor maybe worth wrapping it up with try catch?

I tested this migration for all supported DBs (upgrade/downgrade), no errors.
Also tested for a case when dashboard_config is already JSON column, also goes well, just skips altering.

talboren (approved):
LGTM

talboren (approved):
LGTM

Issue #4398, opened 2025-04-03
[🐛 Bug]: Installed providers disappear periodically


roma55443:
**Describe the bug**
Installed providers disappear periodically, causing errors when executing workflows.
The same behavior is observed with secrets.
As a result, workflows sometimes fail due to missing providers or secrets.

**To Reproduce**
	1.	Create an installed provider.
	2.	Refresh the page multiple times. Eventually, the installed providers list appears empty.

**Expected behavior**
Installed providers and secrets should not disappear when refreshing the page.

**Video**
Installed providers bug:

https://github.com/user-attachments/assets/13f6cf1f-195a-4c77-ab64-5b9a611f74a6

Keep Secrets bug:

https://github.com/user-attachments/assets/4bf0f31a-05cf-42ae-8c5d-af46a3935200

**Additional context**
Due to this issue, we cannot use Keep in a production environment, as workflows occasionally fail with provider or secret errors.

KeepHQ version: 0.41.5

talboren:
@roma55443 im sorry to hear that! That’s definitely nothing we observed before and does not reproduce on my end. May I know how are you deploying Keep? Can you share backend env variables maybe?

Knyage:
@talboren We have deployed KeepHQ to kubernetes via helm. 

env backaend: 
```yaml
env:
    - name: SECRET_MANAGER_TYPE
      value: k8s
    - name: PORT
      value: "8080"
    - name: PUSHER_APP_ID
      value: 1
    - name: PUSHER_APP_KEY
      value: keepappkey
    - name: PUSHER_APP_SECRET
      value: keepappsecret
    - name: PUSHER_HOST
      value: keep-websocket
    - name: PUSHER_PORT
      value: 6001
    - name: POSTHOG_DISABLED
      value: true
```

This problem is very critical and is blocking us :(

talboren:
> [@talboren](https://github.com/talboren) We have deployed KeepHQ to kubernetes via helm.
> 
> env backaend:
> 
> env:
>     - name: SECRET_MANAGER_TYPE
>       value: k8s
>     - name: PORT
>       value: "8080"
>     - name: PUSHER_APP_ID
>       value: 1
>     - name: PUSHER_APP_KEY
>       value: keepappkey
>     - name: PUSHER_APP_SECRET
>       value: keepappsecret
>     - name: PUSHER_HOST
>       value: keep-websocket
>     - name: PUSHER_PORT
>       value: 6001
[...]

PR #4399, opened 2025-04-03
feat: enhance provider and deduplication rule provisioning logic

### LINKED ISSUE #4401: [🐛 Bug]: Provisioned provider not updated with new configurations
Hi team, I found a bug when we provision providers via `KEEP_PROVIDERS` OR `KEEP_PROVIDER_DIRECTORY`. If the first time, I provision this provider:

```
KEEP_PROVIDERS={"keepVictoriaMetrics":{"type":"victoriametrics","authentication":{"VMAlertHost":"http://vm.host.com","VMAlertPort":5678}}
```

So if I now want to change the configure of the existing provisioned provider (not change its name but let's say change the host name to `http://localhost`)

```
KEEP_PROVIDERS={"keepVictoriaMetrics":{"type":"victoriametrics","authentication":{"VMAlertHost":"http://localhost","VMAlertPort":1234}}
```

But unfortunately, this new configurations will not be updated into existing provisioned provider. This is because of those logic:

```
if ProvidersService.is_provider_installed(tenant_id, provider_name):
    logger.info(f"Provider {provider_name} already installed")
    continue
```

Cannot update the existing provisioned provider. So I think there is better to drop this conditional statement check.

However, to make the provider configurations to be updated, so I think we should:
1. First, delete all existing provisioned providers.
2. Then, provision new providers from either the environment variable or the directory.
3. If any error occurs, rollback everything to the original state.

So each time the backend starts, restarts, ... The progress re-provisioning providers will happen. What if each time the backend restarts, the configurations not change, so I think we first calculate the hash of the configurations, then save it in either redis or local file system (in dir state).

tuantran0910:
Closes #4401

## 📑 Description

When Keep starts or restarts, it follows these steps to manage provider configurations:

1. **Read Configurations**: Loads provider definitions from either the `KEEP_PROVIDERS` environment variable or YAML files in the `KEEP_PROVIDERS_DIRECTORY`.
2. **Create New Providers**: Installs any providers listed in the configuration that are not already present.
3. **Update When Changed**: If an existing provider's configuration has changed, Keep reapplies the configuration, including deduplication rules. If errors occur during this update, changes are automatically rolled back.
4. **Delete Providers**: Deletes any currently installed providers that are not found in the loaded configuration.

## ✅ Checks


tuantran0910:
Hmm, storing the hash value when provisioning in local file storage might not be a good way when we have many backend replicas as mentioned in this issue #4398.

tuantran0910:
> hey @tuantran0910 - although we want to get this functionality, we feel it's a bit over-engineering here. let's sync if see if there is simpler solution for that.

Hi @shahargl, I completely agree with you. We always aim for the simplest solution to our problems. I don’t have much experience before, so I would really appreciate any advice you could share. Thanks! :D

tuantran0910:
Hi @talboren, @shahargl, I've simplified the provider management logic (as described in the issue/PR description). The process is now:
1. Iterate through the providers defined in the configuration.
2. If a provider doesn't exist locally, install it.
3. If a provider already exists, update it based on the configuration.
4. After processing all configured providers, delete any existing providers that were not found in the current configuration (stale providers).

How do you guys think about this strategy ?

talboren:
@tuantran0910 and I are discussing in Slack. Will re-open with an easier implementation for this.

PR #4420, opened 2025-04-05
fix: prevent fallback icon infinite loop

### LINKED ISSUE #4237: [🐛 Bug]:  non stop requests for get icon
I have few custom providers without icon
When I update keephq to 0.40.7 version I have bug with non stop requests for icon image.
When I want set icon image I haven't choice for zabbixcustom and dynatracecustom (look to screen)

Image

Image

35C4n0r:
Closes #4237 

## ✅ Checks

### DIFF
--- a/keep-ui/shared/api/ApiClient.ts
+++ b/keep-ui/shared/api/ApiClient.ts
@@ -48,6 +48,7 @@ export class ApiClient {
     }
     return {
       Authorization: `Bearer ${this.session.accessToken}`,
+      "ngrok-skip-browser-warning": true,
       ...this.additionalHeaders,
     };
   }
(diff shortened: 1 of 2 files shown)


keep-ui/shared/api/ApiClient.ts
```diff
@@ -48,6 +48,7 @@ export class ApiClient {
     }
     return {
       Authorization: `Bearer ${this.session.accessToken}`,
+      "ngrok-skip-browser-warning": true,
```
talboren, line 51: ?

35C4n0r: This is to prevent getting the warning page you get from the ngrok (if frontend is not on localhost). So instead of getting API response you get a HTML document as a response on the frontend, I always kept a patch for it on my local, it would be nice if we can get this merged.

talboren (approved):

PR #4509, opened 2025-04-11
feat: provision install webhook and http provider `fail_on_error`

### LINKED ISSUE #4506: [➕ Feature]: Provider provisionning webhook installation
As of now, we can provision providers by either environment variable or helm charts values but the webhook isn't getting created:

Image

Be able to install webhook in a programmatic mode too.
i.e. webhookinstall: true or false

Manually create the provider.

**Additional context**

### LINKED ISSUE #4508: How to ensure that an unsuccessful HTTP request does not lead to a workflow error?
### Discussed in https://github.com/keephq/keep/discussions/4507

<div type='discussions-op-text'>

<sup>Originally posted by **roma55443** April 10, 2025</sup>
Task:
I need to check if a channel already exists in the messenger before creating a new channel.

Problem:
When I send a request to check if the channel exists (using the HTTP Step), and the channel is not found, the messenger returns an error with code 400. As a result, this error breaks the entire workflow since status 400 is treated as a failure.

Goal:
I need to configure the HTTP request handling so that when the channel doesn’t exist (and a 400 status code is returned), it does not cause an error and does not interrupt the workflow. Instead, the result of this step should be used as a condition for creating a new channel.
</div>

talboren:
close #4508
close #4507 
close #4506

### DIFF
--- a/docs/snippets/providers/http-snippet-autogenerated.mdx
+++ b/docs/snippets/providers/http-snippet-autogenerated.mdx
@@ -20,6 +20,7 @@ steps:
         body: {value}
         params: {value}
         proxies: {value}
+        fail_on_error: {value}
 ```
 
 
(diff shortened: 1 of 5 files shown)


shahargl (approved):
lgtm

PR #4522, opened 2025-04-14
fix: Inverted condition in facet options gives weird results

### LINKED ISSUE #4521: [🐛 Bug]: Inverted condition in facet options gives weird results
**Describe the bug**
When facet options are limited to 50, the facet panel becomes unaware of values outside this subset. As a result, using a negation-based filter like:
`!(someField in [null, 'value1', 'value2', ...])`
causes incorrect results. This happens because the filter excludes the listed values but does not account for other possible values that weren’t returned due to the limit. Therefore, some rows that should be excluded remain included, leading to inconsistent or unexpected filtering.

**To Reproduce**

1. Go to feed
2. Select ONE option in facet having more than 50 rows
3. See that you have more results than the selected facet option counter

**Expected behavior**
Facets should apply filters using only the explicitly selected options with a positive match, such as:
`someField in ['value1', 'value2']`
Avoid using negation (!in) for filtering, as it depends on knowing the full set of values—which is not guaranteed when facets are limited. This ensures consistent and predictable filtering behavior based solely on user-selected options.

skynetigor:
Closes #4521 
## ✅ Checks

### DIFF
--- a/keep-ui/features/filter/facets-panel.tsx
+++ b/keep-ui/features/filter/facets-panel.tsx
@@ -31,9 +31,20 @@ function buildCel(
     .filter((facet) => facet.id in facetsState)
     .filter((facet) => facetOptions[facet.id])
     .map((facet) => {
-      const notSelectedOptions = Object.values(facetOptions[facet.id])
-        .filter((facetOption) =>
-          facetsState[facet.id]?.has(facetOption.display_name)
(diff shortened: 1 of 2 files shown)


shahargl (dismissed):
lgtm

talboren (dismissed):
lgtm

talboren (approved):

PR #4525, opened 2025-04-14
fix: validation for secrets and consts

### LINKED ISSUE #4513: [🐛 Bug]: workflow validation errors
**Describe the bug**
I encountered several problems during workflow validation:
	1.	An error occurs when I try to access “secrets”
	2.	Error in the http provider when using verify: false
— The verify: false parameter (which is sometimes needed when dealing with certificate issues) results in an error during validation.
	3.	consts.rocketChatNotCriticalAlertMessage is not highlighted as an error, even though such a constant does not exist (this is not critical)
	4.	Accessing an "action" via {{ steps }} is highlighted as an error
— When accessing an action result via {{steps.send_message_to_not_critical_chat.results }}, it is flagged as an error.
— However, using {{ actions.send_message_to_not_critical_chat.results }} does not raise an error, even though according to this [discussion](https://github.com/keephq/keep/discussions/4390), the correct way is to use steps, not actions.
	5.	Error: Variable: providers.default-mock - Provider "default-mock" not found.
— I use the mock provider for alert enrichment, as described in the [documentation](https://docs.keephq.dev/workflows/syntax/enrichment)
— But validation fails to recognize default-mock as a valid provider.

**Screenshots**
Image

Image

talboren:
close #4513

### DIFF
--- a/docs/snippets/providers/http-snippet-autogenerated.mdx
+++ b/docs/snippets/providers/http-snippet-autogenerated.mdx
@@ -21,6 +21,7 @@ steps:
         params: {value}
         proxies: {value}
         fail_on_error: {value}
+        verify: {value}
 ```
 
 
@@ -37,6 +38,7 @@ actions:
         body: {value}
         params: {value}
         proxies: {value}
+        verify: {value}
 ```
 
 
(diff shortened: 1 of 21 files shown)


shahargl (approved):
lgtm

Issue #2445, opened 2024-11-12
[🐛 Bug]: Topology applications is buggy when service does not have display name


talboren:
![image](https://github.com/user-attachments/assets/7d4a1bef-7827-4128-95cf-e96b51c78139)
![image](https://github.com/user-attachments/assets/645fd857-a517-4bde-bd18-490f0f480c28)

talboren:
cc: @Kiryous

PR #4562, opened 2025-04-16
fix: topology services without names

Matvey-Kuk:
close https://github.com/keephq/keep/issues/2445

Just a simple fallback


talboren (approved):
LGTM

PR #4566, opened 2025-04-16
fix: workflow execution trigger links

### LINKED ISSUE #4565: [🐛 Bug]: Workflow execution table links for alerts and incidents not working
When viewing workflow executions, clicking on alerts/incidents in the `Triggered by` column sometimes redirects to incorrect links with undefined IDs, causing users to view unrelated items or broken links.

The reason is that the id of the alert may contains special characters like colons and plus signs. For example: `dbt_docs_generate-generate_and_upload_metadata-scheduled__2025-04-16T14:30:00+00:00`. I know this is a weird id :D 

Image

We can see that instead of displaying `name: dbt-docs-..., id: dbt-docs-...`, it just display the a part in a name, after that is the splitted part after the plus sign `00:00`.

If we click into this alert in the `Triggered by` column, it will redirect to incorrect link.

Image

tuantran0910:
Closes #4565

## 📑 Description

When viewing workflow executions, clicking on alerts/incidents in the Triggered by column sometimes redirects to incorrect links with undefined IDs, causing users to view unrelated items or broken links.

The reason is that the id of the alert may contains special characters like colons and plus signs. For example: `dbt_docs_generate-generate_and_upload_metadata-scheduled__2025-04-16T14:30:00+00:00`.

**Before**:

image

image

**After**:

image

image


talboren (changes requested):
why not just use [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)?

tuantran0910:
> why not just use [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)?

Will check it later.

talboren:
Closing for now, as there's probably an easier win for this.

PR #4568, opened 2025-04-17
feat: batch alerts enrichment using cel

### LINKED ISSUE #4567: [➕ Feature]: Batch Enrichment by CEL Expression
This is ChatGPT generated description

## Description
Implement a new endpoint that allows users to enrich multiple alerts at once by specifying a CEL (Common Expression Language) query instead of listing individual alert fingerprints.

## Use Cases
- Enrich alerts by attributes like severity or name pattern
- Target alerts with specific labels or properties
- Bulk acknowledge alerts from specific regions or services
- Apply status changes to groups of related alerts

## Examples
```
// Acknowledge all critical severity alerts
{
  "cel": "labels.severity == 'critical'",
  "enrichments": {
    "status": "acknowledged"
  }
}

// Assign all memory-related alerts in a specific region
{
  "cel": "name.contains('Memory') && labels.region == 'us-east-1'",
  "enrichments": {
    "assignee": "infrastructure-team@example.com"
  }
}
```

tuantran0910:
Closes #4567

## 📑 Description

Implement a new endpoint that allows users to enrich multiple alerts at once by specifying a CEL (Common Expression Language) query instead of listing individual alert fingerprints.

## Usage

```bash
curl --location '127.0.0.1:8080/alerts/batch_enrich?dispose_on_new_alert=true' \
--header 'Content-Type: application/json' \
--header 'X-API-KEY: 5df8d505-dc43-49a6-ba3d-5ccaeb8c7e68' \
--data '{
    "enrichments": {
        "status": "resolved"
    },
    "cel": "name=='\''Pod api-service-production lacks memory'\''"
}'
```

## Additional

https://github.com/user-attachments/assets/a6bd1bd8-7a29-4dc1-83cd-21f055a1c626


talboren (commented):
cool stuff! just thinking out loud: I think it would be simpler to keep just 1 endpoint for batch enrichment and give the user the option to either supply fingerprints or supply a CEL query to find those fingerprints.

tuantran0910:
> cool stuff! just thinking out loud: I think it would be simpler to keep just 1 endpoint for batch enrichment and give the user the option to either supply fingerprints or supply a CEL query to find those fingerprints.

I am thinking about this, so now we only have one endpoint `/batch_enrich` and we have multiple options in the body request like this:

```python
class BatchEnrichAlertRequestBody(BaseModel):
    enrichments: dict[str, str]
    fingerprints: list[str]
    cel: str
    limit: int = 1000
    offset: int = 0
```

The endpoint accepts either `fingerprints` field or `cel` (in order to get the corresponding alerts's fingerprints). It will be the if else statement. How do you think about this @talboren ? This can reduce the duplication which you mentioned above.

talboren:
> > cool stuff! just thinking out loud: I think it would be simpler to keep just 1 endpoint for batch enrichment and give the user the option to either supply fingerprints or supply a CEL query to find those fingerprints.
> 
> I am thinking about this, so now we only have one endpoint `/batch_enrich` and we have multiple options in the body request like this:
> 
> ```python
> class BatchEnrichAlertRequestBody(BaseModel):
>     enrichments: dict[str, str]
>     fingerprints: list[str]
>     cel: str
>     limit: int = 1000
>     offset: int = 0
> ```
> 
> The endpoint accepts either `fingerprints` field or `cel` (in order to get the corresponding alerts's fingerprints). It will be the if else statement. How do you think about this @talboren ? This can reduce the duplication which you mentioned above.

Yep! makes sense. If you provide both cel and fingerprints, just throw an exception.

1. if you provider fingerprints: it acts the same as today.
2. if you provide cel: you first get the fingerprints and the act the same as today.

if that make sense :)

tuantran0910:
> > > cool stuff! just thinking out loud: I think it would be simpler to keep just 1 endpoint for batch enrichment and give the user the option to either supply fingerprints or supply a CEL query to find those fingerprints.
> > 
> > 
> > I am thinking about this, so now we only have one endpoint `/batch_enrich` and we have multiple options in the body request like this:
> > ```python
> > class BatchEnrichAlertRequestBody(BaseModel):
> >     enrichments: dict[str, str]
> >     fingerprints: list[str]
> >     cel: str
> >     limit: int = 1000
> >     offset: int = 0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The endpoint accepts either `fingerprints` field or `cel` (in order to get the corresponding alerts's fingerprints). It will be the if else statement. How do you think about this @talboren ? This can reduce the duplication which you mentioned above.
> 
> Yep! makes sense. If you provide both cel and fingerprints, just throw an exception.
> 
> 1. if you provider fingerprints: it acts the same as today.
> 2. if you provide cel: you first get the fingerprints and the act the same as today.
> 
> if that make sense :)

Cool, will work on that approach.

tuantran0910:
Hi @talboren, I have finished modifying the logic, but there is one test failed in **Integration Test**. I don't really understand why the test was failed, can you help me with this ? Thanks :D

shahargl:
@tuantran0910 something doesn't adds up with the openapi.json (new 10k lines of code)

tuantran0910:
Hi @talboren, @shahargl, I think everything is fine now. The PR is ready to review and hopefully merge.

skynetigor:
Hey guys!
Since we're gonna have CEL for this, what do you think about deprecating `fingerprints` property in request payload?
With CEL we can do this: `fingerprint in ['one', 'two', 'three']`, so we won't need to maintain additional not necessary thing.
CC: @shahargl @talboren

skynetigor:
Nice job @tuantran0910 🔥

keep/api/routes/alerts.py
```diff
@@ -805,6 +804,78 @@ def batch_enrich_alerts(
...
+                "tenant_id": tenant_id,
+            },
+        )
+
+        try:
+            db_alerts, total_count = query_last_alerts(
+                tenant_id=tenant_id,
+                query=QueryDto(
+                    cel=enrich_data.cel,
+                    limit=enrich_data.limit,
```
skynetigor, line 832: Just thinking out loud:
Shouldn't the backend handle this internally?
It seems more appropriate for the backend to enrich all alerts matching the provided CEL expression, rather than offloading that responsibility to the client.
Also, having limit and offset parameters for enrichment feels a bit confusing to me.
What do you think?
Please let me know if I missed something.

tuantran0910: Hi @skynetigor, can you provide more explanation about the statement `Shouldn't the backend handle this internally? It seems more appropriate for the backend to enrich all alerts matching the provided CEL expression, rather than offloading that responsibility to the client.` since I don't get what is the client and backend in our context. Sorry, just a beginner to the software development journey :D 

I add the option to provide limit and offset because the `QueryDTO` had it, also if you don't provide, the default value of limit will be `1000` and offset is `0`. What about either we add option that limit can be `-1` which is no limitation or change the default to no limitation instead of `1000`.

tuantran0910: Thanks @skynetigor, I understand it.

Just one more question: Is the reason for setting the default limit to 1000 to avoid returning a large amount of data?

If so, should we consider implementing a technique to fetch all data in batches sequentially and then process enriching each batch, instead of imposing a hard limit of only `1000` ? For example if total count is `10000`, then we may have to process each batch `1000` one by one.

skynetigor: Yes, I think it's a good way to go

tuantran0910: Thanks @skynetigor, for simplicity, I think we can go ahead and get this PR reviewed and merged first. We can introduce the improvements in future PRs. 😄

keep/api/models/alert.py
```diff
@@ -367,7 +367,10 @@ class EnrichAlertRequestBody(BaseModel):
 
 class BatchEnrichAlertRequestBody(BaseModel):
     enrichments: dict[str, str]
-    fingerprints: list[str]
+    fingerprints: Optional[list[str]] = None
+    cel: Optional[str] = None
+    limit: int = 1000
```
skynetigor, line 372: Let's maybe remove these limit/offset parameters from batch enrichment request and hardcode limit:1000 offset:0
in the place where it's used now. I think no need to make limit and offset being part of public api especially because it's gonna be change in future.

keep/api/routes/alerts.py
skynetigor: I don't mind merging it.
Left a final comment.

keep/api/models/alert.py
```diff
@@ -367,7 +367,10 @@ class EnrichAlertRequestBody(BaseModel):
...
+    offset: int = 0
```
skynetigor, line 373:
```suggestion
```

keep/api/routes/alerts.py
```diff
@@ -805,6 +804,78 @@ def batch_enrich_alerts(
...
+                    offset=enrich_data.offset,
```
skynetigor, line 833:
```suggestion
                    offset=0,
```

tuantran0910: Delete at https://github.com/keephq/keep/pull/4568/commits/524d00c72d1ec0b5f985577dbb7f21a67a0dc34c.

keep/api/routes/alerts.py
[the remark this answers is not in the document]
tuantran0910: Deleted at https://github.com/keephq/keep/pull/4568/commits/524d00c72d1ec0b5f985577dbb7f21a67a0dc34c.

keep/api/models/alert.py
tuantran0910: Deleted at https://github.com/keephq/keep/pull/4568/commits/524d00c72d1ec0b5f985577dbb7f21a67a0dc34c.

skynetigor (approved):
LGTM

tuantran0910:
Hi @talboren, can I get this merged :D Thanks

skynetigor:
@tuantran0910 Merging the PR. Thank you for the contribution 🙏

PR #4577, opened 2025-04-17
feat: workflow trigger with cel expression

talboren:
### _**THIS IS DEPRECATING OLD FILTERS MECHANISM IN FAVOR OF NEW CEL MECHAINSM AND MIGHT BREAK THINGS. WE DO BEST EFFORTS TO HAVE BACKWARD COMPATIBILITY BUT SOME OLD FILTERS MAY BE FLAKY**_

Still needs to be done:

- [x] Update documentation to elaborate on filters with CEL and mention that old filters is deprecated (and backwards compatibility is best-effort) 
- [ ] Make the UI show a warning when using deprecated filters
- [ ] Announce a potentially breaking change


talboren:
Still needs to be done:

- [ ] Update documentation to elaborate on filters with CEL and mention that old filters is deprecated (and backwards compatibility is best-effort) 
- [ ] Make the UI show a warning when using deprecated filters

keep-ui/entities/workflows/model/schema.ts
```diff
@@ -33,6 +33,7 @@ export const V2StepAlertTriggerSchema = z.object({
   properties: z.object({
     alert: AlertTriggerValueSchema,
     source: z.string().optional(),
```
shahargl, line 35: Shouldn't it be removed now? only cel? or do we support backward compatibility?

talboren: we have backward compatibility. It's only for the purpose of validation, so having `source` will work as expected

shahargl:
things to check (?)

1. Trigger workflow with alert - i think it tries to extract fields from filters

shahargl (approved):
lgtm

skynetigor (approved):
LGTM

keep-ui/entities/workflows/lib/__tests__/validation.test.ts
```diff
@@ -496,27 +496,6 @@ describe("validateGlobalPure", () => {
     ]);
   });
 
-  it("should detect empty alert trigger", () => {
```
Kiryous, line 499: @talboren why did you remove this check? now it should check either cel or filters, right?

talboren: Because empty alerts filter was OK even before. It meant to trigger every alert.

keep-ui/features/workflows/builder/lib/utils.tsx
```diff
@@ -26,7 +26,7 @@ const alertTriggerTemplate: V2StepTrigger = {
   id: "alert",
   properties: {
     alert: {
```
Kiryous, line 28: i think we should change the internal representation of an alert trigger , something like 

```ts
properties: {
  alert: {
    cel: string;
    filters: Record<string, string>; // deprecated filters
  }
}
```

and the logic in `parser.ts` should be updated to parse deprecated filters into filters key https://github.com/keephq/keep/blob/e78953fe751fa678e80497074ecce5d201566315/keep-ui/entities/workflows/lib/parser.ts#L221-L229

keep-ui/features/workflows/builder/ui/Editor/TriggerEditor.tsx
```diff
@@ -100,30 +107,55 @@ export function TriggerEditor() {
...
-                      }
-                    />
-                    <Icon
-                      icon={BackspaceIcon}
-                      className="cursor-pointer"
-                      color="red"
-                      tooltip={`Remove ${filter} filter`}
-                      onClick={() => deleteFilter(filter)}
-                    />
+              Object.keys(properties.alert ?? {}).map((filter) =>
```
Kiryous, line 110: I think we should place CEL input above deprecated filters and add some indication of deprecation, something like this

Screenshot 2025-04-29 at 18 23 08

keep-ui/entities/workflows/model/schema.ts
```diff
@@ -33,6 +33,7 @@ export const V2StepAlertTriggerSchema = z.object({
...
+    cel: z.string().optional(),
```
Kiryous, line 36: UI for trigger do not show correct values for valid CEL



https://github.com/user-attachments/assets/e9529b8f-3ecd-4b44-92be-5207cd76bbed

Screenshot 2025-04-29 at 17 57 17

Kiryous (commented):
Awesome work! Some critical things regarding UI should be addressed; otherwise, it looks good!

Kiryous (changes requested):
See prev. comment

keep-ui/features/workflows/builder/lib/utils.tsx
talboren: @Kiryous i'm not sure I follow on this comment, can you maybe elaborate?

> and the logic in parser.ts should be updated to parse deprecated filters into filters key

what does that mean?

keep-ui/features/workflows/builder/ui/Editor/TriggerEditor.tsx
```diff
@@ -33,9 +44,20 @@ export function TriggerEditor() {
...
     const currentFilters = properties.alert || {};
-    const updatedFilters = { ...currentFilters, [filter]: value };
-    updateV2Properties({ alert: updatedFilters });
-    saveNodeDataDebounced("properties", { alert: updatedFilters });
+    if (!currentFilters.filters) {
+      currentFilters.filters = {};
+    }
+    currentFilters.filters[filter] = value;
+    updateV2Properties({ alert: currentFilters });
+    saveNodeDataDebounced("properties", { alert: currentFilters });
```
Kiryous, line 52: it's confusing, but to correctly display the trigger node `saveNodeDataDebounced` should receive the properties not wrapped in {alert: ...}, because it's different from workflow properties.

Also, it's better to create new object to avoid possible state inconsistency.
```suggestion
    saveNodeDataDebounced("properties", {...currentFilters, [filter]: value});
```

keep-ui/features/workflows/builder/ui/Editor/TriggerEditor.tsx
```diff
@@ -33,9 +44,20 @@ export function TriggerEditor() {
...
+  };
+
+  const updateAlertCel = (value: string) => {
+    const currentFilters = properties.alert || {};
+    updateV2Properties({ alert: { ...currentFilters, cel: value } });
+    saveNodeDataDebounced("properties", {
+      alert: { ...currentFilters, cel: value },
```
Kiryous, line 59: it's confusing, but to correctly display the trigger node saveNodeDataDebounced should receive the properties not wrapped in {alert: ...}, because it's different from workflow properties. [2]

```suggestion
      { ...currentFilters, cel: value }
```

shahargl (approved):
Lgtm

PR #4595, opened 2025-04-19
Feature/jinja template engine in workflow templates

### LINKED ISSUE #4594: [➕ Feature]: Jinja template engine in workflow templates
**Is your feature request related to a problem? Please describe**
In our system, some alerts include all the labels, while others do not. When certain metrics or fields are missing, it causes exceptions during template rendering. This makes it difficult to build robust templates that can handle all alert variations gracefully. It's currently difficult due to the limitations of custom templating. Using Jinja instead would make it much easier and more flexible.

Replace the current custom templating system with Jinja. This would provide greater flexibility and power when creating templates, especially when handling optional fields and conditional logic.

- Extend the current custom templating to support conditional (if) blocks
- Add support for default values when a metric or label is missing in the workflow

EnotShow:
Closes #4594 

## 📑 Description
This PR added the whole power of Jinja2 templating to the KeepHQ workflow

## ✅ Checks

- [ x] My pull request adheres to the code style of this project
- [ x] I have updated the documentation as required

### DIFF
--- a/docs/workflows/syntax/templating.mdx
+++ b/docs/workflows/syntax/templating.mdx
@@ -1,0 +1,132 @@
+---
+title: Templating
+---
+KeepHQ allows templating within YAML workflow definitions using [Jinja2](https://jinja.palletsprojects.com/) – a flexible and powerful templating language for Python. This is especially useful for dynamically generating descriptions, titles, and other content in alerts or actions like Jira ticket creation.
+---
+
+### Basic Syntax
+
+#### Variables
+Use `{{ variable_name }}` to interpolate variables.
+```jinja2
+{{ alert.alertname }}
+{{ alert.annotations.summary }}
+{{ alert.generatorURL }}
+```
+
+#### Logic Conditions
+Use `{% if %}`...`{% endif %}` blocks to conditionally include fields.
+
+```jinja2
+{% if alert.channel %}
+*Channel:* {{ alert.channel }}
+{% endif %}
+```
+
+You can chain multiple independent conditions:
+```jinja2
+{% if alert.cluster %}
+*Cluster:* {{ alert.cluster }}
+{% endif %}
+{% if alert.namespace %}
+*Namespace:* {{ alert.namespace }}
+{% endif %}
+```
+
+#### Default Values
+Use `or` to set a default value when a variable may be empty or undefined:
+```jinja2
+*Alertgroup:* {{ alert.alertgroup or 'None' }}
+```
+
+---
+
+### 📦 Example: JiraOnPrem Action Description Block
+
+Here's a real-world usage pattern in a `jiraonprem` provider action:
+
+```jinja2
+workflows:
+  - id: jiraonprem-investigate
+    name: Create investigation
+    description: Creates investigation task in jira
+    disabled: false
+    triggers:
+      - type: manual
+    consts: {}
+    owners: []
+    services: []
+    steps: []
+    actions:
+      - name: jiraonprem-action
+        provider:
+          type: jiraonprem
+          config: "{{` {{ providers.jiraOnPrem }} `}}"
+          with:
+            board_name: SA
+            components:
+              - Monitoring
+            custom_fields: ""
+            description: "--------------------------------------------------------------------
+
+            *Alertname:* {{ alert.alertname }}
+
+            *Summary:* {{ alert.annotations.summary }}
+
+            *Description:* {{ alert.description }}
+
+            *Alertgroup:* {{ alert.alertgroup or 'None' }}
+
+            {% if alert.channel %}
+            *Channel:* {{ alert.channel }}
+            {% endif %}
+
+            {% if alert.cluster %}
+            *Cluster:* {{ alert.cluster }}
+            {% endif %}
+
+            {% if alert.namespace %}
+            *Namespace:* {{ alert.namespace }}
+            {% endif %}
+
+            {% if alert.project %}
+            *Project:* {{ alert.project }}
+            {% endif %}
+
+            *Severity:* {{ alert.severity }}
+
+            --------------------------------------------------------------------
+
+            {% if alert.annotations.runbook_url %}
+            Runbook: [Runbook|{{ alert.annotations.runbook_url }}]
+            {% endif %}
+
+            {% if alert.generatorURL %}
+            Query: [Query|{{ alert.generatorURL }}]
+            {% endif %}
+
+            --------------------------------------------------------------------"
+```
+
+---
+
+### ✅ Best Practices
+
+- **Escape YAML properly**: If using multiline Jinja in YAML, wrap with double quotes (`"`) and indent consistently.
+- **Avoid logic nesting when possible**: Keep your `if` statements clean and flat for readability.
+- **Always check for nulls**: Use conditionals for optional values to avoid rendering issues.
+- **Use `or` for fallback values**: This keeps your output graceful even with missing data.
+
+---
+
+### 📘 Useful Filters
+
+- `default` – Provide fallback if a variable is undefined:
+  ```jinja2
+  {{ alert.cluster | default('Unknown Cluster') }}
+  ```
+- `join` – Join lists into strings:
+  ```jinja2
+  {{ alert.labels | join(', ') }}
+  ```
+---
--- a/keep/iohandler/iohandler.py
+++ b/keep/iohandler/iohandler.py
@@ -1,10 +1,8 @@
 import ast
 import copy
 import html
-
 # TODO: fix this! It screws up the eval statement if these are not imported
 import inspect
-import io
 import json
 import logging
 import re
@@ -12,12 +10,18 @@ import sys
 
 import astunparse
 import requests
-from jinja2 import Template
+from jinja2 import Undefined, Environment
 
 import keep.functions as keep_functions
 from keep.contextmanager.contextmanager import ContextManager
 from keep.step.step_provider_parameter import StepProviderParameter
 
+missing_keys = set()
+
+class TrackingUndefined(Undefined):
+    def __str__(self):
+        missing_keys.add(self._undefined_name)
+        return super().__str__()
 
 class RenderException(Exception):
     def __init__(self, message, missing_keys=None):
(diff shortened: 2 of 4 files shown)


shahargl:
@EnotShow lot of unit tests are failing, can you fix that? :)

shahargl (changes requested):
tests failing

talboren:
Also a quick question about backwards compatibility - not sure if I understand if this will still support mustache or we're replacing to Jinja completely? Maybe we should support `templating=mustahce` or `templating=jinja` in the top of the workflow to decide which engine we want to use. Will the `keep.` functions work?

EnotShow:
@talboren

1. This PR fully replaces Chevron with Jinja2 in workflows. I’ve proposed sticking with it, since once this is done, the previous workflows need to remain fully compatible. But it makes sense to proceed cautiously, especially since we don’t yet know all the potential issues this might introduce. So if the team decides it has to be done, I’m happy to go ahead with it.

2. Yes, keep functions working perfectly!

EnotShow:
Ahh, now that I've gone through the Chevron docs, I see why Mustache needs to stay

EnotShow:
I need some help with workflow execution tests for jinja2 cose, I not so sure about how it work, but except it work seems like done

EnotShow:
Done!

talboren:
@EnotShow i'll review the PR asap. a bit busy the last couple of days but excited about it 🤩

keep/iohandler/iohandler.py
```diff
@@ -24,10 +23,18 @@ def __init__(self, message, missing_keys=None):
         self.missing_keys = missing_keys
         super().__init__(message)
 
+class TemplateEngine(enum.StrEnum):
+    JINJA2 = "jinja2"
+    MUSTACHE = "mustache"
 
 class IOHandler:
```
skynetigor, line 30: Hey! 👋
Thanks for the contribution!

After reviewing the PR, I have a concern regarding the changes to IOHandler.
From what I understand, up until now IOHandler was responsible only for Mustache templating. With this PR, it starts supporting both Mustache and Jinja2, but the integration adds quite a bit of complexity — especially with multiple if template == jinja conditions scattered across different functions.

I'm a bit worried that this approach will make IOHandler messy and fragile over time, making it easier to accidentally break one or the other templating engine when modifying the code.

**Suggestion:**
Instead of mixing both engines into the same handler, it might be better to create a separate Jinja2IOHandler that implements the same interface but is dedicated to Jinja logic only (without any Mustache-specific code). Then, depending on the selected template engine in the workflow, we could simply switch between MustacheIOHandler and Jinja2IOHandler (they must share same interface).

This separation would keep the code cleaner, easier to maintain, and less error-prone.

Of course, these are my first impressions based on the review — if I misunderstood something about the templating architecture, feel free to correct me!

Thanks again for the work on this! 🚀

EnotShow: The previous implementation was similar to this one, because I believe that someday we could mark Mustache as deprecated and keep only Jinja2, since Jinja2 is a more suitable solution in my opinion.

I have rewritten my solution as you suggested. Now we have a base class and two separate classes: MustacheIOHandler and Jinja2Handler. I also added a mixin that checks whether the user is using the correct syntax for the selected template engine and raises a RenderError if not.

EnotShow:
I’m encountering a strange error in my tests that I can’t reproduce locally. The error message is:

FAILED tests/e2e_tests/incidents_alerts_tests/test_filtering_sort_search_on_alerts.py::test_alerts_stream - AssertionError: Locator expected to have count '20'
Actual value: 30 
Call log:
  - LocatorAssertions.to_have_count with timeout 5000ms
  -   - waiting for locator("[data-testid='alerts-table'] table tbody tr")
  -     9 × locator resolved to 30 elements
  -       - unexpected value "30"
============ 1 failed, 32 passed, 15 warnings in 152.97s (0:02:32) =============

The test expects 20 alerts, but it’s finding 30. Interestingly, when I run the test locally, it passes with exactly 20 alerts.

Could the issue be that the database is not being recreated properly for every test run?

skynetigor:
Hi @EnotShow!
Now it looks good.
Just a quick question: Mustache works as well as before and is unchanged in terms of functionality it provides, right?

EnotShow:
Hey @skynetigor,
The main goal was to preserve Mustache’s functionality and make Jinja2 behave the same way as Mustache. So yes, Mustache continues to work just as it did before.

skynetigor (dismissed):
LGTM

docs/workflows/syntax/templating.mdx
```diff
@@ -0,0 +1,171 @@
+---
+title: Templating
```
skynetigor, line 2: Something wrong in that file with syntax that fails doc tests

EnotShow: Fixed

skynetigor (dismissed):
LGTM

EnotShow:
Hi @talboren, @shahargl!
All tests are passed, and I received approval from @skynetigor, so it seems ready for merge!

keep/iohandler/iohandler.py
```diff
@@ -24,8 +23,84 @@ def __init__(self, message, missing_keys=None):
...
+            'inverted': r'{{(\^|\^!)[^}]+}}',  # {{^section}} or {{^!section}}
+            'end_section': r'{{/[^}]+}}',  # {{/section}}
+            'unescaped': r'{{{[^}]+}}}',  # {{{ variable }}}
+            'comment': r'{{![^}]+}}',  # {{! comment }}
+            'partial': r'{{>[^}]+}}',  # {{> partial }}
+        }
+
+        has_jinja2 = False
+        has_mustache = False
+
```
shahargl, line 64: I think there is a better and simpler way to check jinja2 expressions, 

e.g.:

```python

from jinja2 import Environment, TemplateSyntaxError
 env = Environment()

# valid
In [4]: env.parse("Hello, {{ name }}!")
Out[4]: Template(body=[Output(nodes=[TemplateData(data='Hello, '), Name(name='name', ctx='load'), TemplateData(data='!')])])

# invalid
In [8]: a = """{% for item in items %}
   ...:   {{ item }}
   ...: {% endif %}"""

In [9]: env.parse(a)
-------------------------------------------------------------------
TemplateSyntaxError               Traceback (most recent call last)
Cell In[9], line 1
----> 1 env.parse(a)

File ~/git/keep/.venv/lib/python3.11/site-packages/jinja2/environment.py:616, in Environment.parse(self, source, name, filename)
    614     return self._parse(source, name, filename)
    615 except TemplateSyntaxError:
--> 616     self.handle_exception(source=source)

File ~/git/keep/.venv/lib/python3.11/site-packages/jinja2/environment.py:942, in Environment.handle_exception(self, source)
    937 """Exception handling helper.  This is used internally to either raise
    938 rewritten exceptions or return a rendered traceback for the template.
    939 """
    940 from .debug import rewrite_traceback_stack
--> 942 raise rewrite_traceback_stack(source=source)

File <unknown>:3, in template()

TemplateSyntaxError: Encountered unknown tag 'endif'. Jinja was looking for the following tags: 'endfor' or 'else'. The innermost block that needs to be closed is 'for'.
```

shahargl (changes requested):
seems almost done, see comments.

and if you can add more tests it will be great <3 just a big change so want to make sure we are ok.

keep/iohandler/iohandler.py
EnotShow: It’s a good approach, but my goal here isn’t to check whether the Jinja2 template is syntactically valid. Instead, I want to detect if the user accidentally used Mustache syntax when they selected Jinja2 as the template engine (or vice versa), and raise an informative error.

So it's more about catching cross-engine mistakes — for example, using {{#section}}...{{/section}} inside a Jinja2 context, or {% if ... %} inside a Mustache one.

That said, I agree this logic could probably be simplified. I’ll take another look and try to clean it up.

keep/workflowmanager/workflow.py
```diff
@@ -25,6 +25,7 @@ def __init__(
         workflow_id: str,
         workflow_revision: int,
         workflow_name: str,
+        workflow_templating: TemplateEngine,
```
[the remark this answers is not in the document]
EnotShow: It's not necessary, but sure, why not

keep/iohandler/iohandler.py
shahargl: 👑

shahargl:
@EnotShow let me know when you want the final review + merging

EnotShow:
@talboren This is ready for review, but I ran into an issue with Jinja2.

Jinja2 doesn't allow hyphens in identifier names when accessed as attributes in templates. To work around this, I implemented a function which replaces code like:

```
Pod status report:
{% for pod in steps.get-pods.results %}
Pod name: {{ pod.metadata.name }} || Namespace: {{ pod.metadata.namespace }} || Status: {{ pod.status.phase }}
{% endfor %}
```

with:

```
Pod status report:
{% for pod in steps['get-pods'].results %}
Pod name: {{ pod.metadata.name }} || Namespace: {{ pod.metadata.namespace }} || Status: {{ pod.status.phase }}
{% endfor %}
```
This change uses dictionary-style access, which is valid in Jinja2 for keys that contain hyphens. It seems to work, but I not sure if this solution is proper enough. What you think?

talboren:
> @talboren This is ready for review, but I ran into an issue with Jinja2.
> 
> Jinja2 doesn't allow hyphens in identifier names when accessed as attributes in templates. To work around this, I implemented a function which replaces code like:
> 
> ```
> Pod status report:
> {% for pod in steps.get-pods.results %}
> Pod name: {{ pod.metadata.name }} || Namespace: {{ pod.metadata.namespace }} || Status: {{ pod.status.phase }}
> {% endfor %}
> ```
> 
> with:
> 
> ```
> Pod status report:
> {% for pod in steps['get-pods'].results %}
> Pod name: {{ pod.metadata.name }} || Namespace: {{ pod.metadata.namespace }} || Status: {{ pod.status.phase }}
> {% endfor %}
> ```
> 
> This change uses dictionary-style access, which is valid in Jinja2 for keys that contain hyphens. It seems to work, but I not sure if this solution is proper enough. What you think?

makes absolute sense! I'll try to review it ASAP.

korolenkowork:
@shahargl @talboren any updates on this?

shahargl:
@korolenkowork see comment on the other PR. This is crazy great but I need to see how I prioritize reviewing it.

korolenkowork:
Hey, @shahargl does it means that changes permanently declined?

PR #4611, opened 2025-04-23
fix: Improve "in" CEL operator behavior with null

### LINKED ISSUE #4573: [🐛 Bug]: Can't query incident field of alerts
@skynetigor Using MySQL and Version: 0.41.26 | Build: 460b78

I am trying to look up for alerts, that went orphan / didn't get correlated into incidents. I'm aware of the field "incident", that I can see in the alert payload. But it's quite confusing :

For instance, if I make a filter like : incident.name in [null] OR If I click on incident (in my facet) "None", and filter :
It gives me a list of alerts. and if I open the first one's payload, I see :

"incident": "1adb2362-a3b5-41d5-be98-5ef7957f233c".

I can then go the /incidents/that and it's valid.

If I make a new facet with incident.id, and I click filter Only on "None", it find nothing, despite the big number besides :

Clicking on incident.id None, makes this query in the background : (incident.id in [null])

Image

Image

skynetigor:
Closes #4573

## ✅ Checks

### DIFF
--- a/keep/api/core/cel_to_sql/sql_providers/base.py
+++ b/keep/api/core/cel_to_sql/sql_providers/base.py
@@ -311,6 +311,9 @@ class BaseCelToSqlProvider:
         return result
 
     def _visit_equal(self, first_operand: str, second_operand: str) -> str:
+        if second_operand is None:
+            return f"{first_operand} IS NULL"
+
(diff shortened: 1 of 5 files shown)


VladimirFilonov (approved):
LGTM

PR #4645, opened 2025-04-29
fix: Skip instance-level enrichment for elastic. Ensure enrichments are applied conditionally

### LINKED ISSUE #4342: [🐛 Bug]: Problem with enriching alerts in Elasticsearch when we use enrichment in the workflow (webhook)
As I see, there is an issue with alert enrichment at the workflow level.
If we look into Elasticsearch, the _id of the document is as follows: “42dec3af-c35e-4387-805d-4b0fe8348db9”. However, during the document update in Elasticsearch, the workflow tries to search for the document with _id:f88f364d-a13f-44f2-bf82-f55ca19527e6 and receives a 404 error.
Looks like at first you use the ALERT ID, but in the next step, you try to search for the document by the EVENT ID.

VladimirFilonov:
Added checks to confirm the presence of `_enrichments` and `disposable_enrichments` before applying them in the alert enrichment process. Also updated the enrichment logic to prevent saving instance-level enrichments for Elastic environments.

Closes #4342 

## ✅ Checks

### DIFF
--- a/keep/api/bl/enrichments_bl.py
+++ b/keep/api/bl/enrichments_bl.py
@@ -716,6 +716,8 @@ class EnrichmentsBl:
         alert_id = UUIDType(binary=False).process_bind_param(
             last_alert.alert_id, self.db_session.bind.dialect
         )
+        # For elastic we do not save instance-level enrichments
+        common_kwargs["should_exist"] = False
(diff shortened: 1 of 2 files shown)


talboren (approved):
LGTM

PR #4646, opened 2025-04-29
feat: Add support for displaying incident details in alerts

### LINKED ISSUE #4610: [➕ Feature]: Easy way to open list of linked alerts to an incident, from the feed
I'd like either to be able to open each alert, like in the feed (with HTML intepreting) : 

Image

Or, to find a direct link, from the incident, to /alerts/feed, that would contains them.

For instance, if my incident is /incident/abcd1234, to find a link that would bring me to the alerts/feed filter incident.id="abcd1234"

And it seems to work rather "easily", just tried, the only trick is that the incident id in its url has '-', in it, but not in alert.incident.id

Image

VladimirFilonov:
Introduced `incident_dto` to enhance alerts with detailed incident data. Updated UI components to render incident links and names, leveraging `IncidentDto` structures. Adjusted backend logic to include and process `incident_dto` for enriched incident representation.

Also add "View in Feed" link to incident alerts tab

Closes #4610 

## 📑 Description
image

image

image

## ✅ Checks


skynetigor (approved):
LGTM

PR #4658, opened 2025-04-30
feat: add providerName field to alert payload

### LINKED ISSUE #3546: [➕ Feature]: Add "provider name" to alert payload
Current behaviour: 
1. Currently alert have providerId and providerType but not a providerName (which is the name of the provider)

Expected behaviour:
1. alert will also have the providerName

tuantran0910:
Closes #3546

## 📑 Description

Add providerName field to alert payload.

## ✅ Checks

## ℹ Additional Information

https://github.com/user-attachments/assets/6be52418-352a-4585-9960-5ab826638f18


talboren:
@tuantran0910 thank you for that but i'm not sure what's the real use case for adding that. provider_instance in most cases will be None, since we don't pass it.
I'll go ahead and close the issue and PR for now and let's re-open it if there's a real use case.

For now, in the UI side, we do the "translation" via getting the providers in the UI:

CleanShot 2025-05-01 at 21 04 24@2x

tuantran0910:
For me, this will be useful in case we provision two instances of the same provider type. In the feed, only show the icon won't help us distinguish two instances. Therefore, this field can address this issue.

PR #4664, opened 2025-04-30
fix: delete workflows that are not provisioned anymore

### LINKED ISSUE #4603: [🐛 Bug]: Workflow not deleted when we don't provision it anymore
Hi team,

I am provisioning a workflow via the environment, but when I stop provisioning it, the workflow continues to run normally instead of being deleted.

Additionally, I’ve noticed that if I repeatedly stop provisioning and restart the backend, the number of workflows increases significantly.

Kiryous:
Closes #4603


keep/workflowmanager/workflowstore.py
```diff
@@ -330,43 +340,53 @@ def provision_workflows(
...
-                    if not workflow.id == workflow_id:
-                        logger.info(
-                            f"Deprovisioning workflow {workflow.id} as its id doesn't match the provisioned workflow provided in the env"
-                        )
-                        delete_workflow_by_provisioned_file(
-                            tenant_id, workflow.provisioned_file
-                        )
+                    if (
+                        not pre_parsed_workflow
+                        or not workflow.id == pre_parsed_workflow.id
```
tuantran0910, line 362: I believe comparing `id` here is incorrect, as `pre_parsed_workflow.id` (which comes from the provisioning workflow) returns a random UUID. It would be better to compare `name` instead.

Kiryous: ah, nice catch

Kiryous:
@greptileai

talboren (approved):

PR #4670, opened 2025-05-01
fix: labels namespace filtering in rules not working

### LINKED ISSUE #4662: [🐛 Bug]: labels.namespace filtering in rules
**Describe the bug**
When saving correlation rules that include filters based on labels.namespace the rule seems to fail when saving, and displays no active filters when viewing the subsequent correlation rule.

**To Reproduce**

1. Go to correlation page
2. Create new correlation rule
3. Set filter to match labels.namespace = something
4. Save the correlation rule
5. Correlation rule should now show no filters, editing it will show a filter without conditions

**Expected behavior**
Rule should save successfully and display the filters in place.

**Screenshots**
Image
Image

**Additional context**
Speaking with the team on Slack it seems like this is an edge case introduced by the way namespace is used in the backend.

skynetigor:
Closes #4662

## 📑 Description

- **Improved CEL AST Serialization:** Made the CEL AST JSON-friendly by integrating Pydantic models.  
- **Simplified AST Structure:** Removed the `MethodAccess` node to streamline the CEL AST structure.  
- **Normalized String Matching Nodes:** Replaced `contains`, `startswith`, and `endswith` method nodes with `ComparisonNode` using appropriate operators.  
- **Read-Only Rule Display:** Updated `CelInput` component to support read-only mode for displaying rules in the Correlation Rules table.  
- **Reliable AST Parsing:** Replaced the buggy `react-querybuilder/parseCEL` with a new approach that uses CEL AST JSON to construct the `QueryBuilder` structure.  
- **Alert Query Refactor:** Replaced the `/alerts/search` endpoint with `/alerts/query` for Correlation Rules.  

image

## ✅ Checks


shahargl (approved):
lgtm

PR #4677, opened 2025-05-01
fix: 4238 bug zabbix provider is not compatible with zabbix 60 and zabbix 62

### LINKED ISSUE #4674: [🐛 Bug]: Kubernetes provider get_pods error
While trying to use "get_pods" inside a workflow am getting error: "Failed to run step get-pods with error 'latin-1' codec can't encode character '\u2502' in position 202: ordinal not in range(256)"

workflow is finding the pod, namespace and container from the alert

Backend and frontend env is set by default:
LC_ALL=en_US.UTF-8
LANG=en_US.UTF-8

Someone experiencing the same or know how to fix?

talboren:
close #4674


shahargl (approved):
Lgtm

PR #4678, opened 2025-05-01
feat: auto-downgrading revision

### LINKED ISSUE #4665: [➕ Feature]: Implement "auto downgrades"
Description
Currently, when a deployment is rolled back after database migrations have been applied, Keep fails to start because the database schema is ahead of the application code. This creates a critical failure scenario where the application cannot recover without manual intervention.
## Current behavior:

User deploys `v2` of the application
Keep runs and executes the `v2`migrations successfully
User encounters a bug and initiates a rollback to `v1`
Keep attempts to start `v1` but fails because the database schema is at `v2` state

This forces users to manually downgrade the database or perform other workarounds, resulting in extended downtime and potential data issues.

## Proposed Solution
Implement an automatic database schema detection and downgrade mechanism that runs during application startup to ensure the database schema is compatible with the running code version.
Technical Approach

Add a startup check that determines if the current database schema version (from alembic_version table) matches the expected version for the running code.
If a version mismatch is detected (database ahead of code), automatically execute a downgrade to the appropriate revision.
Implement a version mapping system that ties application versions to their corresponding alembic revisions.

```python
# Pseudo-code for the implementation
def ensure_db_schema_compatibility():
    # Get current database revision from alembic_version table
    current_db_revision = get_current_db_revision()
    
    # Get expected revision for this application version
    expected_revision = get_expected_revision_for_app_version()
    
    # If database is ahead of code (rollback scenario)
    if current_db_revision != expected_revision:
        log.warning(f"Database schema ({current_db_revision}) doesn't match application version ({expected_revision})")
        
        # Execute downgrade to match application version
        execute_alembic_downgrade(expected_revision)
        
        log.info(f"Successfully downgraded database schema to {expected_revision}")
```
Integrate this check into the application startup sequence before any database operations are attempted.

## Benefits

Zero-downtime rollbacks: Application can recover automatically from failed deployments
Reduced operational overhead: Eliminates the need for manual database interventions during rollbacks
Improved reliability: Ensures database schema and application code stay in sync even during deployment issues

## Implementation Considerations

Ensure all migrations have proper downgrade() implementations
Add comprehensive logging for the downgrade process
Consider adding a configuration flag to disable automatic downgrade in specific environments
Add monitoring/alerting when automatic downgrades occur
Create a test suite that verifies the downgrade functionality works as expected

## Acceptance Criteria

- Keep can be rolled back from v2 to v1 and automatically downgrade the database schema
- The process is properly logged for debugging and audit purposes
- Comprehensive tests verify the functionality works across multiple version transitions
- Documentation is updated to explain the automatic downgrade capability

korolenkowork:
Closes #4665

## 📑 Description
This PR introduces automatic database downgrading support when rolling back KeepHQ to a previous version.

## ✅ Checks
- [x] Main logic completed
- [x] New logic covered by test
- [x] E2E is written

## ℹ Additional Information

- From this commit onward, all migration files will be copied to the SECRET_MANAGER_DIRECTORY to enable safe database downgrades during rollback scenarios.
- A new environment variable ALLOW_DB_DOWNGRADE has been added to explicitly allow or prevent database downgrades, reducing the risk of accidental downgrades.


shahargl:
hey @EnotShow - thanks for this PR!

as this PR introduces some major changes to how Keep works, I thought I would add some e2e tests for it (mainly install Keep, do some migration, then downgrade). wdyt?

shahargl (changes requested):
see comments

EnotShow:
Hi @shahargl, @talboren!
So, I've completed this. I created the new workflow for migration tests, and it’s working overall. However, there's an issue: there’s no earlier version of KeepHQ that supports database downgrading. As a result, the final stage of the workflow is failing.

tests/migrations/alembic.ini
```diff
@@ -0,0 +1,51 @@
+[alembic]
```
shahargl, line 1: shouldn't we use the actual `alembic.ini`? because if someone will change the original one, the tests won't "test" the real case.

tests/migrations/env.py
```diff
@@ -0,0 +1,136 @@
+import asyncio
```
shahargl, line 1: shouldn't we use the actual `env.py`? because if someone will change the original one, the tests won't "test" the real case.

shahargl:
@EnotShow some impressive work! let's get it done.

shahargl:
@EnotShow I see you are still working on it? need help?

EnotShow:
Actually, I’ve just completed the e2e pipeline. So now I just need to fix a small issue with the path to the migrations folder, and then it will be ready for review.

EnotShow:
@shahargl I've finally finished wrestling with the pipelines — it's all working now and ready for review

EnotShow:
Downgrade Flow:

1. On every migration run, copy the migrations to local storage.
2. In case a downgrade is needed, restore the migrations from local storage and perform the downgrade.

E2E Pipeline Flow:

1. Run an older version of Keep.
2. Upgrade to the source version.
3. Run all tests to verify the app works correctly after the migrations.
4. Add a dummy migration.
5. Run the source version again to test the downgrade process.

shahargl:
@EnotShow gonna review this now

keep/api/core/db_on_start.py
```diff
@@ -167,23 +169,99 @@ def try_create_single_tenant(tenant_id: str, create_default_user=True) -> None:
...
+        context = MigrationContext.configure(connection)
+        return context.get_current_revision()
 
-def migrate_db():
+def copy_migrations(app_migrations_path, local_migrations_path):
+    """Copy migrations to a local backup folder for safe downgrade purposes."""
+    source_versions_path = os.path.join(app_migrations_path, "versions")
+
+    # Ensure destination exists
+    os.makedirs(local_migrations_path, exist_ok=True)
```
shahargl, line 183: add try/except, in certain circumstances (e.g. openshift) this will fail to create the directories

keep/api/core/db_on_start.py
```diff
@@ -167,23 +169,99 @@ def try_create_single_tenant(tenant_id: str, create_default_user=True) -> None:
...
+            os.remove(file_path)
+
+    # Copy new migrations
+    for item in os.listdir(source_versions_path):
+        src = os.path.join(source_versions_path, item)
+        dst = os.path.join(local_migrations_path, item)
+        if os.path.isdir(src):
+            shutil.copytree(src, dst, dirs_exist_ok=True)
+        else:
+            shutil.copy(src, dst)
```
shahargl, line 198: can you add explain what's the process here?

shahargl: like - why we do that?

keep/api/core/db_on_start.py
```diff
@@ -167,23 +169,99 @@ def try_create_single_tenant(tenant_id: str, create_default_user=True) -> None:
...
 
-    logger.info("Running migrations...")
-    config_path = os.path.dirname(os.path.abspath(__file__)) + "/../../" + "alembic.ini"
+    config_path = config_path or os.path.dirname(os.path.abspath(__file__)) + "/../../" + "alembic.ini"
     config = alembic.config.Config(file_=config_path)
     # Re-defined because alembic.ini uses relative paths which doesn't work
     # when running the app as a pyhton pakage (could happen form any path)
+
+    # This path will be used to save migrations locally for safe downgrade purposes
+    local_migrations_path = os.environ.get("SECRET_MANAGER_DIRECTORY", "/state") + "/migrations"
```
shahargl, line 234: use another env var for that, default to `/tmp` cuz it has permissions to write

keep/api/core/db_on_start.py
```diff
@@ -167,23 +169,99 @@ def try_create_single_tenant(tenant_id: str, create_default_user=True) -> None:
...
+    if current_revision and expected_revision and current_revision == expected_revision:
+        logger.info("Database schema is up-to-date!")
+        return None
+
+    logger.warning(f"Database schema ({current_revision}) doesn't match application version ({expected_revision})")
+    logger.info("Running migrations...")
+    try:
+        alembic.command.upgrade(config, "head")
+    except Exception as e:
+        logger.error(f"{e} it's seems like KeepHQ was rolled back to a previous version")
```
shahargl, line 255: `KeepHQ` -> `Keep`

shahargl (changes requested):
see comments

shahargl:
@EnotShow let tests pass and if they will I'll review

EnotShow:
Hey @shahargl, it's resolved!

I also added a separate Docker Compose file for E2E migrations testing, which attaches a dedicated volume. Because it seems that accessing the same volume from different workflows was causing concurrent map writes, like [this](https://github.com/keephq/keep/actions/runs/15328509161/job/43129213716).

tuantran0910:
Hi @shahargl, @talboren, can we get this PR merged ? I think it would be a good feature and we are planning to deploy Keep in our production environment.

korolenkowork:
@shahargl @talboren any updates on this?

shahargl:
Hey @korolenkowork, honestly it looks good but it requires some big effort for me to review it’s ready. This one touches the main flow so I’m bit worried.

shahargl:
closing as "not for now"

PR #4709, opened 2025-05-06
fix: Replace alerts_count calculation with subquery and add recalculation migration

### LINKED ISSUE #4477: [🐛 Bug]: Count of alerts displayed in an incident does not match the real count
Hello ! I've got a strange case, I only noticed since yesterday. I'm using build 2586a4. In the /incident page or inside an incident /incident/<ID>, the count of alerts is wrong.

It looks like, it's often -1 (still trying to confirm if they are all -1).

Strange thing is, when you load the /incident/<ID> page, it does a request on /alerts and the count is good there. In my below example, I have 3 alerts in an incident, incident displays 2 :

/alerts response
{
    "limit": 20,
    "offset": 0,
    "count": 3,
    "items": [
        {

Image
Image

Some, are correct : 

Image

VladimirFilonov:
Closes #4477 

## ✅ Checks

### DIFF
--- a/keep/api/core/db.py
+++ b/keep/api/core/db.py
@@ -42,6 +42,7 @@ from sqlalchemy.exc import IntegrityError, OperationalError
 from sqlalchemy.orm import foreign, joinedload, subqueryload
 from sqlalchemy.orm.exc import StaleDataError
 from sqlalchemy.sql import exists, expression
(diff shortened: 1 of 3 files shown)


talboren (approved):
LGTM

PR #4783, opened 2025-05-14
fix: use 0.01 for 0s executions and use log scale in wf graphs

### LINKED ISSUE #4778: [🐛 Bug]: Executions Graph is currently not displayed correctly
**Describe the bug**
There seems to be a problem with the displaying of the Executions Graph.

**Screenshots**
Image

Image

Kiryous:
Closes #4778 

Problem: if workflow executed in 0s, graph will not show it as success.
Fix: return 0.01 instead of 0.

Long term we might want to change seconds to milliseconds or something like that.

## Before
Screenshot 2025-05-14 at 16 25 31
Screenshot 2025-05-14 at 16 30 15

## After
Screenshot 2025-05-14 at 16 25 50
Screenshot 2025-05-14 at 16 25 59


talboren (approved):
LGTM

PR #4793, opened 2025-05-14
feat: Add threshold support for Correlation Rule

### LINKED ISSUE #4318: [➕ Feature]: Incident creation conditional to the number of occurrences of alert
**Context :** 

I have a wehbook provider, that pushes alerts from a jenkins server.

I would like, with correlation rules, to identify cases where I would have multiple incidents on the same thing, for instance :
data : Failure A
data : Failure A
data : Failure B
I would just like to create an incident, on the 2 alerts A. Incident creation becoming conditional to the multiple occurrence of an alert.

My issue is, if I do that with the correlation, and I take all my alerts, and choose to group on particular fields, it will create a huge number of incidents, also some that have 1 alert linked.
And I would like to create an incident, only if the resulting amount of correlated alerts within that incident would be greater than x.

**Alternative :**

Another alternative could be in the correlation, to filter on alerts that were grouped, maybe by fingerprint ? (Section grouping, of the doc https://docs.keephq.dev/overview/fingerprints#grouping)

VladimirFilonov:
Closes #4318

## ✅ Checks

### DIFF
--- a/keep-ui/app/(keep)/rules/CorrelationSidebar/CorrelationForm.tsx
+++ b/keep-ui/app/(keep)/rules/CorrelationSidebar/CorrelationForm.tsx
@@ -264,6 +264,41 @@ export const CorrelationForm = ({
             )}
           />
         </div>
+
+        <div>
+          <label
+            className="flex items-center text-tremor-default font-medium text-tremor-content-strong"
+            htmlFor="threshold"
+          >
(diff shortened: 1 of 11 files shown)


Matvey-Kuk (approved):

Kiryous (approved):
LGTM! Long awaited feature

PR #4680, opened 2025-05-02
feat: Pagerduty Provider: incident enrichments #4663

bbcorp:
Closes #4663

## 📑 Description
PagerDuty Provider: Add useful metadata via enrichment

## ✅ Checks

- [ ] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [ ] All the tests have passed

## Additional Information
**For the future:** maybe add a parameter in the provider to add this enrichment.
For all provider in fact.

Issue #4663, opened 2025-04-30
[➕ Feature]: PagerDuty Provider - add useful metadata via enrichment


bbcorp:
**Is your feature request related to a problem? Please describe.**
Some useful information on PagerDuty incidents or alerts are missing.
We can gather them with some workflows but it's tedious.

**Describe the solution you'd like**
Add any useful metadata from PagerDuty incidents or alerts.
i.e.: provider name, URL of the external incident, service ID, yada yada 

**Describe alternatives you've considered**
Use of workflows. But its tedious and slow

shahargl:
@bbcorp assigned

talboren:
Re-opening this, since it was merged and reverted (the PR wasn't ready..)

PR #4811, opened 2025-05-15
fix: Revert "feat: Pagerduty Provider: incident enrichments #4663"

Matvey-Kuk:
Reverts keephq/keep#4680

close https://github.com/keephq/keep/issues/4812


skynetigor (approved):
LGTM

PR #4820, opened 2025-05-16
feat: Make elastic index refresh strategy configurable via env variable

### LINKED ISSUE #4819: [➕ Feature]: add env var config option making elasticsearch refresh on index optional
Currently the index operations from the elasticsearch client always uses `refresh="true"`

Elasticsearch cluster also has a refresh interval (1s default). Under some circumstances it is preferable to rely on the interval cluster-wide refresh instead of always refresh after document index.

In our environment the PUT request times dropped at least an order of magnitude after setting `refresh=None`

**Proposed Solution**
Configurable toggle via env var `ELASTIC_DISABLE_REFRESH` which continues to default to `"true"` to ensure backwards compatibility. When set (to any value) calls to elasticsearch client (`.index` or `.bulk`) get `refresh=None`.

The lines effected are [in keep/api/core/elastic.py](https://github.com/search?q=repo%3Akeephq%2Fkeep%20refresh%3D%22true%22&type=code)

VladimirFilonov:
Closes #4819

## ✅ Checks

### DIFF
--- a/keep/api/core/elastic.py
+++ b/keep/api/core/elastic.py
@@ -50,6 +50,7 @@ class ElasticClient:
         if not self.enabled:
             return
 
+        self.refresh_strategy = os.environ.get("ELASTIC_REFRESH_STRATEGY", "true")
         self.api_key = api_key or os.environ.get("ELASTIC_API_KEY")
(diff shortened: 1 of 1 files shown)


talboren (approved):
LGTM

Issue #3680, opened 2025-02-25
[🐛 Bug]: Sometimes I can run workflows, sometimes I can't


Matvey-Kuk:
Such a workflow:

<img width="550" alt="Image" src="https://github.com/user-attachments/assets/64c05853-75ab-4290-b631-7cf8e1682e8b" />

```
workflow:
  id: 236fe306-560b-4453-a1bf-c3ec160dfab5
  name: incident-update-to-slack
  description: Share Incident Updates to Slack
  disabled: false
  triggers:
    - events:
        - created
      type: incident
  consts: {}
  owners: []
  services: []
  steps: []
  actions:
    - name: slack-action
      provider:
        config: "{{ providers.keephq }}"
        type: slack
        with:
          channel: C06PF9TCWUF
          message: "New incident: {{ incident.user_generated_name }} https://platform.keephq.dev/incidents/{{incident.id}}"
```

I can click "Test run" but not "Run" from the editor. I also can't run it from the list:

<img width="415" alt="Image" src="https://github.com/user-attachments/assets/b8caed35-4713-4e5c-ad30-375b033cae98" />

But If I test-run it from the editor, I get an exception:

<img width="1582" alt="Image" src="https://github.com/user-attachments/assets/40c7d235-6cb7-40e8-87c4-6e96d0a61a75" />

Matvey-Kuk:
Also: <img width="407" alt="Image" src="https://github.com/user-attachments/assets/22258b4b-9fa0-49d6-8d5d-6cef0c68bd84" />

Kiryous:
Now you will be prompted with required incident fields, if 
<img width="2048" alt="Image" src="https://github.com/user-attachments/assets/0b28d309-37e5-407b-88eb-84e2acd21ed3" />

Also, the browser label suggesting why you can't run a workflow:

<img width="630" alt="Image" src="https://github.com/user-attachments/assets/0de95b56-efa5-4b16-85b2-fac53e6bd7fa" />

PR #4837, opened 2025-05-19
fix: Fix isProviderInstalled logic and improve shared utils with JSDoc and tests

### LINKED ISSUE #4836: [🐛 Bug]: isProviderInstalled function returns opposite result due to negated condition
**Describe the bug**
The isProviderInstalled function in keep-ui/shared/lib/provider-utils.ts had a logical error where it was incorrectly checking if a provider was installed. The function was
using a negated some() condition (!Object.values(providers).some(...)) which produced exactly the opposite of the intended result.

**To Reproduce**
1. Use the isProviderInstalled function to check if a provider is installed
2. Pass a provider that has installed: false but has configured instances of the same type in the providers array
3. The function would incorrectly return false (indicating not installed) when it should return true

**Expected behavior**
The function should return true if either:
1. The provider's installed flag is true, OR
2. There exists at least one configured provider of the same type with a non-empty config

**Actual behavior**
The function was returning true only if:
1. The provider's installed flag is true, OR
2. There are NO configured providers of the same type with a non-empty config

This is the exact opposite logic for the second condition, causing the function to incorrectly report providers as not installed when they should be considered installed.

**Impact**
This issue would cause the UI to incorrectly show providers as not installed even when they had configured instances, potentially leading to duplicate provider installations or confusion for users who had already set up a provider.

lukeod:
Closes #4836
Possibly related #3680 
## 📑 Description
This PR fixes a critical bug in the `isProviderInstalled` function where a negated condition was producing the opposite of the intended result. The bug caused the UI to incorrectly indicate providers as not installed when they were actually configured.

Changes included:
- [x] Fix reversed logic in `isProviderInstalled` function
- [x] Add comprehensive JSDoc comments to utility functions
- [x] Migrate `buildNestedObject` function to `object-utils.ts`
- [x] Add extensive unit tests for utility functions

## ✅ Checks

## ℹ Additional Information
The bug in `isProviderInstalled` was a simple but critical reversal of logic:
```diff
return (
  provider.installed ||
- !Object.values(providers || {}).some(
+ Object.values(providers || {}).some(
    (p) =>
      p.type === provider.type && p.config && Object.keys(p.config).length > 0
  )
);
```
This was discovered while adding test coverage to shared utility functions.

Note: Apologies for the PR size. Given that most changes are non-functional, and are generally contained in the one area, I thought that splitting them would probably just result in more work for the reviewer(s). Happy to split if preferred.


keep-ui/shared/lib/provider-utils.ts
```diff
@@ -1,12 +1,23 @@
...
   provider: Pick<Provider, "type" | "installed">,
   providers: Provider[]
 ) {
   return (
     provider.installed ||
-    !Object.values(providers || {}).some(
+    Object.values(providers || {}).some(
       (p) =>
         p.type === provider.type && p.config && Object.keys(p.config).length > 0
     )
```
talboren, line 23: In what use case is this being used? It sounds a bit weird - a provider will be marked as installed if it has other provider of the same type installed? o.O

talboren (commented):
Thank you so much for that, looks like great work! Left 1 comment :)

keep-ui/shared/lib/provider-utils.ts
talboren: @lukeod you mentioned a "critical bug" that is being fixed because of a negated condition: can you mention where this bug appears? maybe add a screenshot on how this fix solves it? FYI you have all e2e tests failing - so I guess something needs to be fixed in there too..

lukeod: After some more investigation, i think i overstated the impact, and was too hasty with proposing that change. 

I think the existing logic represents an issue with the provider states, and may cause some unexpected behavior in workflows. But clearly it needs more validation - it seems that the e2e tests rely on the current logic.

I'll close this PR and submit a revised one that doesn't change that behavior. I'll do some more digging later to better understand the provider state lifecycle.

PR #4846, opened 2025-05-20
fix: self assign should dispose on new alert

talboren:


talboren:
Closing to propose better solution

PR #4858, opened 2025-05-21
fix: no workflows found for execution from alert or incident

### LINKED ISSUE #4857: [🐛 Bug]: No workflows found for execution from alert or incident
After updating to 0.44.1, the workflows are not triggerable manually anymore : (capture)
I can still see them all in Workflows, just can't trigger them from an incident for instance.
My workflow have a manual trigger

Image

Kiryous:
Closes #4857


skynetigor (approved):
LGTM

PR #4864, opened 2025-05-21
fix: lookup_by_name if updating workflows from a workflow

### LINKED ISSUE #4863: [🐛 Bug]: lookup by name when updating workflow from workflow
Use case:
User stores workflow yamls on s3 and want to regularly sync them from s3 to keep. He uses https://github.com/keephq/keep/blob/main/examples/workflows/update_workflows_from_s3.yml

Problem:
The workflow will duplicate every workflow because internally it looks up by id and id is random uuid assigned at creation.

Additionally, if user tries to use optional step to delete previous workflows, he will see error:

```yaml
    # optional: delete all other workflows before updating for full sync
     - name: delete-all-other-workflows
       provider:
         type: keep
         with:
           delete_all_other_workflows: true
```

Image

Kiryous:
Closes #4863


talboren (approved):
LGTM, adding some unit tests to cover this would've been nice though!

Kiryous:
@talboren added tests!

talboren (approved):
LGTM

skynetigor (approved):
LGTM

PR #4873, opened 2025-05-22
fix: status not getting updated after acknowledged

### LINKED ISSUE #4839: [🐛 Bug]:Alert status not getting updated after acknowledged
**Describe the bug**
When we self assign the alert, its status is changing to acknowledged. And after that even we get the resolved alerts, status is not getting change automatically to resolved.

**Expected behavior**
Upon receiving the resolved alerts, alert should be marked as resolved.

talboren:
close #4839


shahargl (approved):
lgtm

PR #4914, opened 2025-05-26
test: add tests for invoking provider methods

### LINKED ISSUE #4913: [🐛 Bug]: provider method invocation errors all return 500
Original issue is lost, and the status code should be 400, not 500

Image

Image

Kiryous:
Closes #4913

### DIFF
--- a/tests/test_providers_api.py
+++ b/tests/test_providers_api.py
@@ -1,0 +1,367 @@
+import json
+import pytest
+from datetime import datetime
+from unittest.mock import Mock, patch, MagicMock
+from fastapi.testclient import TestClient
+from sqlmodel import Session
+
+from keep.api.core.dependencies import SINGLE_TENANT_UUID
+from keep.api.models.db.provider import Provider
+from keep.providers.base.provider_exceptions import ProviderMethodException
+from keep.providers.providers_factory import ProviderConfigurationException
+from tests.fixtures.client import client, setup_api_key, test_app  # noqa
+
+VALID_API_KEY = "valid_api_key"
+
+
+@pytest.fixture
+def mock_provider_in_db(db_session: Session):
+    """Create a mock provider in the database."""
+    provider = Provider(
+        id="test_provider_id",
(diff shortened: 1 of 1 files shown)


keep/api/routes/providers.py
```diff
@@ -648,85 +649,145 @@ def invoke_provider_method(
...
+            },
         )
         raise HTTPException(status_code=500, detail="Internal server error") from e
 
 
+def _get_provider_configuration(
+    tenant_id: str, provider_id: str, body: dict | None, session: Session
+) -> dict:
+    """
+    Extract provider configuration from database or request body.
```
talboren, line 734: in what cases will we extract it from the request body?

Kiryous: https://github.com/keephq/keep/blob/28e5ff33001acbb43490c7687172393166a85e29/keep-ui/features/workflows/builder/ui/Editor/StepTest.tsx#L20-L30

talboren: still a bit confusing because i'm not sure how can you test a provider that is not configured? where does the configuration come from?

keep/api/routes/providers.py
```diff
@@ -648,85 +649,145 @@ def invoke_provider_method(
...
-            )
 
-    try:
+        # Step 2: Initialize provider and invoke method
+        context_manager = ContextManager(tenant_id=tenant_id)
         provider_instance = ProvidersFactory.get_provider(
-            context_manager, provider_id, provider_type, provider_config
+            context_manager,
+            provider_config["provider_id"],
+            provider_config["provider_type"],
```
shahargl, line 664: `.get`

keep/api/routes/providers.py
Kiryous: On a closer look, the logic can be simplified. For default providers such as console, bash, python we will try to parse id, for others — get config from db. No more `providerInfo` in body.

talboren (approved):
LGTM

PR #4952, opened 2025-06-01
feat: add `from_timestamp` keep function

### LINKED ISSUE #4951: [🐛 Bug]: `from_timestamp` keep function
If user receives date as a unix timestamp (e.g. query prometheus), and wants to use it as text in the next step (e.g. send message to slack/teams), it's not possible to use date functions as this timestamp is not a date.

Kiryous:
Closes #4951


shahargl (approved):
Lgtm

PR #4956, opened 2025-06-01
fix: Pagination not reset after filtering   results in empty page

### LINKED ISSUE #4785: [🐛 Bug]: Pagination not reset after filtering - results in empty page
**Describe the bug**
When using pagination in the alert list and navigating to a later page (e.g., page 10), applying a filter afterward doesn't reset the pagination index. If the filtered result has fewer pages (e.g., 5 pages), the user remains on an invalid page index (e.g., page 10 of 5), resulting in an empty screen.

**To Reproduce**
1. Go to the Alerts screen  
2. Navigate to a high-numbered page
3. Apply a filter that reduces the number of pages to less than that
4. Observe that the current page is blank and the pager says e.g., "Showing page X of (Y<X)"

**Expected behavior**
After applying a filter, the pagination should reset to the first available page or the last page if the current one is out of bounds.

skynetigor:
Closes #4785

## 📑 Description
- Remove unnecessary AlertPaginationServerSide component
- Fix not resetting pagination for Alerts after filters are applied
- Fix not resetting pagination for Workflows after filters are applied
- Fix glitch with CEL on Alerts page caused by re-rendering of FacetsPanelServerSide component when searchCel changes.

## ✅ Checks


Kiryous (approved):
LGTM!

PR #4963, opened 2025-06-03
fix: return 400 from /run or /test if workflow is invalid

### LINKED ISSUE #4959: [🐛 Bug]: `{workflow_id}/run` endpoint returns 500 if there's an error in yaml def
If you will try to run workflow with the following provider config:
```yaml
  steps:
    - name: trace-logs
      provider:
        type: webhook
        config: "{{ providers['Quickstart Webhook'] }}"
        with:
```
you will only see uninformative`Failed to run workflow` error:
Image

Only admins can see actual problem:

Image

Kiryous:
Closes #4959


Kiryous:
@greptileai

PR #4973, opened 2025-06-04
feat: auto incident assignee

talboren:
CleanShot 2025-06-04 at 10 45 38@2x
CleanShot 2025-06-04 at 10 20 54@2x


shahargl (approved):
lgtm

PR #4989, opened 2025-06-08
fix: A bug was addressed where CEL queries in the URL were not preserved on page reload.

### LINKED ISSUE #4982: [🐛 Bug]: CEL queries update URL with query params but they do not survive page load/reload
Example, I search for `message.contains("CPU")`

My URL in address bar becomes `http://localhost:3000/alerts/feed?cel=message.contains%28%22CPU%22%29`

When I page reload (or copy+paste to share this results page with a friend) then what happens is that the alerts feed page loads and then my address bar gets re-written as `http://localhost:3000/alerts/feed` and my query is lost/never ran

talboren:
Close #4982 

A bug was addressed where CEL queries in the URL were not preserved on page reload.

The `useCelState` hook in `keep-ui/features/cel-input/use-cel-state.ts` was modified:
*   `useEffect` cleanup now prevents the `cel` query parameter from being removed on initial page load.
*   Cleanup logic was refined to only trigger when the pathname genuinely changes during navigation, not on component re-renders or initial mount.
*   The `router` object was added to the dependency array of the `useEffect` responsible for updating the URL.

The `AlertsRulesBuilder` component in `keep-ui/features/presets/presets-manager/ui/alerts-rules-builder.tsx` was updated:
*   The `useCelState` hook now prioritizes the `cel` query parameter from the URL over the preset's default CEL rules when initializing the state.

A failing test in `keep-ui/features/cel-input/__tests__/use-cel-state.test.ts` was updated to reflect the new, correct behavior of the `useCelState` hook, where cleanup does not occur immediately on a pathname change.

These changes ensure CEL query parameters persist in the URL across page reloads, allowing filtered views to be shared and bookmarked.

### DIFF
--- a/keep-ui/features/cel-input/__tests__/use-cel-state.test.ts
+++ b/keep-ui/features/cel-input/__tests__/use-cel-state.test.ts
@@ -127,14 +127,18 @@ describe("useCelState", () => {
       })
     );
 
-    (usePathname as jest.Mock).mockReturnValue("/new/pathname");
+    (usePathname as jest.Mock).mockReturnValue("/alerts/feed");
 
-    const { result, unmount } = renderHook(() =>
+    const { result, rerender } = renderHook(() =>
       useCelState({ enableQueryParams: true, defaultCel: "" })
     );
 
-    unmount();
+    // Change the pathname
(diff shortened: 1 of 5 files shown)


talboren:
@greptile

PR #4990, opened 2025-06-09
Update providers_service.py

zcc0077:
1. method `install_providers` in providers_service.py need to use `provider_unique_id` as provider_id(currently use type) to save specific consumer provider to db and make sure when removing specific one instead of all consumers with same type.
2. method remove_consumer in event_subscriber.py wants `str` but given `object`: `provider_model`, will never take effective, need to send `provider_model.id`.
3. method `update_provider` in providers_service.py: need to re-add consumer to make sure configuration is latest in provider..

Closes # 

## ✅ Checks

- [ ✅] My pull request adheres to the code style of this project
- [ ✅ ] My code requires changes to the documentation
- [ ✅ ] I have updated the documentation as required
- [ ✅ ] All the tests have passed

### DIFF
--- a/keep/providers/providers_service.py
+++ b/keep/providers/providers_service.py
@@ -171,7 +171,7 @@ class ProvidersService:
         context_manager = ContextManager(tenant_id=tenant_id)
         try:
             provider = ProvidersFactory.get_provider(
-                context_manager, provider_id, provider_type, config
+                context_manager, provider_unique_id, provider_type, config
             )
         except Exception as e:
             raise HTTPException(status_code=400, detail=str(e))
@@ -298,6 +298,15 @@ class ProvidersService:
             provider.installed_by = updated_by
             provider.validatedScopes = validated_scopes
             provider.pulling_enabled = pulling_enabled
+
+            if provider.consumer:
+                try:
+                    event_subscriber = EventSubscriber.get_instance()
+                    event_subscriber.remove_consumer(provider.id)
+                    event_subscriber.add_consumer(provider_instance)
+                except Exception:
+                    logger.exception(f"Failed to update consumer provider id: {provider_id}, type: {provider.type}")
+            
             session.commit()
 
             logger.info(
@@ -348,7 +357,7 @@ class ProvidersService:
             if provider_model.consumer:
                 try:
                     event_subscriber = EventSubscriber.get_instance()
-                    event_subscriber.remove_consumer(provider_model)
+                    event_subscriber.remove_consumer(provider_model.id)
                 except Exception:
                     logger.exception("Failed to unregister provider as a consumer")
 


keep/providers/providers_service.py
```diff
@@ -171,7 +171,7 @@ def install_provider(
         context_manager = ContextManager(tenant_id=tenant_id)
         try:
             provider = ProvidersFactory.get_provider(
-                context_manager, provider_id, provider_type, config
+                context_manager, provider_unique_id, provider_type, config
```
shahargl, line 174: I don't understand this. How does it work today? I need some way to reproduce this.

keep/providers/providers_service.py
```diff
@@ -298,6 +298,17 @@ def update_provider(
             provider.installed_by = updated_by
             provider.validatedScopes = validated_scopes
             provider.pulling_enabled = pulling_enabled
+
+            if provider.consumer:
+                try:
+                    event_subscriber = EventSubscriber.get_instance()
+                    event_subscriber.remove_consumer(provider.id)
```
talboren, line 305: why do you need to first remove the consumer?

keep/providers/providers_service.py
```diff
@@ -298,6 +298,17 @@ def update_provider(
...
+                    event_subscriber.add_consumer(provider_instance)
+                except Exception:
+                    err_msg = f"Failed to update consumer provider with id: {provider_id}, type: {provider.type}"
+                    logger.exception(err_msg)
+                    raise HTTPException(status_code=400, detail=err_msg)
```
shahargl, line 310: add logs (info)

talboren:
can you add a test that covers this?

keep/providers/providers_service.py
```diff
@@ -348,7 +359,7 @@ def delete_provider(
             if provider_model.consumer:
                 try:
                     event_subscriber = EventSubscriber.get_instance()
-                    event_subscriber.remove_consumer(provider_model)
+                    event_subscriber.remove_consumer(provider_model.id)
```
shahargl, line 362: add logs

shahargl (changes requested):
see comments. in general, please add a test that will fail and this fix will make it pass.

other than that - great work!

keep/providers/providers_service.py
zcc0077: > why do you need to first remove the consumer?

The reason is when `ProvidersFactory.get_provider` invoke, a new provider instance is created, while consumer thread still use old provider instance, and there seems no proper way to update configurations in old provider instance. As far as I know, configurations transfer to provider only happens when provider instance instantiation(no parent method can send provider config to instance except instantiate method `__init__`).

keep/providers/providers_service.py
zcc0077: > I don't understand this. How does it work today? I need some way to reproduce this.

provider_id in `/install` is actually `provider type`(yes, use id to send type firstly due to id is generated in backend which make sense here) : 
image
and `id` in response is real provider id:
image
Why use provider_unique_id in this? Because when [deleting consumer provider](https://github.com/keephq/keep/blob/9c5dcecd6d2745add992ebae434cb86377fde5a9/keep/event_subscriber/event_subscriber.py#L85),  it use `provider id` in models from db to compare which is reasonable, for now, it use object to compare which is a bug and will keep running endlessly until the app is closed. One more, if change to use `type` to compare, it will delete all consumer providers with same `type` instead of the only specific one with `id` you want to delete which is catastrophic.

shahargl: But installation and deletion work, so I fail to understand why this only affects the `consuming` providers?

zcc0077: I mentioned [here](https://github.com/keephq/keep/issues/4991)
Delete doesn't work with consuming providers.
You can reproduce with simple consumer provider below, try connect it and disconnect after it push alert, you'll find it still pushing alert.
```
import time
import pydantic
import dataclasses
from keep.contextmanager.contextmanager import ContextManager
from keep.providers.base.base_provider import BaseProvider
from keep.providers.models.provider_config import ProviderConfig, ProviderScope
from keep.providers.providers_factory import ProvidersFactory


@pydantic.dataclasses.dataclass
class MytestProviderAuthConfig:
    param: str = dataclasses.field(
        metadata={
            "required": True,
            "description": "Test",
            "sensitive": False,
        }
    )


class MytestProvider(BaseProvider):
    PROVIDER_DISPLAY_NAME = "MyTestProvider"
    PROVIDER_CATEGORY = ["Monitoring"]

    def __init__(
            self, context_manager: ContextManager, provider_id: str, config: ProviderConfig
    ) -> None:
        self.consume = False
        print("my test provider init called=====================")

    def dispose(self):
        pass

    def validate_config(self):
        self.authentication_config = MytestProviderAuthConfig(
            **self.config.authentication
        )

    def start_consume(self):
        self.consume = True
        while self.consume:
            self._push_alert({"name":self.authentication_config.param})
            print(f'{self.PROVIDER_CATEGORY[0]}: {self.authentication_config.param}====================')
            time.sleep(5)

    def stop_consume(self):
        self.consume = False
        print("my test provider consume stop called====================")

    def _query(self, **kwargs: dict):
        pass


if __name__ == "__main__":
    import logging

    logging.basicConfig(level=logging.DEBUG, handlers=[logging.StreamHandler()])
    context_manager = ContextManager(
        tenant_id="singletenant",
        workflow_id="test",
    )

    provider_config = {
        "authentication":
            {
                "param": "myparam1"
            },
    }
    provider = ProvidersFactory.get_provider(
        context_manager=context_manager,
        provider_id="keep-mytest",
        provider_type="mytest",
        provider_config=provider_config,
    )

```

zcc0077:
https://github.com/usr-attachments/assets/92132cd8-7a29-4dd9-bf98-7a4992e730bf

@talboren @shahargl  here is screen record for how to reproduce this bug

shahargl:
@zcc0077 closing it. ping me on Slack when you have some news <3

PR #5031, opened 2025-06-16
test: foreach jumps to the end of workflow in UI Builder

### LINKED ISSUE #4996: [🐛 Bug]: foreach jumps to end of workflow
**Describe the bug**
I tried to add a step that happens after a foreach, but after validation, the added step "jumps" above the foreach.

**To Reproduce**
1. Adding a foreach with a step
2. Adding another step under the foreach
3. The validation works, and the step "jumps" above the foreach

**Expected behavior**
Validation allows a step after the foreach

Kiryous:
Closes #4996


Kiryous:
@greptileai

PR #5058, opened 2025-06-18
fix: Fix bug on incidents pagination

### LINKED ISSUE #5055: [🐛 Bug]: Changing Page, or Number of incidents per page, has conflicting logic
Hello Everyone ! I found something I'm able to trigger directly on the platform keep.
If you create > 20 incidents, it will always take two clicks / tries to :

- Change the number of incidents displayed
For instance, when changing from 20 to 100, I see a first request is fetching 100, then another instantaneously fetches again 20

Image

- Change page
Same spirit, when changing to page 2, a first request is sent out, with offset = 20, and is instantly overriden by one with offset = 0. 

Image

Also, in the alerts feed, we can't untick facets. We can click 'only', but can't untick, it has no effect front / back end. 

Image

skynetigor:
Closes #5055

## 📑 Description

- Move toggle/selectOne/selectAll functions to store from Facet component
- Write unit-tests for toggle/selectOne/selectAll in store
- Fix pagination bug on incidents

## ✅ Checks

### DIFF
--- a/keep-ui/features/filter/facet-value.tsx
+++ b/keep-ui/features/filter/facet-value.tsx
@@ -53,7 +53,7 @@ export const FacetValue: React.FC<FacetValueProps> = ({
       <div className="flex items-center min-w-[24px]">
         <input
           type="checkbox"
-          readOnly // Fixes "You provided a `checked` prop to a form field without an `onChange` handler."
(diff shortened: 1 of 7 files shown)


shahargl (approved):
lgtm

PR #5059, opened 2025-06-18
fix: Add filter to exclude test runs in workflow query

VladimirFilonov:
Introduced a condition to ensure workflows with `is_test_run` set to true are excluded from the query. This ensures only non-test workflows are processed while adhering to the existing limit per workflow.

Closes # 

## ✅ Checks


skynetigor (dismissed):
LGTM

Kiryous (approved):
LGTM!

PR #5061, opened 2025-06-18
fix: keep_provider _notify foreach

### LINKED ISSUE #5016: [🐛 Bug]: Nested foreach, the second foreach looping over the same results gives an error.
**Describe the bug**
when using foreach for 2 different action steps iterating over the results of the same previous step, when doing the second foreach it says that argument type of NoneType is not iterable.

**To Reproduce**

1. Create a workflow with a step that outputs a list (e.g., users).
2. Add two separate actions that each use foreach to iterate over the output of that step.
3. Observe that during the Workflow Execution Results, the first action executes correctly, but the second one fails with a NoneType error.

Example:

```yaml
actions:
    - name: add-to-DB
      foreach: "{{ steps.python-step.results.users }}"
      provider:
        type: postgres
        config: "{{ providers.postgres }}"
        with:
          query: query
    - name: create-alert
      foreach: "{{ steps.python-step.results.users }}"
      provider:
        type: keep
        config: "{{ providers.default-keep }}"
        with:
          override_source_with: elastic
          read_only: true
          alert:
            name: alert
            description: "{{}}"
            severity: info
            labels:
              description: "{{foreach.value.description}}"
          fingerprint: "{{foreach.value.user}}"
```

**Expected behavior**
Both actions should independently iterate over the same list without error.

**Screenshots**

Image

Kiryous:
Closes #5016 

# Context 
Creating alerts with Keep provider, the whole list of alerts is needed for business logic. Currently, when the `foreach` specified on action, action will run only with context of current item. For this reason, keep provider has the hack: look at the previous step results and assume it is the array of alerts user want to create.

# Problem
The logic is implicit, and the previous step might return dict, e.g. { 'users': [...] }, or nothing, like in case of original issue.

# Proposed solution
 Include the whole list of items in context while we iterating over it. In Keep provider get the `foreach_context.items` as `alert_results` and fallback to previous step results for backward compatibility.

### DIFF
--- a/docs/snippets/providers/graylog-snippet-autogenerated.mdx
+++ b/docs/snippets/providers/graylog-snippet-autogenerated.mdx
@@ -9,8 +9,8 @@ This provider requires authentication.
 - **verify**: Verify SSL certificates (required: False, sensitive: False)
 
 Certain scopes may be required to perform specific actions or queries via the provider. Below is a summary of relevant scopes and their use cases:
-- **authenticated**: Mandatory for all operations, ensures the user is authenticated. (mandatory)
-- **authorized**: Mandatory for querying incidents and managing resources, ensures the user has `Admin` privileges. (mandatory)
+- **authenticated**: Mandatory for all operations, ensures the user is authenticated. (mandatory) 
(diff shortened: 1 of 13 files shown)


Kiryous:
@greptileai

shahargl (approved):
lgtm

PR #5063, opened 2025-06-18
fix: Time delta in keep provider

### LINKED ISSUE #5060: [🐛 Bug]: Time_delta in keep provider
**Describe the bug**
When using the time_delta parameter in the Keep provider, the expected behavior is to filter alerts based on how many days ago they were triggered. However, specifying fractional days (e.g., 0.1667 for approximately 4 hours) does not appear to correctly filter alerts by time. Returns alerts triggered farther ago than the defined (e.g. 4 hours). Even when the time_delta is not fractional, an integer it also does not seem to work.

**To Reproduce**

```yaml
steps:
  - name: query-keep
    provider:
      type: keep
      config: "{{ providers.default-keep }}"
    with:
      version: 2
      time_delta: 0.1667  # Expecting this to represent ~4 hours
      filter: name == "XXXX"
```

**Expected behavior**
The time_delta: 0.1667 should filter alerts to include only those triggered within the last 4 hours (0.1667 days). Combined with the filter clause, the step should return only alerts with the name "XXXX" that were triggered within that time frame. But time_delta appears to have no effect. Alerts outside the expected time window are still returned.

**Screenshots**

Log:
Image

Yaml:
Image

One of selected alerts info:
Image

skynetigor:
Closes #5060

## ✅ Checks

### DIFF
--- a/keep/providers/keep_provider/keep_provider.py
+++ b/keep/providers/keep_provider/keep_provider.py
@@ -125,7 +125,7 @@ class KeepProvider(BaseProvider):
                 raise ValueError("Filter is required for version 2")
             try:
                 alerts = search_engine.search_alerts_by_cel(
-                    cel_query=filter, limit=limit or 100, timeframe=int(time_delta)
+                    cel_query=filter, limit=limit or 100, timeframe=float(time_delta)
                 )
             except Exception as e:
                 self.logger.exception(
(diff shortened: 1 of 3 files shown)


shahargl (dismissed):
👨‍🍳

PR #5119, opened 2025-06-30
test: test comparing numbers in `if` condition

Kiryous:


talboren:
@Kiryous is this still needed? can we make it pass and merge?

Kiryous:
@talboren it was intended as a demonstration of the issue for @shahargl. no fix yet, as changing of behavior will break existing workflows. closing for now.

PR #5134, opened 2025-07-02
fix: Add parentheses for search cel being sent to BE to fix searcCel/facetFilterCel logic

### LINKED ISSUE #5114: [🐛 Bug]: Facet Filters in Preset Do Not Apply Correctly (e.g., Status or Severity)
**Describe the bug**
When using a preset with a filter like status == "firing" || status == "acknowledged", the Facet filter section does not behave correctly. For example, clicking "only" on "acknowledged" still shows all alerts instead of narrowing down the view. Other facet filters such as severity also do not work when using a preset.

**To Reproduce**

1. Create a preset with a filter: status == "firing" || status == "acknowledged".
2. Go to the Facet filter section on the preset view.
3. Click "only" on the "acknowledged" status filter.
4. Observe that alerts from both statuses (or all alerts) are still shown, rather than just the acknowledged ones.
5. Try other facet filters (e.g., severity) — observe they also do not narrow results as expected.

**Expected behavior**
Clicking "only" on a facet value (e.g., "acknowledged" under status) should update the alert view to show only matching results, even when using a preset with pre-defined filters.

**Additional context**
version 0.45.9

skynetigor:
Closes #5114

## ✅ Checks

### DIFF
--- a/keep-ui/widgets/alerts-table/ui/__tests__/useAlertsTableData.test.ts
+++ b/keep-ui/widgets/alerts-table/ui/__tests__/useAlertsTableData.test.ts
@@ -1,0 +1,318 @@
+import { renderHook, act } from "@testing-library/react";
+import {
+  useAlertsTableData,
+  AlertsTableDataQuery,
+} from "../useAlertsTableData";
+import { AlertsQuery, useAlerts } from "@/entities/alerts/model";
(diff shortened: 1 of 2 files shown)


shahargl (approved):
lgtm

PR #5183, opened 2025-07-16
fix: Improve date manage in Windows Maintenance

### LINKED ISSUE #5184: [🐛 Bug]: Date in Maintenance Windows doesn't work
**Describe the bug**
Once a window is created and set as enable, the start time doesn't matter and the maintenance window is taking into account from now, instead of waiting to the start time.

Moreover, there is a mismatch between the Calendar hour format selected (typically the user will create the window base on the current (local) hour. Despite of that, in the panel it takes the UTC format, in this way you can create a window for the 3 PM and suddenly you will se in the right panel this new window as 1PM (Taking into account GMT+2) 

**To Reproduce**

1. Go to Maintenance Windows section
2. Fill in the attributes
3. Select a future Start Date in the Calendar
4. **Here, you  will see a mismatch with the picked hour in the right Panel**
5. Select both switch (Supressed and Enable)
6. Send an alert with the same CEL to match with the windows one.
7. **In this point, the alert will show as Supressed, despite the start date of the windows is later.**

**Expected behavior**

**Screenshots**
Creation step:

Image

Panel with mismatch hour:

Image

diegovb-sys:
closes #5184

## 📑 Description
In order to improve the use of data by the windows maintenance I've introduced 2 minor fixes:

1. I've added as a new filter the start_date, in this way this fix a bug which makes the date was not taking into account.
2. As the date in calendar is set to Local, it makes sense that the right panel use the same format. 

As I added a new filter, the tests should be changed to take this into account.

## ✅ Checks


shahargl:
@diegovb-sys can you sign the CLA?

shahargl (approved):
lgtm

diegovb-sys:
> @diegovb-sys can you sign the CLA?

Sure! Now should be signed.

Ty!

shahargl:
merged. thanks @diegovb-sys and welcome to Keep contributors group :P

diegovb-sys:
The pleasure is mine! @shahargl ! I think we will see each other often 😅

Issue #5047, opened 2025-06-17
[🐛 Bug]: CEL filters not returning alerts with dismissed: false after dismissedUntil expires


lucaspgz:
When an alert is dismissed using a workflow (by enriching dismissed: true and dismissedUntil), and that dismissal expires (i.e., the dismissedUntil time passes and the alert is set back to dismissed: false), the alert no longer appears when filtering by dismissed: false, even though its payload shows dismissed: false.

This affects both the sidebar filters in the alert feed and CEL-based filters using dismissed == false.

To Reproduce

 1 Send an alert with dismissed: false.
 2 Apply a workflow that enriches the alert with:
        dismissed: true
        dismissedUntil: [future timestamp]
        disposable_dismissed, disposable_dismissedUntil, disposable_note, and disposable_status
 3  Wait until the dismissedUntil timestamp has passed.
 4 Try filtering the alert feed with:
        The sidebar filter “Not dismissed”
        A CEL filter like dismissed == false

Alerts with dismissed: false should appear in the feed and be returned when filtering with dismissed == false, regardless of any prior dismissal period that has already expired.

lucaspgz:
Hello!
While testing this issue further, we found the possible problem:
The function that undismisses alerts after the dismissedUntil period expires does not clear the disposable_dismissed field.

As a result, when the dismissal period ends, dismissed is correctly set back to false, but disposable_dismissed remains true, causing the alert to still be treated as dismissed.

I tested this by manually setting the disposable_dismissed field to false via the API, and after that, the alert correctly appeared as not dismissed.

So i don't think the problem was the CEL evaluation, but rather the function that undismisses the alert

PR #5188, opened 2025-07-16
fix: cleaning of fields after dismissUntil expiration

lucaspgz:
Closes https://github.com/keephq/keep/issues/5047

📑 Description
This pull request addresses two main points:

Fixes the bug related to cleaning the dismissed, dismissUntil, and disposable_* fields of an alert after dismissUntil expiration, and also restores the alert's previous status (sent in batch_enrich now) when dismissUntil expires.
Fixes the bug in the UI where selecting two alerts in the feed, even if both are dismissed, triggers the dismiss pop-up, instead of the restore one in that case.
✅ Checks
 My pull request adheres to the code style of this project
 My code requires changes to the documentation
 I have updated the documentation as required
 All the tests have passed
ℹ Additional Information
It has been tested both with dedicated tests and via direct alert sending through API, dismissing them with a workflow, and also via the UI tool. It works correctly for single or multiple alerts, as needed.


shahargl:
this is solved with a better approach by https://github.com/keephq/keep/pull/5270

PR #5258, opened 2025-08-18
fix: changed get alert by fingerprint

### LINKED ISSUE #5255: [🐛 Bug]: Route get alerts/{fingerprint} doesn't work as expected, can't fetch an alert that isn't in the last 1000 alerts
**Describe the bug**
the function that gets the alert by fingerprint uses a function that fetches the last 1000 alerts (get_last_alerts)

**To Reproduce**
try to get an alert by fingerprint that isn't from the last 1000 alerts 

**Expected behavior**
the route should use the function get_last_alerts_by_fingerprints
**Screenshots**

Image
fix: 

Image

**Additional context**
I think the same behavior is repeated when trying to query a preset of "old" alerts

DolevNe:
Closes #5255

## 📑 Description
the function that gets the alert by fingerprint uses a function that fetches the last 1000 alerts (get_last_alerts)
the route should use the function get_last_alerts_by_fingerprints

## ✅ Checks

### DIFF
--- a/keep/api/routes/alerts.py
+++ b/keep/api/routes/alerts.py
@@ -757,8 +757,8 @@ def get_alert(
             "tenant_id": tenant_id,
         },
     )
-    all_alerts = get_all_alerts(authenticated_entity=authenticated_entity)
-    alert = list(filter(lambda alert: alert.fingerprint == fingerprint, all_alerts))
+    all_alerts = get_alerts_by_fingerprint(tenant_id=tenant_id,fingerprint=fingerprint)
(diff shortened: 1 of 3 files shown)


talboren:
@DolevNe can you add a quick unit test to make sure we don't fall here again? also - what element in the ui uses this api? lets make sure it works properly too

shahargl (changes requested):
looking at the test - wouldn't it pass regardless of the fix? if I understand correctly - the bug occur when > 1000 alerts?

PR #5300, opened 2025-09-03
fix: incident ticket linking modal ticket_id field to be optional

### LINKED ISSUE #5299: [🐛 Bug]: ticket id in ticket linking to incident option isn't actually optional
**Describe the bug**
when trying to link an incident to an existing ticket it seems that the ticket id field is optional and it will let you submit, but it wont show the incident is linked to a ticket afterwards
**To Reproduce**
try to connect an incident to a ticket and don't mention a a ticket id

DolevNe:
Closes #5299 

## 📑 Description
the incident linking option was looking for the 'ticket_id" attribute instead of the "ticket_url" one (which is required), so i changed it
## ✅ Checks


shahargl (approved):
lgtm

PR #5337, opened 2025-09-29
fix: Searching by incident.id in CEL filter

### LINKED ISSUE #5293: [🐛 Bug]: Incident null in Alert table
**Describe the bug**
As far as I've checked, the table Alerts contains in the CEL event, among others, the incident field.

Image

Once an incident is created, this field continues being null. This inconsistency makes fail the CEL search.

The CEL bar allows us to filter by any field of the alert, but if the user wants to check what fields are available in the Payload, like:

Image

there will be a mismatch between the Alert able (the action zone of the CEL) and the Payload.

Image

Here, in the code, we can see that the alerts are retrieved by the query, i.e the CEL, and once, they will be enriched with the incidents. Therefore, if we need to filter by incidents we can't do it.

**To Reproduce**

1. Having an alert with incident.
2. Search by incident == null.

**Expected behavior**
The behavior is to get a real filter against incidents

diegovb-sys:
Closes #5293

## 📑 Description
Filtering by incident.id==null in the Feed, we may have some of alert with incidents, this happened if there was an alert with more than one incident, where one of them was different to Firing.

I've change the Incident.status == Firing OuterJoin by a subquery. This subquery allows us to make null the OuterJoin in case of there is an alert with one or more Incidents. Hence, the filter incident.id == / != null will work nice.

## ✅ Checks

### DIFF
--- a/keep/api/core/alerts.py
+++ b/keep/api/core/alerts.py
@@ -267,6 +267,21 @@ def __build_query_for_filtering(
         )
 
     if fetch_incidents or force_fetch:
+        # Fingerprint with active incidents subquery, i.e  in Firing status
+        firing_subq = (
+            select(LastAlert.fingerprint)
+            .join(
+                LastAlertToIncident,
+                LastAlert.fingerprint == LastAlertToIncident.fingerprint
+            )
+            .join(
+                Incident,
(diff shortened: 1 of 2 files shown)


shahargl (approved):
lgtm

PR #5338, opened 2025-09-29
fix: Incident autoresolved 

### LINKED ISSUE #5325: [🐛 Bug]: Resolve ON - Incidents
**Describe the bug**
The resolve on Last and First alert is relaying on the is_edge_incident_resolved(). 
There are 2 sub-issues.

On of them is about the compatibility with Postgres, due to the having with the max() function and the group_by.

The other is the logical behind it.

The **First Resolved** should evaluate the first alert received. Nevertheless, the idea here (in the code) is to select the oldest Alert, but if I've received an update to that fingerprint, the oldest and hence the First will be other...

The **Last Resolved** should evaluate either the last alert, in overall received, or the last alert by fingerprint. Nevertheless, here we have again a mismatch. 

To finalize, the test shows this mismatch. Focusing on test_incident_resolution_on_edge test, and the (ResolveOn.LAST.value, ("fp2", "fp1")) parameters:

1. FP1 Firing alert comes.
2. FP2 Firing alert comes.
3. FP2 Resolved alert comes.

**This asserts Firing, but the Last alert is FP2.. so the incident should be resolved.**

diegovb-sys:
Closes #5325

## 📑 Description
The current implementation doesn't seem to cover the idea behind the last resolved and first resolved. Moreover, based on a Postgres environment there is a Warning/Error because of the use of HAVING/GROUP_BY

## ✅ Checks

### DIFF
--- a/ee/identitymanager/identity_managers/auth0/auth0_authverifier.py
+++ b/ee/identitymanager/identity_managers/auth0/auth0_authverifier.py
@@ -1,16 +1,55 @@
+import logging
 import os
 
 import jwt
+import requests
 from fastapi import HTTPException
 
 from keep.identitymanager.authenticatedentity import AuthenticatedEntity
 from keep.identitymanager.authverifierbase import AuthVerifierBase
 from keep.identitymanager.rbac import Admin as AdminRole
 
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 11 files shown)


shahargl (approved):
lgtm

diegovb-sys:
Hi @shahargl could you merge this requests? Thx

PR #5347, opened 2025-10-02
fix: Catch errors about comparison CEL

### LINKED ISSUE #5340: [🐛 Bug]: Enrichments vs CEL comparison
**Describe the bug**
Mapping or another enrichment feature allows us to modify some parameters regarding the original alert. Nevertheless, when we are handling different types to string, this could cause problems.In the next section, I show an example.

**To Reproduce**

1. Creating a Mapping to enrich with dismissed==true
2. Creating a WF/Incident with the CEL -> cel: 'dismissed==true'
3. Check the logs

**Expected behavior**
The expected behavior is to have a WF(Incident) CEL evaluation.

**Screenshots**

Image

diegovb-sys:
Closes #5340

## 📑 Description
As I've posted in the Issue. There are some scenarios, where specific alert fields can be casted to other type than the expected. 
An example is to use Mapping to dismiss an alert and after of that, read that field in a WF. The WF would expect a boolean value, but it will received a string, so this will cause a celpy error.

In order to face that problem, I've implemented (inside the Except) in WF and Incidents scopes a cast to the original values. (Using the ones defined in AlertDTO)

## ✅ Checks


shahargl:
@diegovb-sys can you fix the failing UT?

PR #5364, opened 2025-10-14
fix: Avoid race condition in set_last_alert

### LINKED ISSUE #5354: [🐛 Bug]: Race condition in set_last_alert
**Describe the bug**
`set_last_alert` can lead to a race condition if an alert is received by "duplicate" in the same moment.

**To Reproduce**
In order to reproduce the follow conditions should be occurred:

- The alert in the Alert table can't be present.
- Hence, in the lastalert table there is not any entry related to that.
- 2 alerts are send in the same moment with some distinct field to avoid the deduplication.
- More than one worker should be up.

As 2 workers are handling the same alert, this will make a race condition, the faster alert will INSERT itself in lastalert and at the same time, the other one will try to INSERT itself but there will be already a lastAlert.

**Expected behavior**
The expected behavior is to have the real alert in any moment.

**Screenshots**
Table Alert with same fingerprint:

Image

Add to DB error:

Image

Image

**Impact**
There are some scenarios with more impact than others, but in the worse case, we can receive a Firing alert just after the Resolve alert, in that case all of the flows that retrieves the lastAlert would show false information.

Moreover, the processes which happen after that call, won't be applied:

- Mapping.
- Resolve incident if it applies.
- Enrichments.
- Pusher notifications.
- WF executions.

diegovb-sys:
Closes #5354

## 📑 Description
In order to face the race condition. it has been modified the checking strategy  after if the value is present. 
With the current strategy, a query is built where we include the action in case of conflict, i.e when the lastalert already exists.
Moreover, as this strategy depends on the underlying engine, it has been implemented for 3 engines:
- SQLite
- PostreSQL
- MySQL

About the fails catching, it has been moved to the decorator.

With this modification race conditions should be avoided, when the same alert (tenant+fingerprint) comes in twice.

## ✅ Checks

## ℹ Additional Information
Since the function modified is so called in the principal flow, be careful in its review.

### DIFF
--- a/keep/api/core/db_utils.py
+++ b/keep/api/core/db_utils.py
@@ -225,8 +225,11 @@ def insert_update_conflict(table: SQLModel, session: Session, data_to_insert: di
         query = mysql_insert(table).values(data_to_insert)
         if update_newer:
             data_to_update = {
-                k: case((table.timestamp < query.inserted.timestamp, v), else_=getattr(table, k))
-                for k, v in data_to_update.items()
+                k: case(
+                    (table.timestamp < query.inserted.timestamp, query.inserted[k]),
+                    else_=getattr(table, k)
+                )
+                for k in data_to_update.keys()
             }
         query = query.on_duplicate_key_update(data_to_update)
     elif session.bind.dialect.name == "sqlite":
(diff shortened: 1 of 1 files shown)


shahargl:
@diegovb-sys it looks like linter fails, wdyt can you fix it?

diegovb-sys:
Hi @shahargl sorry for the delay, the linter is fixed. Nevertheless, the actions continue showing some fails, these fails are not related with this PR.

diegovb-sys:
Hello @shahargl could you review and merge this pls?

shahargl:
@diegovb-sys we have some problem with the CICD I need to fix before that :(

PR #5392, opened 2025-10-24
docs: Fix capitalization in GitHub Workflows documentation

gecube:
fix docs.


talboren:
This gets overridden by an automated script.

gecube:
@talboren hey! Thanks for the explanation. How could we then fix it? It should not sound too complicated, but I am a little bit lost where to find a proper place.

talboren:
use `scripts/docs_get_providers_list.py` to generate recent providers list and update the file.

PR #5419, opened 2025-10-28
fix: DBSecretManager raises an exception

### LINKED ISSUE #5417: [🐛 Bug]: Secrets in db
**Describe the bug**
The method read_secret in DbSecretManager has a return type, in its signature, of str | dict. However, in case of the secret doesn't exist this method won't return anything... Causing an error.

As far as I could seen, the properly way to handle when the secret doesn't exist is to generate a exception.

**To Reproduce**

1. Use the SECRET_MANAGER_TYPE=DB env var in the backend.
2. Go to WF page and search for secrets in the inspector.

**Screenshots**
Inspector:
Image

diegovb-sys:
Closes #5417

## 📑 Description
In the same way as other SecretManager(see filesecretmanager), I've implemented a exception in case of the value doesn't exist in the DB, the callers will caught this and return an empty dict. 

Right now, as I commented in the issue, in case of empty value in db, nothing value was returned, causing this an error.

## ✅ Checks

## ℹ Additional Information
Call with empty value:

imagen

### DIFF
--- a/keep/secretmanager/dbsecretmanager.py
+++ b/keep/secretmanager/dbsecretmanager.py
@@ -26,6 +26,8 @@ class DbSecretManager(BaseSecretManager):
                     if is_json:
                         return json.loads(secret_model.value)
                     return secret_model.value
+                else:
+                    raise KeyError(f"Secret {secret_name} not found")
             except Exception as e:    
                 self.logger.error(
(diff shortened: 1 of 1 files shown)


talboren (approved):
LGTM

Issue #2112, opened 2024-10-08
[🔌 Provider]: SNMP provider


shahargl:
Send SNMP traps/events into Keep as alerts

Matvey-Kuk:
Labeling this one as a Good First Issue.

Use https://docs.keephq.dev/providers/adding-a-new-provider as a starting point :)

If you want to work on this, join Slack and we will help: https://slack.keephq.dev/ !

michael-azogu:
/attempt #2112

im looking into this

shahargl:
Hey @onyedikachi-david, do you have any idea how to implement it? Any experience with SNMP?

michael-azogu:
@shahargl from my research...
its basically like Prometheus for network devices.
traps (like os interrupts) are pushed when certain events occur with a context payload.
I need to figure out how to parse & whether ill need pysnmp client

onyedikachi-david:
@shahargl Not in SNMP in particular, but have a good experience with opentelemetry/observability, I am currently extending an instrumentation by adding event-based tracking implementation across providers for a Open-source observability platform for LLM applications (openllmetry).

shahargl:
Ok let’s try! Assigned

Matvey-Kuk:
This issue is up for grabs :)

c0d33ngr:
/attempt #2112

Let me see what I can do here

Omar8345:
@shahargl Does Algora have PayPal payouts? If not, is there manual PayPal payout from keep directly?

rahulrajesh21:
@shahargl  Can you review my pr.

advenk:
/attempt #2112

i'd like to give this a shot if its still open.

Harshvardhan-91:
@shahargl can i work on this?

Kesavaraja67:
Hi @shahargl, seeing that the previous attempts from September seem inactive, I would like to pick this up.

Plan: I will implement the SnmpProvider using pysnmp to handle trap reception and parsing, converting them into Keep alerts.

/attempt #2112

matiasmagni:
/attempt #2112

### Implementation Plan:

1. Analyze existing optimization functions and models in the Python backend to understand current metric weighting and identify potential extension points.
2. Review existing data models (e.g., Plan or configuration) to determine where new optimization preferences (quality, speed, resource) could be stored.
3. Modify or create new Python backend functions to accept an optimization_target parameter.
[...]

PR #5525, opened 2026-01-27
feat: Add SNMP provider for receiving SNMP traps as alerts

SBALAVIGNESH123:
Adds a new SNMP provider that receives SNMP traps/events as alerts.
Closes #2112


shahargl:
N/A

PR #5578, opened 2026-02-07
feat: allow custom JSON payload for Grafana OnCall

### LINKED ISSUE #5316: [➕ Feature]: Allow custom JSON input for Grafana OnCall Provider
With the current version of the Grafana OnCall Provider, there is a limited set of information we can send with our alerts. Grafana supports custom templating and accepts key/value pair in the body of requests sent to their OnCall API yet we're currently limited to only a select few parameters.

Simply adding a new parameter (such as `body`, `payload` or `json`) should be enough. If the parameter is passed to the provider, we send the value instead of creating a json object based on other parameters ourselves.

PoornaChandra2005:
Fixes #5316

This PR adds a new payload parameter to the Grafana OnCall notify action. Users can now send custom JSON data directly to the OnCall API while maintaining backward compatibility with existing fields.

### DIFF
--- a/keep/providers/grafana_oncall_provider/grafana_oncall_provider.py
+++ b/keep/providers/grafana_oncall_provider/grafana_oncall_provider.py
@@ -1,9 +1,10 @@
-"""
-Grafana Provider is a class that allows to ingest/digest data from Grafana.
-"""
+\"\"\"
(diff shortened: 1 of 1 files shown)


shahargl:
N/A

PR #5605, opened 2026-02-12
fix: make priority field optional in Jira OnPrem provider

### LINKED ISSUE #5534: [🐛 Bug]: Create issue on jira if priority is not present
**Describe the bug**

```
```

**To Reproduce**

**Expected behavior**

zekebawt:
The Jira OnPrem provider was enforcing a default Medium priority field even when the target Jira instance doesn't have priority configured in its screen configuration.

Changes:
- Makes priority parameter optional (default None)  
- Only includes priority field in API request when explicitly provided
- Prevents Field priority cannot be set errors on Jira instances without priority field configured

Updated all three methods: __create_issue, __update_issue, and _notify.

Fixes #5534

### DIFF
--- a/keep/providers/jiraonprem_provider/jiraonprem_provider.py
+++ b/keep/providers/jiraonprem_provider/jiraonprem_provider.py
@@ -284,7 +284,7 @@ class JiraonpremProvider(BaseProvider):
         labels: List[str] = None,
         components: List[str] = None,
         custom_fields: dict = None,
-        priority: str = "Medium",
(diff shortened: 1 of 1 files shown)


shahargl:
?

You committed also unrelated code?

zekebawt:
@shahargl You're absolutely right, I apologize. I committed unrelated changes from a different branch. That was sloppy on my part. I've closed this PR and will be more careful about keeping changes isolated in the future.

PR #5607, opened 2026-02-12
fix: Handle optional priority field in Jira issue creation

### LINKED ISSUE #5534: [🐛 Bug]: Create issue on jira if priority is not present

zekebawt:
## Description
Fixes issue where Jira provider fails when creating issues on projects where the priority field is not available on the issue screen.

## Problem
When the priority field is not configured on a Jira project's issue screen, sending  (or empty/null values) in the issue creation request causes the error:

## Solution
- Filter out priority field when value is 'none', empty string, or null
- Handle priority field in both  and  parameters
- Log informative message when skipping the field
- Preserve all other custom fields

## Changes
- Enhanced  method to filter invalid priority values
- Added comprehensive test suite with 5 test scenarios
- Tests cover various priority value formats and edge cases

## Test Coverage
✅ Priority with value "none" → excluded from request
✅ Priority with empty string → excluded from request  
✅ Priority with null value → excluded from request
✅ Valid priority values → included in request
✅ Other custom fields → preserved when priority is filtered

## Backward Compatibility
This change is fully backward compatible. Valid priority values continue to work as expected, while invalid values are gracefully handled instead of causing errors.

Fixes #5534

### DIFF
--- a/keep/providers/jira_provider/jira_provider.py
+++ b/keep/providers/jira_provider/jira_provider.py
@@ -408,7 +408,24 @@ class JiraProvider(BaseProvider):
                 fields["components"] = [{"name": component} for component in components]
 
             if custom_fields:
-                fields.update(custom_fields)
+                # Filter out priority field if it's set to "none" or empty
+                filtered_fields = {}
+                for key, value in custom_fields.items():
+                    if key == "priority" and (not value or str(value).lower() in ["none", "", "null"]):
+                        self.logger.info(f"Skipping priority field with value '{value}' as it may not be available on the issue screen")
+                        continue
+                    filtered_fields[key] = value
+                fields.update(filtered_fields)
+            
+            # Also handle priority that might come through kwargs
+            if kwargs:
+                filtered_kwargs = {}
+                for key, value in kwargs.items():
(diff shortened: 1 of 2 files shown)


Eliesmbr:
Sorry to open this again but this PR introduced a regression that breaks Jira issue creation for workflows that use enrich_alert.
The newly added kwargs block (lines 419-425) passes all kwargs into the Jira API fields, including Keep-internal fields like enrich_alert. Jira rejects these with:   
`{'errorMessages': [], 'errors': {'enrich_alert': "Field 'enrich_alert' cannot be set. It is not on the appropriate screen, or unknown."}} `

Before this PR, kwargs were never added to fields. The custom_fields filtering for priority is fine, but the kwargs block should be removed  or at minimum, Keep-internal fields like enrich_alert need to be excluded.
This is currently breaking all Jira ticket creation in main branch. cc @zekebawt cc @shahargl

The priority filtering in custom_fields (the upper block) is fine that was the actual intention of the PR. Only the kwargs block is the problem, because kwargs in _notify() contains all sorts of things (enrich_alert, dispose_on_new_alert, etc.) that should never be sent to Jira.

Issue #5480, opened 2025-12-09
[🐛 Bug]: postgresql connection tune


Viste:
**Describe the bug**
Now you can't tune connection from app side

**To Reproduce**
run app with postgress tuned idle connection live settings and get:
```
{"worker_type": "arqworker", "asctime": "2025-12-09 14:09:45,850", "message": "Exception during reset or similar", "levelname": "ERROR", "name": "sqlalchemy.pool.impl.QueuePool", "filename": "base.py", "otelTraceID": "4bc123179dd4dc92086a4d9333467cd6", "otelSpanID": "66e372f202b7142c", "otelTraceSampled": true, "otelServiceName": "keep-api", "threadName": "process_event_worker_1", "process": 34, "module": "base", "exc_info": "Traceback (most recent call last):\n  File \"/venv/lib/python3.13/site-packages/sqlalchemy/pool/base.py\", line 986, in _finalize_fairy\n    fairy._reset(\n    ~~~~~~~~~~~~^\n        pool,\n        ^^^^^\n    ...<2 lines>...\n        asyncio_safe=can_manipulate_connection,\n        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n    )\n    ^\n  File \"/venv/lib/python3.13/site-packages/sqlalchemy/pool/base.py\", line 1432, in _reset\n    pool._dialect.do_rollback(self)\n    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^\n  File \"/venv/lib/python3.13/site-packages/sqlalchemy/engine/default.py\", line 699, in do_rollback\n    dbapi_connection.rollback()\n    ~~~~~~~~~~~~~~~~~~~~~~~~~^^\npsycopg2.OperationalError: server closed the connection unexpectedly\n\tThis probably means the server terminated abnormally\n\tbefore or while processing the request.\n", "taskName": null}
```

**Expected behavior**
you can tune connection settings
**Screenshots**

**Additional context**

PR #5474, opened 2025-12-05
feat: allow to config pool_recycle and pool_timeout

Viste:
close #5480 
## 📑 Description
allow to config pool_recycle and pool_timeout for DBs

PR #5611, opened 2026-02-13
fix: close leaked SQLAlchemy sessions in process_watcher_task

### LINKED ISSUE #5496: [🐛 Bug]: Critical database connection issues under high alert load
### Description

Keep backend has multiple critical database connection issues that make it unusable in high-throughput production environments (1000+ alerts/minute). These issues occur both with direct PostgreSQL connections AND through PgBouncer.

### Environment

- Keep version: 0.48.0
- PostgreSQL: 15.x (direct connection, VM)
- Python: 3.13
- psycopg2: latest
- SQLAlchemy: 2.0.36
- Deployment: 1 pods, 4 gunicorn workers + 4 ARQ workers each

### Configuration tested
```yaml
DATABASE_POOL_SIZE: "20"
DATABASE_MAX_OVERFLOW: "40"
KEEP_DB_PRE_PING_ENABLED: "true"
```

---

### Symptoms

Under moderate load (~50 alerts/second), the connection pool becomes exhausted within minutes:
```
sqlalchemy.exc.TimeoutError: QueuePool limit of size 20 overflow 40 reached, 
connection timed out, timeout 30.00
```

### Analysis

Connections are acquired but never returned to the pool. Even with `pool_pre_ping=True`, dead connections accumulate. PostgreSQL shows connections in `idle` state that are never reused by the application.

**Expected**: 60 connections (20 + 40 overflow) should be recycled
**Actual**: All 60 connections become "stuck", new requests timeout

---

### Error
```python
KeyError: "Deferred loader for attribute 'id' failed to populate correctly"

File "keep/api/core/db.py", line 2716, in create_deduplication_event
    "deduplication_event_id": deduplication_event.id,
```

### Root cause

In `db.py:2716`, code accesses `deduplication_event.id` immediately after `session.add()` but before `session.flush()` or `session.commit()`. The ORM hasn't populated the auto-generated ID yet.

### Fix required
```python
# Current (broken):
session.add(deduplication_event)
return {"deduplication_event_id": deduplication_event.id}  # ID not populated!

# Should be:
session.add(deduplication_event)
session.flush()  # Force INSERT to get ID
return {"deduplication_event_id": deduplication_event.id}
```

---

## Issue 3: PgBouncer incompatibility

When using PgBouncer (even in session mode), psycopg2 produces protocol-level errors:
```
psycopg2.DatabaseError: error with status PGRES_TUPLES_OK and no message from the libpq
sqlalchemy.exc.NoSuchColumnError: Could not locate column in row for column 'workflow.id'
IndexError: tuple index out of range
```

### PgBouncer config tested
```ini
pool_mode = session
server_reset_query = DISCARD ALL
ignore_startup_parameters = extra_float_digits
```

**Result**: Still fails. psycopg2 binary protocol has fundamental incompatibility with PgBouncer.

---

## Issue 4: Stale connections not detected
```
psycopg2.OperationalError: server closed the connection unexpectedly
This probably means the server terminated abnormally before or while processing the request.
```

PostgreSQL closes idle connections (via `idle_session_timeout`), but SQLAlchemy pool doesn't detect this despite `pool_pre_ping=True`.

---

## Summary table

| Setup | Result | Primary error |
|-------|--------|---------------|
| Direct PostgreSQL | ❌ Fails | Pool exhaustion + deferred loader |
| Direct + pre_ping | ❌ Fails | Same issues |
| PgBouncer session mode | ❌ Fails | Protocol corruption |
| PgBouncer + DISCARD ALL | ❌ Fails | Same protocol errors |

This blocks enterprise adoption where connection pooling is mandatory.

Rolled:
## Summary

- Fix SQLAlchemy session leak in `process_watcher_task` cron job that causes connection pool exhaustion in the arqworker
- `MaintenanceWindowsBl.recover_strategy()` and `DismissalExpiryBl.check_dismissal_expiry()` create sessions via `get_session_sync()` when called without a session argument (the `process_watcher_task` code path), but never close them
- Every 60s cron cycle leaks 2 connections stuck in PostgreSQL "idle in transaction" state, exhausting the default pool (size=5 + overflow=10) within ~30 minutes
- This breaks all workflow execution (WorkflowScheduler thread shares the same pool)

## Root Cause

`get_session_sync()` returns a bare `Session(engine)` with no context manager. When `recover_strategy()` and `check_dismissal_expiry()` are called from `process_watcher_task.py` via `run_in_executor`, they receive only `logger` — no session. Both methods create internal sessions but never call `session.close()`, so the connection is never returned to the pool.

## Changes

| File | Change |
|------|--------|
| `keep/api/bl/maintenance_windows_bl.py` | Wrap `recover_strategy` body in `try/finally`; close session in `finally` when internally created |
| `keep/api/bl/dismissal_expiry_bl.py` | Add `session.close()` to existing `finally` block when session is internally created |
| `tests/test_session_cleanup.py` | 6 regression tests verifying session cleanup (internal creation, caller-provided, exception paths) |

Both fixes use `_owns_session = session is None` to track ownership — only sessions created internally are closed. Caller-provided sessions are left to the caller, preserving existing behavior for all other call sites.

## Test plan

- [ ] Existing unit tests pass (recover_strategy tests all provide sessions explicitly — unaffected by this change)
- [ ] New `test_session_cleanup.py` tests verify:
  - `session.close()` called when session created internally (`session=None`)
  - `session.close()` NOT called when caller provides a session
  - `session.close()` called even when exception occurs mid-execution
- [ ] Manual verification: deploy and confirm `pg_stat_activity` no longer accumulates "idle in transaction" connections from arqworker process

## Related issues

- Fixes #5496 (pool exhaustion under load — same root cause)
- Related to #5480 (stale connection errors on `process_event_worker_1`)
- Supersedes PR #5474 (pool_recycle/pool_timeout config — treated symptom, not cause)


AlejandroGarcia95 (approved):
We have tested this on our lab environment and we can confirm this has solved the issue for us.

@shahargl may you take a look?

PR #5630, opened 2026-02-15
fix: filter Keep-internal fields from Jira kwargs

### LINKED ISSUE #5534: [🐛 Bug]: Create issue on jira if priority is not present

zekebawt:
Fixes #5534

Fixes regression introduced in #5607 where Keep-internal workflow fields were being passed to Jira API, causing errors.

## Problem
The kwargs block added in #5607 filtered priority values but passed all remaining kwargs to Jira fields. This included Keep-internal fields that Jira rejects with 'cannot be set' errors.

Reported by @Eliesmbr in #5607 comments.

## Solution
Added explicit filtering for Keep-internal fields:
- enrich_alert, enrich_incident
- dispose_on_new_alert
- audit_enabled
- if, name, condition, foreach, throttle, provider

These are now skipped while legitimate Jira fields continue to work.

## Testing
- ✅ Priority filtering (original #5607 feature) still works
- ✅ Custom fields via custom_fields parameter preserved
- ✅ Keep-internal fields excluded from Jira API calls
- ✅ Syntax validation passes

## Regression Fixed
Resolves: 'Field '\''enrich_alert'\'' cannot be set. It is not on the appropriate screen, or unknown.'


shahargl:
@zekebawt would you mind fix the "test docs" step?

Eliesmbr:
I think also some unit tests are failing because of the new change. Would you mind fixing these @zekebawt ? Currently all new PR pipelines will fail.

Eliesmbr:
The 5 failing tests in `tests/providers/jira_provider/test_jira_priority_fix.py` are not caused by your code changes in this PR. They were introduced in #5607 but never ran in CI because GitHub requires workflow approval for first-time fork contributors the PR was merged before tests were triggered.

The tests use `@responses.activate` to mock HTTP calls. Unlike `@patch("requests.post")` (used by the existing Jira tests in `tests/test_jira_provider.py`), `@responses.activate` lets the real `requests.post()` → `Session.request()` code path execute and only intercepts at the HTTP adapter level.

This hits a monkey-patch in `keep/api/api.py:96-105` that replaces `Session.request` with `no_redirect_request`. When `api.py` is loaded more than once in the test process, `original_request` ends up pointing to `no_redirect_request` itself instead of the real method  causing infinite recursion:

```
keep/api/api.py:102: in no_redirect_request
    return original_request(self, method, url, **kwargs)
keep/api/api.py:102: in no_redirect_request
    return original_request(self, method, url, **kwargs)
...
RecursionError: maximum recursion depth exceeded in comparison
```

The tests should use `@patch("requests.post")` instead of `@responses.activate`, consistent with the existing Jira provider tests in `tests/test_jira_provider.py`.

Unfortunately i can not commit to `zekebawt:fix/jira-kwargs-regression`.

PR #5641, opened 2026-02-17
fix(uptimekuma): fix critical bugs in pull mode and webhook handling

### LINKED ISSUE #5655: [🐛 Bug]: UptimeKuma provider has multiple critical bugs in pull mode and webhook handling
## Description

## Expected Behavior

## Steps to Reproduce

## Fix

ART-R2D275:
## Summary
Fix critical bugs in UptimeKuma provider for both pull mode and webhook handling.

Fixes #5655

### Bug Fixes
- **Fix heartbeats.append() outside for-loop** — Only last monitor was being reported instead of all monitors
- **Fix _format_datetime string concatenation** — Was causing TypeError
- **Fix webhook _format_alert referencing non-existent fields** — Was causing KeyError on every webhook

### Improvements
- Add AlertSeverity mapping from Uptime Kuma status codes
- Add FINGERPRINT_FIELDS for proper alert deduplication
- Add try/finally for connection cleanup
- Add bulk monitor fetching for better performance
- Add STATUS_MAP with both numeric and string key support

## Testing
- Tested pull mode with multiple monitors
- Tested webhook payload handling
- Verified connection cleanup

### DIFF
--- a/keep/providers/uptimekuma_provider/uptimekuma_provider.py
+++ b/keep/providers/uptimekuma_provider/uptimekuma_provider.py
@@ -1,38 +1,55 @@
 """
-UptimeKuma is a class that provides the necessary methods to interact with the UptimeKuma SDK
+Uptime Kuma provider for Keep.
+
+Uptime Kuma is a self-hosted monitoring tool that supports HTTP(s), TCP, Ping,
+DNS and other monitor types. This provider supports both pulling alerts via the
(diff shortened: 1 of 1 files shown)


ART-R2D275:
Hi maintainers 👋

The CLA assistant bot is requesting a signature, but when I try to sign at https://cla-assistant.io/keephq/keep it returns:

> "There is no CLA to sign for keephq/keep (Not Found)"

Could you help resolve this? Happy to sign if there's a working link.

Thanks!

PR #5651, opened 2026-02-19
fix: trigger workflows on auto-resolve of incidents

### LINKED ISSUE #5652: Auto-resolved incidents do not trigger workflow execution
## Description
Incidents that are auto-resolved (e.g., via `resolve_on=ResolveOn.ALL.value`) do not trigger the expected workflow event (`updated` action).

## Steps to Reproduce
1. Create an incident with `resolve_on=ResolveOn.ALL.value`.
2. Add resolved alerts to the incident until the condition is met.
3. Observe that the incident status changes to `RESOLVED`, but no workflow is triggered.

## Expected Behavior
Auto-resolving an incident should trigger a workflow event, similar to manual resolution.

## Actual Behavior
The workflow event is skipped.

## Fix
This issue is addressed by PR #5651.

BrianInAz:
Fixes a bug where auto-resolved incidents did not trigger workflow execution. Added regression test.

Fixes #5652.

### DIFF
--- a/docs/deployment/configuration.mdx
+++ b/docs/deployment/configuration.mdx
@@ -101,6 +101,43 @@ Keep is highly configurable through environment variables. This allows you to cu
 | **KEEP_FORCE_RESET_DEFAULT_PASSWORD** |               Forces reset of default user password               |    No    |    "false"    |                 "true" or "false"                  |
 |       **KEEP_DEFAULT_API_KEYS**       |       Comma-separated list of default API keys to provision       |    No    |      ""       |    Format: "name:role:secret,name:role:secret"     |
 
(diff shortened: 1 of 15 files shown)


BrianInAz:
Thanks for the feedback! I've refactored  to accept a  flag (default True).  now passes  to prevent duplicate events, while  relies on the default behavior to fix the original bug. Added a regression test case for the suppression logic.

BrianInAz:
Thanks for the feedback! I've refactored `resolve_incident_if_require` to accept a `handle_workflow_event` flag (default True). `RulesEngine` now passes `False` to prevent duplicate events, while `process_event_task` relies on the default behavior to fix the original bug. Added a regression test case for the suppression logic.

PR #5666, opened 2026-02-23
fix(ui): prevent false 'Alert fingerprint not found' on alerts re-fetch

### LINKED ISSUE #5667: bug: 'Alert fingerprint not found' toast fires on background alerts re-fetch, closing modal unexpectedly
## Describe the bug

When opening an alert detail via URL (`?alertPayloadFingerprint=...`), the modal opens correctly. However, because the alerts list is continuously re-fetched (polling / WebSocket updates), the `useEffect` in `alerts.tsx` re-runs with `[searchParams, alerts]` as dependencies on every re-fetch.

If the alert is no longer present in the current page of results after a re-fetch (e.g. a new alert pushed it off page 1), the fingerprint lookup returns `undefined` and the error branch fires — showing a false **"Alert fingerprint not found"** toast and closing the modal unexpectedly, even though the alert exists and was already displayed.

## Steps to reproduce

1. Open the alerts list with enough alerts to fill multiple pages
2. Navigate to an alert detail via URL: `?alertPayloadFingerprint=<fingerprint>`
3. Modal opens correctly
4. Wait for a background re-fetch (polling interval or WebSocket push)
5. If the alert moves off page 1 during the re-fetch, the modal closes with an error toast

## Expected behaviour

The modal stays open. Background re-fetches should not close a modal that was already successfully resolved.

## Actual behaviour

Modal closes with a false "Alert fingerprint not found" error toast.

## Environment

- Keep version: 0.49.x (reproduced on main)
- Browser: any

## Relation to #4533

PR #4533 added `&& alerts` guards to prevent the error when alerts haven't loaded yet. This issue is the separate case where alerts *have* loaded but the alert is not on the current page during a *subsequent* re-fetch.

hakatt:
Fixes #5667

## Problem

When a user opens an alert detail via URL (`?alertPayloadFingerprint=...`), the modal opens correctly. However, the alerts list is continuously re-fetched (polling / WebSocket updates). On each re-fetch, the `useEffect` that resolves the fingerprint re-runs with `[searchParams, alerts]` as dependencies.

If the alert is pushed off the current page (e.g. it's no longer in the first page of results after a re-fetch), the `alert` lookup returns `undefined` and the error branch fires — showing a false "Alert fingerprint not found" toast and closing the modal unexpectedly.

## Fix

Track the resolved fingerprint in a `useRef` (`resolvedFingerprintRef`). Once the modal has been successfully opened for a given fingerprint, skip the error branch on subsequent re-fetches. The ref is reset when the user navigates to a different fingerprint.

```tsx
const resolvedFingerprintRef = useRef<string | null>(null);

// Reset when navigating to a different fingerprint
if (fingerprint !== resolvedFingerprintRef.current) {
  resolvedFingerprintRef.current = null;
}

if (alert) {
  resolvedFingerprintRef.current = fingerprint;
  setViewAlertModal(alert);
} else if (!resolvedFingerprintRef.current) {
  // Only fire the error on the first attempt, not on background re-fetches
  showErrorToast(null, "Alert fingerprint not found");
  resetUrlAfterModal();
}
```

## Relation to #4533

PR #4533 fixed the case where alerts hadn't loaded yet (`&& alerts` guard). This PR fixes the separate case where alerts *have* loaded but the specific alert is not on the current page during a subsequent re-fetch.

## Testing

- Open an alert detail via URL with `?alertPayloadFingerprint=...`
- Verify modal opens without error toast
- Wait for a background re-fetch (or trigger one) — modal should stay open
- Navigate to a different fingerprint — ref resets, new alert resolves correctly
- Navigate to a non-existent fingerprint — error toast still fires as expected

### DIFF
--- a/keep-ui/app/(keep)/alerts/[id]/ui/alerts.tsx
+++ b/keep-ui/app/(keep)/alerts/[id]/ui/alerts.tsx
@@ -96,27 +96,41 @@ export default function Alerts({ presetName, initialFacets }: AlertsProps) {
     facetsPanelRefreshToken,
   } = useAlertsTableData(alertsTableDataQuery);
 
+  // Track which fingerprint has already been resolved so that a background
+  // alerts re-fetch (polling / WebSocket) doesn't fire "not found" after the
+  // modal was successfully opened.
+  const resolvedFingerprintRef = useRef<string | null>(null);
+
   useEffect(() => {
     const fingerprint = searchParams?.get("alertPayloadFingerprint");
     const enrich = searchParams?.get("enrich");
+
+    // Reset when the user navigates to a different fingerprint.
+    if (fingerprint !== resolvedFingerprintRef.current) {
+      resolvedFingerprintRef.current = null;
+    }
+
     if (fingerprint && enrich && alerts) {
       const alert = alerts?.find((alert) => alert.fingerprint === fingerprint);
       if (alert) {
+        resolvedFingerprintRef.current = fingerprint;
         setEnrichAlertModal(alert);
         setIsEnrichSidebarOpen(true);
-      } else {
+      } else if (!resolvedFingerprintRef.current) {
         showErrorToast(null, "Alert fingerprint not found");
         resetUrlAfterModal();
       }
     } else if (fingerprint && alerts) {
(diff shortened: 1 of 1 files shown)


hakatt:
recheck

PR #5672, opened 2026-02-23
fix(ui): guard fingerprint modal against stale-empty alerts flash

### LINKED ISSUE #5667: bug: 'Alert fingerprint not found' toast fires on background alerts re-fetch, closing modal unexpectedly

hakatt:
## Root cause (deeper than the previous fix)

The `!alertsLoading` guard added earlier wasn't sufficient. The real issue is a 3-render cascade inside `useLastAlerts`:

1. SWR fetch completes → `isLoading = false`, `data = {queryResult: {...}}`
2. **Render R1**: `useLastAlerts` returns `data: []` (the `results` useState still holds the old empty value — the effect that hasn't run yet), `isLoading: false`
3. `useAlertsTableData` effect fires → `setAlertsToReturn([])` (stale empty)
4. `useLastAlerts` effect fires → `setResults([actual alerts])`
5. **Render R2**: `alertsToReturn = []`, `alertsLoading = false` → `alerts.find(fingerprint)` on empty array → **false "not found" error fires**
6. **Render R3**: real data arrives — too late, URL already cleared

## Fix

Introduce a `dataSettled` guard that only acts once data is confirmed real:

```tsx
const dataSettled =
  alerts && !alertsLoading && (alerts.length > 0 || totalCount === 0);
```

- `alerts.length > 0` — actual results have arrived
- `totalCount === 0` — backend confirmed empty result set (genuine not found)
- `totalCount > 0` with `alerts = []` — transitional stale state, wait

`totalCount` comes directly from SWR data (not React state), so it already reflects the real count in the same render where `alerts` is still the stale `[]`.

## Test

Added a regression test covering the stale-empty flash scenario: mocks `useAlertsTableData` returning `alerts=[], totalCount=5` (stale phase), asserts no error fires, then updates to `alerts=[alert], totalCount=1` (real data), asserts modal opens without error.

Fixes #5667

### DIFF
--- a/keep-ui/package-lock.json
+++ b/keep-ui/package-lock.json
@@ -339,7 +339,6 @@
       "version": "7.26.0",
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.26.0.tgz",
       "integrity": "sha512-i1SLeK+DzNnQ3LL/CswPCa/E5u4lh1k6IAEphON8F+cXt0t9euTshDru0q7/IqMa1PMPz5RnHuHscF8/ZJsStg==",
-      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.26.0",
@@ -2062,6 +2061,7 @@
       "version": "2.3.0",
       "resolved": "https://registry.npmjs.org/@browserbasehq/sdk/-/sdk-2.3.0.tgz",
       "integrity": "sha512-H2nu46C6ydWgHY+7yqaP8qpfRJMJFVGxVIgsuHe1cx9HkfJHqzkuIqaK/k8mU4ZeavQgV5ZrJa0UX6MDGYiT4w==",
+      "peer": true,
       "dependencies": {
         "@types/node": "^18.11.18",
         "@types/node-fetch": "^2.6.4",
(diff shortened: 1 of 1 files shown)


shahargl:
@hakatt there is merge conflict - can you resolve it?

hakatt:
Blocked by #5676 — the `frontend-tests` check is failing on all PRs due to an out-of-sync `package-lock.json` (missing `@types/react@18.3.28`). Once that merges, CI should go green here too.

shahargl:
i've merged https://github.com/keephq/keep/pull/5676 let's see what CICD says

PR #5684, opened 2026-02-25
feat(workflow): Mustache lambda helpers for missing-key defaults (fn.*)

### LINKED ISSUE #5682: [Proposal] Workflow template engine: Mustache lambda helpers for missing-key defaults
# Workflow template engine: missing-key handling for custom/enriched fields

## Summary

Workflow templates that reference custom or enriched alert fields (fields not guaranteed by the upstream provider schema) currently crash with `RenderException` because `render_context()` always renders with `safe=True`. There is no way for a workflow author to express a default value for a field that may or may not be present.

## Background: two distinct categories of missing fields

Alert fields in Keep come from two different sources, and they need to be handled differently:

### 1. Provider schema fields

Fields that are part of the upstream service's alert schema — e.g. Grafana's `panelUrl`, `dashboardUrl`, `valueString`; Prometheus's `instance`, `job`, `value`. These fields:

- Have a known, documented shape from the upstream provider
- Should always be present if the provider's `_format_alert()` is correct
- When absent, it indicates a gap in the provider implementation or a schema change upstream

**Correct handling: the provider is responsible.** Each provider should explicitly default any field that may be absent in `_format_alert()`, and the absence should be logged so that provider drift from upstream is visible and can be fixed. This is already the pattern in the codebase and is the right place to enforce schema contracts.

### 2. Custom and enriched fields

Fields that a workflow author adds via `enrich_alert`, custom labels, calculated values, or integration-specific enrichments. Examples: `slack_timestamp` (stored after first Slack post), tenant-specific label mappings, fields from a secondary lookup step. These fields:

- Are not part of any provider schema
- May or may not be present depending on alert history and workflow state
- Are inherently optional from the workflow author's perspective

**No provider can default these** — they are outside the scope of any upstream service. The workflow template itself is the only appropriate place to handle their optionality.

## The problem

`render_context()` in `iohandler.py` hardcodes `safe=True` for all string `with:` parameters. When Chevron encounters a missing key under `safe=True`, it warns to stderr and `_render()` converts that warning into a `RenderException`, aborting the action.

There is currently no supported syntax in workflow YAML for a workflow author to say: "render this field, but if it's missing use a default value". The `keep.*` function system cannot help here because it is evaluated *after* Mustache rendering — by the time `keep.default(...)` would run, the `RenderException` has already been thrown.

## Proposed solution

A two-part approach that keeps the two categories cleanly separated:

### Part A — Provider-level defaulting (already the right pattern, no change needed)

Providers continue to own and default their schema fields in `_format_alert()`. If a provider emits an alert without a field that is part of its upstream schema, the provider implementation should be fixed. This keeps schema contracts explicit and auditable.

### Part B — Mustache section helpers in workflow context (new)

Inject a set of named lambda helpers into the Chevron render context for every workflow render. These use the [Mustache lambda spec](https://mustache.github.io/mustache.5.html#Lambdas) — a callable used as a section tag renders the inner content and can transform or default the result:

```yaml
# Empty or missing → "N/A"
message: "Panel: {{#fn.na}}{{ alert.panelUrl }}{{/fn.na}}"

# Empty or missing → ""
message: "Thread: {{#fn.default}}{{ alert.slack_timestamp }}{{/fn.default}}"

# Transform
message: "Severity: {{#fn.upper}}{{ alert.severity }}{{/fn.upper}}"
```

Proposed initial helper set injected globally into render context:

```python
WORKFLOW_HELPERS = {
    "fn": {
        "default": lambda text, render: render(text) or "",
        "na":      lambda text, render: render(text) or "N/A",
        "upper":   lambda text, render: render(text).upper(),
        "lower":   lambda text, render: render(text).lower(),
        "strip":   lambda text, render: render(text).strip(),
    }
}
```

These lambdas receive the *rendered* inner text, so they operate after field resolution. An absent key renders to `""` inside the lambda (Chevron does not throw when `safe=False` is used, which the lambda call path uses), and the lambda applies the default.

### Required changes

**Backend** (`keep/iohandler/iohandler.py`):
- Define `WORKFLOW_HELPERS` dict with the lambda set
- Merge it into the context passed to `chevron.render()` in `render_context()` and `render_recursively()`

**Frontend** (`keep-ui/entities/workflows/lib/mustache.ts`):
- `extractMustacheVariables()` currently validates every token matched by `MUSTACHE_REGEX` against `ALLOWED_MUSTACHE_VARIABLE_REGEX = /^[a-zA-Z0-9._-\s]+$/`
- Section open `{{#fn.na}}`, section close `{{/fn.na}}`, and inverted `{{^...}}` tokens start with `#`, `/`, `^` and fail this regex, producing spurious "invalid variable" warnings in the builder
- Fix: filter out tokens that start with `#`, `/`, `^`, `!`, `>` before applying the variable name validation — these are Mustache sigils, not variable references

## What this does not change

- Existing workflows with no `fn.*` references are unaffected
- Provider `_format_alert()` defaulting remains the correct place for schema fields
- The `keep.*` function preprocessor is unaffected
- No changes to deduplication, fingerprinting, or alert storage

## Alternatives considered

| Approach | Why not chosen |
|----------|---------------|
| Global `safe=False` | Silently swallows template typos — `{{ alert.naem }}` renders `""` with no feedback |
| Per-workflow `safe: false` flag | Authors can opt out of safety with no defaulting syntax; schema change required |
| Pre-render context patch (scan for missing keys, inject `""`) | Two-pass render complexity; still no per-field default control |
| Jinja2 migration | `\| default("N/A")` filter is the right UX, but a breaking migration for all existing workflows |

## Status

- [x] Chevron lambda behaviour verified locally (flat and nested dict helpers, warn=False path)
- [x] Backend injection implemented in `keep/iohandler/iohandler.py` — `WORKFLOW_HELPERS` merged into Chevron context; `safe=False` auto-enabled when `{{#fn.` detected
- [x] Frontend `mustache.ts` fix implemented — sigil tokens filtered in `extractMustacheVariables()`
- [x] Workflow example: `{{#fn.default}}{{ alert.silenceURL }}{{/fn.default}}` deployed on dev (Neo Grafana → Slack workflow, revision 4)
- [x] Validated end-to-end on dev environment (all five helpers: fn.default, fn.na, fn.upper, fn.lower, fn.strip — exercised by real alert firing on dev Keep 0.49.0-svt.10)

hakatt:
## Summary

Closes #5682. Also addresses the root cause of #5070 and is a lighter-weight alternative to the Jinja migration requested in #4594.

Injects a set of named lambda helpers into every Chevron render context so workflow templates can safely reference custom or enriched alert fields that may be absent — without triggering `RenderException`:

```yaml
# Empty or missing → "N/A"
body: "Panel: {{#fn.na}}{{ alert.panelUrl }}{{/fn.na}}"

# Empty or missing → ""
if: "{{#fn.default}}{{ alert.slack_timestamp }}{{/fn.default}}"

# Transform
body: "Severity: {{#fn.upper}}{{ alert.severity }}{{/fn.upper}}"
```

## Changes

**Backend** `keep/iohandler/iohandler.py` (+23 lines):
- `WORKFLOW_HELPERS` constant with `fn.{default,na,upper,lower,strip}` Chevron lambdas
- Merged into context on every `_render()` call
- Safe mode auto-disabled when template contains `{{#fn.` (lambda already returns a safe default, so `RenderException` must not be raised)

**Frontend** `keep-ui/entities/workflows/lib/mustache.ts` (+3 lines):
- Filter Mustache sigil tokens (`#`, `/`, `^`, `!`, `>`) from `extractMustacheVariables()` so the workflow builder doesn't flag `{{#fn.na}}`/`{{/fn.na}}` as invalid variable references

## Design rationale

See #5682 for the full two-category analysis. The short version:
- **Provider schema fields** (e.g. `panelUrl`, `instance`) → defaulted in `_format_alert()`, no workflow-layer handling needed
- **Custom/enriched fields** (e.g. `slack_timestamp` from `enrich_alert`, optional labels) → no provider can default these; `fn.*` helpers give workflow authors a clean per-field default syntax

This does not change behaviour for any existing workflow that does not use `{{#fn.` sections.

## Test plan

- [x] Unit test: `fn.na` on missing key → `"N/A"`
- [x] Unit test: `fn.default` on missing key → `""`
- [x] Unit test: `fn.upper`/`fn.lower`/`fn.strip` on present value
- [x] Frontend: open workflow builder with `{{#fn.na}}{{ alert.x }}{{/fn.na}}` — no "invalid variable" warning
- [x] End-to-end: all five helpers (fn.default, fn.na, fn.upper, fn.lower, fn.strip) validated on dev Keep 0.49.0-svt.10 — workflows run without `RenderException`

### DIFF
--- a/tests/test_iohandler.py
+++ b/tests/test_iohandler.py
@@ -1023,3 +1023,43 @@ def test_concurrent_render_no_stderr_race(context_manager):
         t.join(timeout=30)
 
     assert not errors, f"Concurrent render raised errors: {errors}"
+
+
+# ── fn.* Mustache lambda helper tests ────────────────────────────────────────
+
+
+def test_fn_na_on_missing_key(mocked_context_manager):
+    """fn.na renders 'N/A' when the wrapped field is absent from the context."""
+    mocked_context_manager.get_full_context.return_value = {
+        "alert": {"name": "test-alert"},  # no 'slack_timestamp' field
+    }
+    iohandler = IOHandler(mocked_context_manager)
+    result = iohandler.render("ts={{#fn.na}}{{ alert.slack_timestamp }}{{/fn.na}}")
+    assert result == "ts=N/A", f"Expected 'ts=N/A', got '{result}'"
+
+
+def test_fn_default_on_missing_key(mocked_context_manager):
+    """fn.default renders an empty string when the wrapped field is absent."""
+    mocked_context_manager.get_full_context.return_value = {
+        "alert": {"name": "test-alert"},  # no 'silenceURL' field
+    }
+    iohandler = IOHandler(mocked_context_manager)
+    result = iohandler.render("url={{#fn.default}}{{ alert.silenceURL }}{{/fn.default}}")
+    assert result == "url=", f"Expected 'url=', got '{result}'"
+
+
+def test_fn_upper_lower_strip_on_present_value(mocked_context_manager):
+    """fn.upper, fn.lower, and fn.strip transform present field values correctly."""
+    mocked_context_manager.get_full_context.return_value = {
+        "alert": {"env": "  Production  "},
+    }
+    iohandler = IOHandler(mocked_context_manager)
+
+    upper = iohandler.render("{{#fn.upper}}{{ alert.env }}{{/fn.upper}}")
+    assert upper == "  PRODUCTION  ", f"fn.upper got '{upper}'"
+
+    lower = iohandler.render("{{#fn.lower}}{{ alert.env }}{{/fn.lower}}")
+    assert lower == "  production  ", f"fn.lower got '{lower}'"
+
+    strip = iohandler.render("{{#fn.strip}}{{ alert.env }}{{/fn.strip}}")
+    assert strip == "Production", f"fn.strip got '{strip}'"


shahargl:
@hakatt @nflx is this active?

nflx:
> @hakatt @nflx is this active?

Yes, we have it running well in our test environment and I will just finish writing the unit tests and rebase it. Thank you for the reminder @shahargl :-)

hakatt:
PR ready, (if accepted).

PR #5714, opened 2026-03-04
fix: add rca_points and traces to PROVISIONED_ENRICHMENTS

### LINKED ISSUE #5715: React error #31 crash on incident overview when enrichments contain rca_points
## Bug Description

The incident overview page crashes with React Minified Error 31 when enrichments contain `rca_points` (array of objects).

## Steps to Reproduce

1. Open an incident that has `rca_points` enrichment (array of objects)
2. Navigate to the incident overview page
3. Page crashes with "Minified React error 31: Objects are not valid as a React child"

## Root Cause

`PROVISIONED_ENRICHMENTS` array in `incident-overview.tsx` does not include `rca_points` and `traces`. When `map()` iterates over enrichments, it passes these objects to `EnrichmentEditableField`, which tries to render them as React children — objects are not valid React children.

## Expected Behavior

Incident overview page should render without crashing regardless of enrichment types present.

Aladex:
Closes #5715

## Summary

Add `rca_points` and `traces` to `PROVISIONED_ENRICHMENTS` array in incident overview.

## Problem

When enrichments contain `rca_points` (array of objects), `map()` iterates over them and passes objects to `EnrichmentEditableField`, which tries to render them as React children — causing "Minified React error 31" crash on the incident overview page.

## Fix

Include `rca_points` and `traces` in the `PROVISIONED_ENRICHMENTS` list so they are excluded from the editable enrichments loop.


shahargl:
hey @Aladex - since `rca_points` and `traces` is specific to your case, I would add some environment varialbe (e.g. `EXCLUDED_ENRICHMENTS` or smth with list of these, and then use it.

e.g. in your case
```
EXCLUDED_ENRICHMENTS=rca_points,traces
```

and then extend the array of excluded enrichments

shahargl (changes requested):
see comment

Aladex:
Hey @shahargl, thanks for the review!

I'd argue that `rca_points` and `traces` are not specific to my case — they are part of Keep's core incident model. `rca_points` already has dedicated rendering via the `RootCauseAnalysis` component in the same file, and `traces` is used by the timeline view. The issue is that these fields were simply missed when `PROVISIONED_ENRICHMENTS` was defined.

Without this fix, any incident that has RCA data will crash the overview page with React error 31, which also breaks the AI chat in the incident view.

An `EXCLUDED_ENRICHMENTS` env var would be useful for truly custom enrichment fields, but for built-in fields that already have dedicated renderers, hardcoding them in `PROVISIONED_ENRICHMENTS` seems like the right approach — consistent with how `services`, `incident_id`, etc. are already handled there.

Aladex:
Hi @shahargl, friendly ping on this one 🙂 Would love to hear your thoughts on my reply above — I believe these fields belong in the hardcoded list since they're part of the core incident model and already have dedicated renderers. Happy to adjust the approach if you still prefer the env var route though\!

shahargl:
ok sounds good to me, @Aladex let's wait for e2e tests to pass :)

PR #5716, opened 2026-03-04
fix: add rca_points and traces to PROVISIONED_ENRICHMENTS

### LINKED ISSUE #5715: React error #31 crash on incident overview when enrichments contain rca_points

Dingding-leo:
### Description
The incident overview page crashes with React Minified Error 31 when enrichments contain `rca_points` (array of objects).

### Root Cause
`PROVISIONED_ENRICHMENTS` array in `incident-overview.tsx` did not include `rca_points` and `traces`. When `map()` iterates over enrichments, it passes these objects to `EnrichmentEditableField`, which tries to render them as React children — objects are not valid React children.

### Solution
Added `rca_points` and `traces` to the `PROVISIONED_ENRICHMENTS` array.

### Changes
- `keep-ui/app/(keep)/incidents/[id]/incident-overview.tsx`: Added `rca_points` and `traces` to `PROVISIONED_ENRICHMENTS`

### DIFF
--- a/keep/providers/uptimekuma_provider/uptimekuma_provider.py
+++ b/keep/providers/uptimekuma_provider/uptimekuma_provider.py
@@ -132,18 +132,18 @@ class UptimekumaProvider(BaseProvider):
                     # Single retry
                     api = self._get_api()
                     name = api.get_monitor(monitor_id)["name"]
-            heartbeats.append(
-                AlertDto(
-                    id=heartbeat["id"],
-                    name=name,
-                    monitor_id=heartbeat["monitor_id"],
-                    description=heartbeat["msg"],
-                    status=self.STATUS_MAP.get(heartbeat["status"], "firing"),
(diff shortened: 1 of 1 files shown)


shahargl:
how is it different than https://github.com/keephq/keep/pull/5714?

PR #5726, opened 2026-03-05
fix: prevent webpack from inlining AUTH_TYPE at build time

### LINKED ISSUE #5725: OAUTH2PROXY auth type silently falls back to NOAUTH in production builds
## Bug Description

When deploying Keep with `AUTH_TYPE=OAUTH2PROXY`, the **frontend** silently falls back to `NOAUTH` authentication. The `/api/auth/providers` endpoint returns `"name": "NoAuth"` instead of `"OAuth2Proxy"`, even though `AUTH_TYPE=OAUTH2PROXY` is correctly set in the pod environment.

Note: The **backend** is not affected — it correctly reads `AUTH_TYPE` at runtime. For example, acknowledging an alert correctly shows the real user's email. The issue is only in the frontend's NextAuth session, which shows `name: "Keep"` and `email: "keep"` (NOAUTH defaults) instead of the actual user identity.

## Root Cause

`auth.config.ts` is imported by both `auth.ts` (Node.js runtime) and `middleware.ts` (Edge Runtime). When webpack bundles the middleware for Edge Runtime, it replaces `process.env.AUTH_TYPE` with its **build-time** value via DefinePlugin.

Since `AUTH_TYPE` is not set during `docker build` (see `Dockerfile.ui` line 26-27), it gets inlined as `undefined`. The provider selection then falls back to NOAUTH:

```typescript
providers:
    baseProviderConfigs[authType as keyof typeof baseProviderConfigs] ||
    baseProviderConfigs[AuthType.NOAUTH],  // ← always hits this fallback
```

## How to Reproduce

1. Build the frontend Docker image without `AUTH_TYPE` set (default behavior)
2. Deploy with `AUTH_TYPE=OAUTH2PROXY` as a runtime environment variable
3. Visit `/api/auth/providers` → shows `"name": "NoAuth"` instead of `"OAuth2Proxy"`

## Expected Behavior

`AUTH_TYPE` should be read at runtime, not inlined at build time. Any auth type set via environment variable should work without needing to rebuild the Docker image.

Eliesmbr:
## Summary
Closes #5725

`auth.config.ts` is imported by `middleware.ts` (Edge Runtime), causing webpack's DefinePlugin to replace `process.env.AUTH_TYPE` with its build-time value. Since `AUTH_TYPE` is not set during `docker build`, it gets inlined as `undefined`, silently falling back to NOAUTH regardless of the runtime environment variable.

- Add `runtimeEnv()` helper that uses `process.env[key]` (bracket notation), which webpack cannot statically analyze
- Replace direct `process.env.AUTH_TYPE` and `process.env.AUTH_SESSION_TIMEOUT` references with `runtimeEnv()` calls

## Test plan
- [x] Build frontend Docker image without `AUTH_TYPE` set (default)
- [x] Run container with `AUTH_TYPE=OAUTH2PROXY` as runtime env var
- [x] Visit `/api/auth/providers` → should show `"name": "OAuth2Proxy"` (not `"NoAuth"`)
- [x] Verify other auth types (`KEYCLOAK`, `AUTH0`, etc.) also work as runtime env vars

### DIFF
--- a/keep-ui/auth.config.ts
+++ b/keep-ui/auth.config.ts
@@ -23,11 +23,19 @@ export class BackendRefusedError extends AuthError {
   static type = "BackendRefusedError";
 }
 
-const authSessionTimeout = process.env.AUTH_SESSION_TIMEOUT
-  ? Number.parseInt(process.env.AUTH_SESSION_TIMEOUT)
+// Read env vars via bracket notation to prevent webpack DefinePlugin from
+// inlining them as `undefined` at build time.  This file is imported by
+// middleware.ts (Edge Runtime) where DefinePlugin replaces direct
+// `process.env.X` references with their build-time values.
+function runtimeEnv(key: string): string | undefined {
+  return process.env[key];
+}
+
+const authSessionTimeout = runtimeEnv("AUTH_SESSION_TIMEOUT")
+  ? Number.parseInt(runtimeEnv("AUTH_SESSION_TIMEOUT")!)
   : 30 * 24 * 60 * 60; // Default to 30 days if not set
 // Determine auth type with backward compatibility
-const authTypeEnv = process.env.AUTH_TYPE;
+const authTypeEnv = runtimeEnv("AUTH_TYPE");
 export const authType =
   authTypeEnv === MULTI_TENANT
     ? AuthType.AUTH0


shahargl:
maybe I'm missing smth - but how did it work until now? @Eliesmbr

Eliesmbr:
The backend is not affected  it reads AUTH_TYPE at runtime correctly. So API calls are authenticated properly (e.g., alerts show the correct user email on acknowledge). Only the frontend's NextAuth session is wrong. 

NOAUTH auto-signs in silently  there's no error, no crash. The user gets a session with name: "Keep" and email: "keep", but since the backend handles actual auth via headers/tokens independently, most things still appear to work.

I just saw that when migrating my instance from DB user login to OAUTH2 proxy. The user on the webui showed me "Keep" although I logged in via my oauth proxy
image

I saw that `/api/auth/session` showed my this:
```{
  "user": {
    "name": "Keep",
    "email": "keep",
    "accessToken": "{\"tenant_id\":\"keep\",\"user_id\":\"keep-user-for-no-auth-purposes\"}",
    "tenantId": "keep",
    "role": "user",
    "tenantIds": [
      {
        "tenant_id": "keep",
        "tenant_name": "Tenant of Keep (tenant_id: keep)"
      },
      {
        "tenant_id": "keep2",
        "tenant_name": "Tenant of another Keep (tenant_id: keep2)"
      }
    ]
  },
  "expires": "2026-04-04T10:08:22.113Z",
  "accessToken": "{\"tenant_id\":\"keep\",\"user_id\":\"keep-user-for-no-auth-purposes\"}",
  "tenantId": "keep",
  "userRole": "user"
}
```
Edit:
`/api/auth/providers`

```
{
  "credentials": {
    "id": "credentials",
    "name": "NoAuth",
    "type": "credentials",
    "signinUrl": "https://xxxx/api/auth/signin/credentials",
    "callbackUrl": "https://xxxx/api/auth/callback/credentials"
  },
  "tenant-switch": {
    "id": "tenant-switch",
    "name": "Tenant Switch",
    "type": "credentials",
    "signinUrl": "https://xxxx/api/auth/signin/tenant-switch",
    "callbackUrl": "https://xxxxx/api/auth/callback/tenant-switch"
  }
}
```

Both backend and frontend are starting with the correct auth type
```
Starting Nextjs [http://keep-backend:8080]
AUTH_TYPE: OAUTH2PROXY
```

PR #6084, opened 2026-03-09
fix: prevent incident table horizontal overflow from long summaries

### LINKED ISSUE #6083: [🐛 Bug]: Incident table stretches horizontally when summary is long
## Bug description

When an incident has a long summary text, the "Incident" column in the incidents table stretches horizontally, pushing the entire table beyond the viewport width. This forces horizontal scrolling and breaks the layout.

The root cause is twofold:
1. No `max-width` on the incident name column container
2. `FormattedContent` renders summary as HTML with `<p>` tags inside a `dangerouslySetInnerHTML` div. The `line-clamp-3` on the outer container doesn't penetrate into block-level child elements like `<p>`, so text is not actually clamped — it overflows horizontally instead.

## How to reproduce

1. Create or generate an incident with a long summary (e.g. a detailed AI-generated summary)
2. Open the Incidents list page
3. Observe that the table stretches beyond the screen width

## Expected behavior

The summary column should have a bounded width. Long summary text should wrap and be truncated with ellipsis.

## Suggested fix

In `incidents-table.tsx`, render summary as plain text (strip HTML tags) instead of using `FormattedContent` with `dangerouslySetInnerHTML`. The incidents list table doesn't need rich HTML formatting — plain text with `line-clamp` works correctly:

```diff
- {summary ? (
-   <FormattedContent
-     content={summary}
-     format="html"
-     className="text-pretty overflow-hidden overflow-ellipsis line-clamp-3"
-   />
- ) : null}
+ {summary ? (
+   <div className="line-clamp-2 overflow-hidden text-sm text-gray-500">
+     {summary.replace(/<[^>]*>/g, "")}
+   </div>
+ ) : null}
```

Related: #3438 (was closed but didn't fully fix the horizontal overflow)

tysoncung:
## What

Fixes #6083

The incidents table stretches horizontally beyond the viewport when an incident has a long AI-generated summary.

## Root Cause

Two issues combine:

1. **`FormattedContent` renders HTML via `dangerouslySetInnerHTML`**, producing block-level elements (`<p>`, `<ul>`, etc.) that create their own block formatting context. `line-clamp` on the outer container cannot penetrate into these child block elements, so the text is never clamped.

2. **The table cell has no `max-width` or overflow constraint**, so it grows unbounded with the unclamped content, stretching the entire table.

### 1. Add `plain` prop to `FormattedContent`

New optional `plain` prop strips HTML tags and renders as plain inline text, allowing `line-clamp` to work correctly.

**`keep-ui/shared/ui/FormattedContent/FormattedContent.tsx`:**
- Added `stripHtmlTags()` utility
- Added `plain` prop to `FormattedContentProps`
- When `plain` is true, strips HTML and renders as a plain text div

### 2. Constrain the Incident column cell

**`keep-ui/features/incidents/incident-list/ui/incidents-table.tsx`:**
- Added `max-w-md overflow-hidden` to the cell container
- Set `plain` prop on `FormattedContent` so `line-clamp-3` works
- Added `text-sm text-gray-500` for visual hierarchy between title and summary

### DIFF
--- a/ee/identitymanager/identity_managers/keycloak/keycloak_authverifier.py
+++ b/ee/identitymanager/identity_managers/keycloak/keycloak_authverifier.py
@@ -346,6 +346,10 @@ class KeycloakAuthVerifier(AuthVerifierBase):
         if self.keycloak_multi_org:
             return super()._authorize(authenticated_entity)
 
+        # API key auth does not carry a Keycloak token; fall back to RBAC
+        if not getattr(authenticated_entity, "token", None):
+            return super()._authorize(authenticated_entity)
+
         # for single tenant Keycloaks, use Keycloak's UMA to authorize
         try:
             permission = UMAPermission(
@@ -369,6 +373,10 @@ class KeycloakAuthVerifier(AuthVerifierBase):
     def authorize_resource(
         self, resource_type, resource_id, authenticated_entity: AuthenticatedEntity
     ) -> None:
+        # API key auth does not carry a Keycloak token; skip per-resource UMA check
+        if not getattr(authenticated_entity, "token", None):
+            return
+
         # use Keycloak's UMA to authorize
         try:
             permission = UMAPermission(
--- a/keep-ui/app/(keep)/incidents/[id]/incident-overview.tsx
+++ b/keep-ui/app/(keep)/incidents/[id]/incident-overview.tsx
@@ -46,6 +46,8 @@ const PROVISIONED_ENRICHMENTS = [
   "incident_title",
   "environments",
   "repositories",
+  "rca_points",
+  "traces",
 ];
 
 interface Props {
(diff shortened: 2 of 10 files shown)


Aladex:
I opened #6083 and went through several iterations trying to fix this. A few notes on this PR:

**`max-w-md` on the div won't work well.** We tried `max-w-sm` first — it clips text mid-character on narrower viewports because the fixed max-width doesn't adapt to the available table space. `max-w-md` will have the same problem, just at a slightly larger breakpoint.

**The fix needs to be on the `<td>`, not the inner div.** With `table-layout: auto` (default), the table cell grows with its content regardless of the inner div's max-width. The constraint needs to be at the cell level. We found that adding `overflow-hidden` via the column's `meta.tdClassName` is what actually prevents the table from stretching:

```tsx
meta: {
  tdClassName: "overflow-hidden",
},
```

We're still testing the right combination on our deployment — will open a proper PR once we've verified it works end-to-end.

shahargl:
@Aladex @tysoncung can you sign the CLA?

Aladex:
> @Aladex @tysoncung can you sign the CLA?

Dear @shahargl 
I’d recommend rejecting this pull request. It will completely break the table view. Please take a look at the other one I made a bit later - it handles this error correctly.
https://github.com/keephq/keep/pull/6085

shahargl:
closing @Aladex due to your recommendation

PR #6085, opened 2026-03-09
fix: prevent incident table horizontal overflow from long summaries

### LINKED ISSUE #6083: [🐛 Bug]: Incident table stretches horizontally when summary is long

Aladex:
## Summary

Fixes #6083

When an incident has a long AI-generated summary, the "Incident" column stretches the table beyond the viewport.

`FormattedContent` renders summary as HTML via `dangerouslySetInnerHTML`, producing block-level elements (`<p>`, `<ul>`) that create their own block formatting context. `line-clamp` on the outer container cannot penetrate into these child elements, so the text is never clamped — and the table cell grows unbounded.

## Changes

- **`FormattedContent.tsx`**: Add `plain` prop that strips HTML tags and renders as plain inline text, allowing `line-clamp` to work correctly
- **`incidents-table.tsx`**: Use `plain` prop on `FormattedContent` for the summary, add `overflow-hidden` on the `<td>` via `meta.tdClassName` to prevent the cell from stretching the table

## Test plan

- [x] Open Incidents list with long AI-generated summaries — table no longer overflows
- [x] Summary shows up to 2 lines with ellipsis truncation
- [x] Incident detail page still renders full HTML summary (no regression from `plain` prop, which is only used in the table)

### DIFF
--- a/tests/e2e_tests/incidents_alerts_tests/test_xss_protection.py
+++ b/tests/e2e_tests/incidents_alerts_tests/test_xss_protection.py
@@ -110,12 +110,14 @@ def test_legit_html_content(
             "table[data-testid='incidents-table'] tbody tr",
             has_text=legit_html_incident["user_generated_name"],
         ).first
-        html_content = incident_row.inner_html()
-        assert "<h2>" in html_content, "H2 tag not found in HTML"
-        assert "<code>" in html_content, "Code tag not found in HTML"
+        # Incident list renders summaries as plain text (tags stripped) for
+        # proper CSS line-clamp. Verify text content is preserved.
+        text_content = incident_row.inner_text()
+        assert "Test Failure" in text_content, "Summary text not found"
         assert (
-            '<a href="https://google.com">' in html_content
-        ), "Link tag not found in HTML"
+            "test_csb_upload_send_two_times_same_sequence_number" in text_content
+        ), "Code text not found in summary"
+        assert "Google" in text_content, "Link text not found in summary"
     except Exception:
         save_failure_artifacts(browser, log_entries=[])
         raise


Aladex:
Hey @shahargl, could you take a look at this one when you get a chance?

We're running Keep in production and currently have to maintain a mounted patch for this fix. The incident table overflow is clearly visible with long AI-generated summaries — they break the table layout horizontally.

Here's how it looks after the fix:

Screenshot_20260313_080437

Would be great to get this merged so we can drop the patch. Thanks!

shahargl:
@Aladex would love to merge it but can you have a look on the e2e test that fails? it look like it fails across all configurations

Aladex:
@shahargl Fixed the failing e2e test in the latest commit.

The test `test_legit_html_content` was checking for raw HTML tags (`<h2>`, `<code>`, `<a>`) in the incident table row. Since the summary column now renders as plain text (tags stripped) to allow proper CSS `line-clamp`, those tags are no longer in the DOM at the table level.

Updated the test to verify that the **text content** is preserved instead (e.g. "Test Failure", "Google", the code snippet text).

To clarify what this PR does:
- **Incident list table**: summaries are rendered as plain text — HTML tags stripped so that `line-clamp` works correctly without block-level elements breaking it
- **Incident detail view**: HTML rendering is untouched, rich content still renders normally

shahargl:
merged :) ty!

PR #6107, opened 2026-03-18
feat: SNMP provider - trap listener + OID mapping

zhaog100:
Closes #2112

## SNMP Provider

- **SNMPv2c & SNMPv3** support with auth (MD5/SHA) and privacy (DES/AES)
- **Trap listener** via pysnmp-lextudio NotificationReceiver on configurable host:port
- **Optional polling** - periodic OID queries against configured targets
- **OID-to-alert mapping** - JSON-configurable OID patterns → alert name + severity
- **Clean lifecycle** - daemon threads with dispose()
- **Optional dependency** - graceful fallback if pysnmp not installed


shahargl:
N/A

PR #6124, opened 2026-03-24
fix: resolve 4 critical bugs in UptimeKuma provider

### LINKED ISSUE #5655: [🐛 Bug]: UptimeKuma provider has multiple critical bugs in pull mode and webhook handling

lalalic:
## Summary

Fixes **4 critical bugs** in the UptimeKuma provider that cause it to malfunction in both pull mode and webhook handling.

Closes #5655

---

### Bug 1: `heartbeats.append()` outside for-loop (indentation error)

**Problem:** Only the last monitor is reported. The `append()` call was outside the `for key in response:` loop due to wrong indentation.

**Fix:** Moved `heartbeats.append()` inside the loop so all monitors are collected.

---

### Bug 2: `_format_datetime` TypeError

**Problem:** `return dt + offset` fails with TypeError — `dt` is a string like `"2023-01-01 12:00:00"` and `offset` is an integer (minutes). You cannot add a string and an int.

**Fix:** Properly parses the `localDateTime` string into a `datetime` object and applies `timezoneOffset` (minutes) as a timezone-aware datetime using `timedelta`.

---

### Bug 3: `_format_alert` KeyError on webhook

**Problem:** Direct dictionary access (`event["monitor"]["url"]`, `event["msg"]`, `event["heartbeat"]["status"]`, etc.) raises `KeyError` when fields are absent — which happens for some monitor types (e.g., ping monitors have no `url`).

**Fix:** Replaced all direct access with `.get()` calls with safe defaults.

---

**Problem:** `api.disconnect()` was only called in the happy path. If any exception occurred during heartbeat collection, the socket.io connection would leak.

**Fix:** Wrapped `api.disconnect()` in a `finally` block to guarantee cleanup.

---

### Testing
All changes use defensive coding patterns (`.get()` with defaults, `try/except`, `finally`) consistent with the rest of the codebase. The fixes are minimal and targeted — no behavioral changes beyond the bug corrections.

### DIFF
--- a/keep/providers/uptimekuma_provider/uptimekuma_provider.py
+++ b/keep/providers/uptimekuma_provider/uptimekuma_provider.py
@@ -173,7 +173,7 @@ class UptimekumaProvider(BaseProvider):
 
         alert = AlertDto(
             id=monitor.get("id"),
-            name=monitor.get("name"),
+            name=monitor.get("name") or "unknown",
             monitor_url=monitor.get("url"),
             status=cls.STATUS_MAP.get(heartbeat.get("status"), "firing"),
             description=event.get("msg") or heartbeat.get("msg", ""),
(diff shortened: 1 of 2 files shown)


lalalic:
All unit tests now pass after commit 1d82d50. The only remaining CI failures are:
- `run-postgresql-without-redis / run-tests` (E2E test — not related to the UptimeKuma provider changes)
- `Vercel` (deployment preview authorization, not code-related)

All the originally failing tests mentioned in the review are now fixed:
- test_completely_empty_event 
- test_all_monitors_reported
- test_connection_closed_on_success
- test_deduplication_sanity_2 (timeout was flaky)
- test_workflow_keep_notify_after_another_foreach (sqlite lock was flaky)

Could you please re-review? Thank you!

PR #6180, opened 2026-04-01
fix: increase secret value column from VARCHAR(255) to TEXT

### LINKED ISSUE #5353: [🐛 Bug]: Secrets truncated in DB mode cause provider OAuth JSON parse errors
**Describe the bug**
When using MySQL DB secret manager (SECRET_MANAGER_TYPE=db), provider OAuth secrets (e.g., PagerDuty) were truncated due to the secret.value column being defined as VARCHAR(255).
As a result, long JSON tokens (access/refresh tokens from OAuth) were cut off at 255 characters.
This caused JSON parsing to fail with errors such as: `Unterminated string starting at: line 1 column 247 (char 246)` and prevented provider installation (HTTP 400/500 during OAuth callback).

**To Reproduce**

1. Deploy Keep with: `SECRET_MANAGER_TYPE=db` and a MySQL database.
2. Attempt to install an OAuth-based provider (e.g., PagerDuty).
3. Observe the provider installation fail and backend logs show truncated JSON or an “unterminated string” error.
4. Inspect the secret table: `SELECT `key`, CHAR_LENGTH(`value`) FROM secret ORDER BY CHAR_LENGTH(`value`) DESC LIMIT 10;`. Values are cut off at exactly 255 characters.

**Expected behavior**
OAuth token JSON should be stored completely, and provider installation should succeed without data truncation or JSON parsing errors.

**Additional context**
Root cause: the value column in the secret table is defined as VARCHAR(255), which is too short for OAuth JSON payloads.

Switching to SECRET_MANAGER_TYPE=k8s resolves the issue entirely since Kubernetes Secrets have no such length limitation.

A permanent schema migration should be added to the Keep backend to widen the column automatically or validate stored value length.

DragonBot00:
### Problem
Fixes #5353

When `SECRET_MANAGER_TYPE=db`, the `secret.value` column in the `Secret` SQLModel was defined as `value: str`, which SQLModel maps to `VARCHAR(255)` by default. This causes any secret value exceeding 255 characters (such as OAuth access tokens, refresh tokens, or JWT tokens) to be **silently truncated** at the database level.

### Root Cause
In `keep/api/models/db/secret.py`, the `value` field lacked an explicit SQLAlchemy column type:

```python
# Before (defaults to VARCHAR(255))
value: str
```

### Solution
Explicitly set the column type to `sa.Text` (unlimited length):

```python
# After (TEXT - no length limit)
value: str = Field(sa_column=sa.Column(sa.Text, nullable=False))
```

### Changes
1. **`keep/api/models/db/secret.py`** — Changed `value` field to use `sa.Text` instead of the implicit `VARCHAR(255)`.
2. **`keep/api/models/db/migrations/versions/2026-04-01-07-55_a1b2c3d4e5f6.py`** — Added Alembic migration to `ALTER TABLE secret ALTER COLUMN value TYPE TEXT` on existing deployments.

### Testing
- New deployments: The `Secret` table will be created with a `TEXT` column for `value`.
- Existing deployments: Running `alembic upgrade head` will apply the migration and widen the column with no data loss.

### Impact
- **No breaking changes** — TEXT columns are fully backward-compatible with VARCHAR.
- Fixes silent truncation of OAuth tokens and any other secrets > 255 chars.

### DIFF
--- a/keep/api/models/db/migrations/versions/2026-04-01-07-55_a1b2c3d4e5f6.py
+++ b/keep/api/models/db/migrations/versions/2026-04-01-07-55_a1b2c3d4e5f6.py
@@ -1,0 +1,40 @@
+"""increase secret value column from VARCHAR to TEXT
+
+Revision ID: a1b2c3d4e5f6
+Revises: 9dd1be4539e0
+Create Date: 2025-04-01 02:50:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "a1b2c3d4e5f6"
+down_revision = "9dd1be4539e0"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Change `secret.value` from VARCHAR(255) to TEXT so that long OAuth tokens
+    # (and other secrets exceeding 255 characters) are no longer silently truncated
+    # when SECRET_MANAGER_TYPE=db is used.
+    op.alter_column(
+        "secret",
+        "value",
+        existing_type=sa.String(length=255),
+        type_=sa.Text(),
+        existing_nullable=False,
+    )
+
+
+def downgrade() -> None:
+    # WARNING: downgrading will truncate any values longer than 255 chars.
+    op.alter_column(
+        "secret",
+        "value",
+        existing_type=sa.Text(),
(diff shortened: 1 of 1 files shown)


DragonBot00:

TheCodingDragon0:

DragonBot00:

PR #6185, opened 2026-04-01
fix: populate condition field when editing extraction rules

### LINKED ISSUE #5201: [🐛 Bug]: Extraction rule: unable to edit condition after saving
**Describe the bug**
If you create an extraction rule with a condition, you cannot edit it afterwards because it does not show up in the edit view.

**To Reproduce**

1. Create extraction rule with condition (e.g. `source.contains('grafana')`
2. See in the overview that there is a condition
3. Click edit
4. No condition is shown

**Expected behavior**
Condition to be shown

**Screenshots**

Overview:

Image

Edit view:

Image

**Additional context**
Version: 0.46.1 | Build: 321e50

DragonBot00:
## Description

When creating an extraction rule with a condition (e.g. source.contains('grafana')), the condition is correctly saved and visible in the extraction rules overview. However, when clicking **Edit** on an existing rule, the condition field is empty — the previously saved condition is not pre-populated.

## Root Cause

The AlertsRulesBuilder component initializes its internal CEL state from constructCELRules(selectedPreset), which returns an empty string when no preset is provided. Although the component accepts a defaultQuery prop, it was never wired into the state initialization, so the condition value passed from the edit form was silently ignored.

Additionally, the extraction edit form passes condition as defaultQuery, but condition is populated asynchronously via a useEffect. By the time condition is set, AlertsRulesBuilder has already mounted with an empty string.

## Fix

Two changes were made:

1. **lerts-rules-builder.tsx**: Wire the defaultQuery prop into the useCelState initialization so it is used as the fallback default when no preset CEL is available.

2. **create-or-update-extraction-rule.tsx**:
   - Derive initialCondition directly from extractionToEdit?.condition (synchronously available on render) instead of relying on the async state value.
   - Add a key prop to AlertsRulesBuilder based on extractionToEdit?.id so the component remounts with the correct initial condition whenever a different rule is opened for editing.

## Testing

1. Create an extraction rule with a condition (e.g. source.contains('grafana'))
2. Click **Edit** on the rule
3. Verify the condition field is pre-populated with the saved value

Fixes #5201

### DIFF
--- a/keep-ui/app/(keep)/extraction/create-or-update-extraction-rule.tsx
+++ b/keep-ui/app/(keep)/extraction/create-or-update-extraction-rule.tsx
@@ -61,9 +61,15 @@ export default function CreateOrUpdateExtractionRule({
       setIsPreFormatting(extractionToEdit.pre);
       setAttribute(extractionToEdit.attribute);
       setRegex(extractionToEdit.regex);
+      // condition is initialized directly from extractionToEdit to ensure
+      // the AlertsRulesBuilder receives it before mounting
       setCondition(extractionToEdit.condition ?? "");
     }
   }, [extractionToEdit]);
+
+  // Derive the condition to pass as defaultQuery so the AlertsRulesBuilder
+  // receives the correct initial value when the edit form opens.
+  const initialCondition = extractionToEdit?.condition ?? "";
 
   const clearForm = () => {
     setExtractionName("");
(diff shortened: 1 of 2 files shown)


DragonBot00:

TheCodingDragon0:

DragonBot00:

PR #6213, opened 2026-04-06
fix: set event in foreach AlertDto path so incident resolution check runs

### LINKED ISSUE #6194: [🐛 Bug]: Incidents with "Resolve on all_resolved" do not resolve when alert is enriched via workflow
**Describe the bug**
When all alerts, in an incident set to resolve on all_resolved, are resolved it does not resolve the incident.

These alerts are resolved by enriching the alert to "resolved" through a workflow.

Manually resolving these alerts does resolve the incident. However for my purposes I need to be able to resolve incidents programatically. 

**To Reproduce**

I am on version 0.51 (latest).

Create an incident via correlation set to Resolve on all_resolved. 

Resolve those alerts via a workflow. The incident will not resolve.

This is the action I am using to resolve the alerts:

`  
actions:

    - name: resolve-dag-alerts
    
      foreach: " {{ steps.get-firing-alerts.results }} "

      provider:

        type: mock

        with:

          enrich_alert:

            - key: status
            
              value: resolved

              disposable: true

            - key: resolved_by
            
              value: keep-workflow

`

**Expected behavior**

When a workflow enriches all alerts to resolve in an incident set to resolve on all alerts resolved, that incident should resolve.

**Screenshots**

Image

**Additional context**

There is another issue describing the same problem which apparently has been fixed, however this does not seem to be the case for me.

https://github.com/keephq/keep/issues/3142

DragonBot00:
Fixes #6194

Problem: When enriching alerts to resolved via a workflow foreach loop with AlertDto objects, the incident resolution check is silently skipped.

Root Cause: In BaseProvider._enrich(), the event variable is set in the tuple path (line 221) but never set in the AlertDto path (line 223-224). The resolution check is gated on if event and should_check_incidents_resolution, so it always evaluates to False for AlertDto foreach.

Fix: One line - assign event = foreach_context in the isinstance(foreach_context, AlertDto) branch.

### DIFF
--- a/keep/providers/base/base_provider.py
+++ b/keep/providers/base/base_provider.py
@@ -222,6 +222,7 @@ class BaseProvider(metaclass=abc.ABCMeta):
 
             if isinstance(foreach_context, AlertDto):
                 fingerprint = foreach_context.fingerprint
+                event = foreach_context
             # if we are in a dict context, use the fingerprint from the dict
             elif isinstance(foreach_context, dict) and "fingerprint" in foreach_context:
                 fingerprint = foreach_context.get("fingerprint")
--- a/keep/providers/site24x7_provider/site24x7_provider.py
+++ b/keep/providers/site24x7_provider/site24x7_provider.py
@@ -236,7 +236,7 @@ class Site24X7Provider(BaseProvider):
 
         labels = event.get("LABELS", "")
         if isinstance(labels, str) and labels:
-            labels = [part.strip() for part in labels.split(",") if part.strip()]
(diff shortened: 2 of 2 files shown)


DragonBot00:
Hi maintainers, this one-line fix resolves #6194 — incidents don't resolve when alerts are enriched via workflow foreach because the event variable is never set in the AlertDto path, silently skipping check_incident_resolution().

The root cause was confirmed by multiple community members. Would love to get this merged — it's been sitting for a few days. Thanks!

shahargl (approved):
Lgtm

PR #6219, opened 2026-04-06
fix: generate UUID for prometheus alerts when id is not a valid uuid

### LINKED ISSUE #4985: [🐛 Bug]: Prometheus provider stores alert name as alert id
**Describe the bug**
Prometheus provider stores string values in `id` which breaks links in workflows:

Image

**To Reproduce**

1. Configure Prometheus+Alertmanager to send alerts to Keep via webhook API (`/backend/alerts/event`)
2. Receive some events
3. Build a workflow based on those events
4. Observe 500 when clicking on the alert in "Triggered by":

Image

**Expected behavior**
Generate usable UUID

**Screenshots**
included

**Additional context**

Looks like it happens here: https://github.com/keephq/keep/blob/3caff53a48e2d6ee7c86988e11a2637327c43d53/keep/providers/prometheus_provider/prometheus_provider.py#L182

DragonBot00:
Fixes #4985

When Prometheus sends alerts via Alertmanager webhook, the id field contains the alertname string (e.g., HostOomKillDetected) rather than a UUID. This causes 500 errors when clicking alerts in workflow 'Triggered by' links, which expect a valid alert ID.

## Changes
- Added UUID generation for alert IDs that are not valid UUIDs
- Preserves existing UUID-format IDs from the payload
- Falls back to generating a new UUID when neither id nor lertname is UUID-formatted

## How it works
- If id is present and is a valid UUID → use it as-is
- If id is present but not a valid UUID (e.g., alertname string) → generate a new UUID
- If id is missing → try labels.alertname, or generate a UUID as last resort

## Testing
- [ ] Existing prometheus provider tests pass
- [ ] Alert with UUID-format id preserves the original id
- [ ] Alert with string alertname (e.g., HostOomKillDetected) generates a new UUID

### DIFF
--- a/keep/providers/prometheus_provider/prometheus_provider.py
+++ b/keep/providers/prometheus_provider/prometheus_provider.py
@@ -5,6 +5,7 @@ PrometheusProvider is a class that provides a way to read data from Prometheus.
 import dataclasses
 import datetime
 import os
+import uuid as uuid_module
 
 import pydantic
 import requests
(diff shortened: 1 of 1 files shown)


DragonBot00:
Hi maintainers, this PR has been open for a day with no review. The fix is a one-line change that resolves #4985 — Prometheus alerts store the alertname string as id instead of a UUID, breaking workflow links.

Would appreciate a quick review when you get a chance. Happy to adjust based on feedback.

PR #6247, opened 2026-04-14
fix: seed tenant for SSO auth types (OKTA/KEYCLOAK/AZUREAD) on fresh database

### LINKED ISSUE #6246: [🐛 Bug]: AUTH_TYPE=OKTA/KEYCLOAK/AZUREAD skips tenant seeding on fresh database → infinite redirect loop
# [🐛 Bug]: AUTH_TYPE=OKTA/KEYCLOAK/AZUREAD skips tenant seeding on fresh database → infinite redirect loop

**Describe the bug**

When Keep is deployed with `AUTH_TYPE=OKTA` (or `KEYCLOAK` / `AZUREAD`) against a fresh database (new PG cluster, PVC recreated, etc.), the `tenant` table remains empty. Authenticated users land in an infinite redirect loop between `/signin` and `/incidents` — Okta login succeeds and the session cookie is set, but every backend API call returns empty results because no tenant row exists for the hardcoded `tenant_id="keep"`.

The root cause is in `keep/api/config.py`: the `on_starting()` function gates tenant creation behind an explicit auth-type allowlist that includes `DB`, `NOAUTH`, `OAUTH2PROXY`, and `ONELOGIN` — but omits `OKTA`, `KEYCLOAK`, and `AZUREAD`. So `try_create_single_tenant()` is never called for these auth types and the tenant table stays empty.

This went unnoticed because in normal operation the PVC persists across redeployments — the tenant row created during initial setup survives indefinitely. The bug only surfaces when the database is bootstrapped from scratch.

**To Reproduce**

1. Deploy Keep v0.51.0 with `AUTH_TYPE=OKTA` (or `KEYCLOAK` / `AZUREAD`)
2. Point it at a fresh PostgreSQL database (empty, post-migration)
3. Log in via Okta SSO — authentication succeeds
4. Observe the UI enters an infinite redirect loop between `/signin` and `/incidents`

**Expected behavior**

`try_create_single_tenant()` should run for all single-tenant auth types (including SSO-based ones), seeding the `tenant` table on startup so the application is functional immediately after a fresh deployment.

**Screenshots**

N/A — the symptom is an infinite redirect loop with no rendered error page.

Backend API calls return `{"items":[],"count":0}` (Content-Length: 43) for every tenant-scoped query because `OktaAuthVerifier` returns the hardcoded `tenant_id="keep"` which has no matching row in the `tenant` table.

**Additional context**

Affected auth types: any SSO-based single-tenant auth type not in the allowlist — **OKTA**, **KEYCLOAK**, **AZUREAD**.

**Proposed fix:** Add the missing auth types to the allowlist in `keep/api/config.py`:

```python
if AUTH_TYPE in [
    IdentityManagerTypes.DB.value,
    IdentityManagerTypes.NOAUTH.value,
    IdentityManagerTypes.OAUTH2PROXY.value,
    IdentityManagerTypes.ONELOGIN.value,
    IdentityManagerTypes.OKTA.value,         # <-- add
    IdentityManagerTypes.KEYCLOAK.value,     # <-- add
    "azuread",                               # <-- add (enterprise-only, no enum entry)
    "no_auth",
    "single_tenant",
]:
```

For SSO auth types, `create_default_user` should be `False` (authentication is delegated to the IdP):

```python
excluded_from_default_user = [
    IdentityManagerTypes.OAUTH2PROXY.value,
    IdentityManagerTypes.ONELOGIN.value,
    IdentityManagerTypes.OKTA.value,
    IdentityManagerTypes.KEYCLOAK.value,
    "azuread",
]
```

`try_create_single_tenant()` is already idempotent, so this change is safe for existing deployments.

Note: `AZUREAD` does not have an entry in the `IdentityManagerTypes` enum — it is enterprise-only (`ee/`) and loaded dynamically by string. The fix uses the raw string `"azuread"` accordingly.

**Workaround:** Add an idempotent tenant seed to the backend init container:

```python
conn = psycopg2.connect(host=host, user=user, password=pw, dbname="keep", connect_timeout=3)
conn.autocommit = True
with conn.cursor() as cur:
    cur.execute(
        "INSERT INTO tenant (id, name) VALUES ('keep', 'Single Tenant') "
        "ON CONFLICT (id) DO NOTHING;"
    )
conn.close()
```

**File references:**
- Bug location: `keep/api/config.py` → `on_starting()`
- Tenant creation: `keep/api/core/db_on_start.py` → `try_create_single_tenant()`
- Okta auth verifier hardcodes `tenant_id="keep"`: `keep/identitymanager/identity_managers/okta/okta_authverifier.py`

**Version:** Keep v0.51.0

ahbeigi:
on_starting() skipped try_create_single_tenant() for OKTA, KEYCLOAK, and AZUREAD auth types because they were missing from the allowlist. On a fresh database this leaves the tenant table empty, causing an infinite redirect loop after successful SSO login.

Add all three to the allowlist and to excluded_from_default_user (SSO types delegate auth to the IdP so no local user is needed).

try_create_single_tenant() is already idempotent, so existing deployments are unaffected.

Closes #6246  

## 📑 Description

When `AUTH_TYPE` is set to `OKTA`, `KEYCLOAK`, or `AZUREAD` and the database is fresh (new PG cluster, PVC recreated, etc.), `on_starting()` in `keep/api/config.py` never calls `try_create_single_tenant()` because these auth types are missing from the allowlist. The `tenant` table stays empty, and authenticated users hit an infinite redirect loop between `/signin` and `/incidents`.

This PR adds the three missing auth types to the allowlist and to `excluded_from_default_user` so the tenant row is seeded on startup without creating a local default user (SSO delegates auth to the IdP).

Note: `AZUREAD` has no `IdentityManagerTypes` enum entry (it is enterprise-only under `ee/` and loaded dynamically), so the raw string `"azuread"` is used.

- [x] Add `OKTA`, `KEYCLOAK`, `"azuread"` to auth-type allowlist
- [x] Add `OKTA`, `KEYCLOAK`, `"azuread"` to `excluded_from_default_user`

## ✅ Checks

## ℹ Additional Information

The bug went unnoticed because the PVC normally persists across redeployments — the tenant row created during initial setup survives indefinitely. It only surfaces on a fresh database bootstrap.

Workaround for anyone on older versions: add an idempotent `INSERT INTO tenant (id, name) VALUES ('keep', 'Single Tenant') ON CONFLICT (id) DO NOTHING;` to the backend init container.

### DIFF
--- a/keep/api/config.py
+++ b/keep/api/config.py
@@ -59,12 +59,18 @@ def on_starting(server=None):
         IdentityManagerTypes.NOAUTH.value,
         IdentityManagerTypes.OAUTH2PROXY.value,
         IdentityManagerTypes.ONELOGIN.value,
+        IdentityManagerTypes.OKTA.value,
+        IdentityManagerTypes.KEYCLOAK.value,
+        "azuread",  # enterprise-only, no enum entry
         "no_auth",  # backwards compatibility
         "single_tenant",  # backwards compatibility
     ]:
         excluded_from_default_user = [
             IdentityManagerTypes.OAUTH2PROXY.value,
             IdentityManagerTypes.ONELOGIN.value,
+            IdentityManagerTypes.OKTA.value,
+            IdentityManagerTypes.KEYCLOAK.value,
+            "azuread",
         ]
         # for oauth2proxy, we don't want to create the default user
         try_create_single_tenant(


shahargl:
closing since I've already merged smth similar

PR #6254, opened 2026-04-15
fix(auth): restore get_roles() for Okta — API key creation broken under AUTH_TYPE=OKTA

### LINKED ISSUE #6253: [🐛 Bug]: AUTH_TYPE=OKTA — API key creation broken; role dropdown shows "No options"
**Describe the bug**
When Keep is deployed with `AUTH_TYPE=OKTA`, the Settings → API Keys UI is non-functional. The role dropdown renders "No options" and no API key can be created through the UI.

Root cause: `OktaIdentityManager.get_roles()` explicitly overrides the base class with `return []`. The base class (`BaseIdentityManager`) already has a complete, correct implementation that builds the role list from the locally-defined predefined roles in `rbac.py` (`admin`, `noc`, `webhook`, `workflowrunner`) — none of which require any Okta API call. The override is an incomplete stub left from the initial implementation.

**To Reproduce**
1. Deploy Keep with `AUTH_TYPE=OKTA`
2. Log in successfully via Okta SSO
3. Navigate to Settings → API Keys
4. Click "Create API Key"
5. Observe the role dropdown is empty ("No options") — key creation is blocked

**Expected behavior**
The role dropdown should list the predefined Keep roles (`admin`, `noc`, `webhook`, `workflowrunner`), identical to behaviour under other auth modes.

**Screenshots**

Image

**Additional context**
The same stub pattern may affect other SSO identity managers (`KEYCLOAK`, `AZUREAD`, `ONELOGIN`) — worth auditing their `get_roles()` overrides.

Workaround: create keys directly via the backend pod using `kubectl exec` + Python against the DB and secret manager.

ahbeigi:
OktaIdentityManager.get_roles() was overriding the base class with return [] — causing the role dropdown in Settings → API Keys to render "No options" and blocking API key creation entirely under AUTH_TYPE=OKTA.

Predefined roles (admin, noc, webhook, workflowrunner) are defined locally in rbac.py and require no Okta API call, so the base class implementation is correct as-is. Remove the stub override so it falls through to BaseIdentityManager.get_roles().

Also fix DEFAULT_ROLE_NAME in OktaAuthVerifier: "user" is not a valid role in rbac.py and causes a 403 for tokens without an explicit role claim. Change to "noc" (least-privileged predefined role).

Closes #6253

## 📑 Description

- [x] Remove `get_roles()` override in `OktaIdentityManager` — falls through to `BaseIdentityManager.get_roles()` which correctly returns predefined roles from `rbac.py`
- [x] Fix `DEFAULT_ROLE_NAME` in `OktaAuthVerifier` from `"user"` to `"noc"`

## ✅ Checks

## ℹ Additional Information

The same stub pattern may be present in other SSO identity managers (`KEYCLOAK`, `AZUREAD`, `ONELOGIN`) — worth a follow-up audit of their `get_roles()` overrides.

### DIFF
--- a/keep/identitymanager/identity_managers/okta/okta_authverifier.py
+++ b/keep/identitymanager/identity_managers/okta/okta_authverifier.py
@@ -10,7 +10,7 @@ from keep.identitymanager.authverifierbase import AuthVerifierBase, oauth2_schem
 logger = logging.getLogger(__name__)
 
 # Define constant locally instead of importing it
-DEFAULT_ROLE_NAME = "user"  # Default role name for user access
+DEFAULT_ROLE_NAME = "noc"  # Default role name for user access
 
 class OktaAuthVerifier(AuthVerifierBase):
     """Handles authentication and authorization for Okta"""
--- a/keep/identitymanager/identity_managers/okta/okta_identitymanager.py
+++ b/keep/identitymanager/identity_managers/okta/okta_identitymanager.py
@@ -114,12 +114,7 @@ class OktaIdentityManager(BaseIdentityManager):
         """Create a role in Okta - disabled"""
         self.logger.info("create_role called but management functions are disabled")
         return ""
-    
-    def get_roles(self) -> list[Role]:
-        """Get all roles from Okta - disabled"""
-        self.logger.info("get_roles called but management functions are disabled")
-        return []
-    
+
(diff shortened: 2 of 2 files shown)


shahargl (approved):
lgtm

PR #6282, opened 2026-04-23
fix: expose is_visible in IncidentDto for workflow conditions

### LINKED ISSUE #6279: [🐛 Bug]: {{ incident.is_visible }} is empty in if condition for workflow.triggers.type: incident
**Describe the bug**
When using `{{ incident.is_visible }}` in an `if` condition within a workflow triggered by `incident.created` or `incident.updated` or `incident.deleted`, the variable evaluates to an empty string - even though the database confirms `is_visible = t`.

**To Reproduce**

1. Create or update a workflow with trigger type incident.
2. Add an action with `if: "{{ incident.is_visible }}"`
3. Trigger the workflow by updating an incident status

**Actual behavior**
The variable resolves to an empty string (''), which evaluates to false, causing the action to be skipped incorrectly.
```
2026-04-22T06:56:10.647000 - Action Default route to Demo-provider evaluated NOT to run, Reason: '' evaluated to false. [before evaluation: '{{ incident.is_visible }}']
```
```
keep=# select is_visible from incident where id = '<UUID>';
 is_visible 
------------
 t
(1 row)
```

**Expected behavior**

`{{ incident.is_visible }}` should not be empty

**Screenshots**

**Additional context**
Currently, we are receiving many unnecessary incidents in the provider - those that are not yet visible (incident.is_visible = false).

The goal is to avoid sending non-visible incidents to external providers.

alpar:
## Problem

`{{ incident.is_visible }}` resolves to an empty string in workflow if-conditions, even though the database has `is_visible = t`. This causes actions to be incorrectly skipped.

Reported in #6279.

## Root Cause

The `is_visible` field exists on the database `Incident` model but was not included in `IncidentDto`. The `from_db_incident()` method maps many fields but skipped `is_visible`, so when the workflow engine renders `{{ incident.is_visible }}`, Mustache finds an empty value.

Notably, the similar boolean flags `is_candidate` and `is_predicted` are already exposed in the DTO — `is_visible` was just missed when it was introduced.

## Fix

Add `is_visible` to `IncidentDto` and its serialization methods, following the exact same pattern as `is_predicted` and `is_candidate`:

1. Add `is_visible: bool = True` field to `IncidentDto`
2. Include `is_visible=db_incident.is_visible` in `from_db_incident()`
3. Include `is_visible=self.is_visible` in `to_db_incident()` for round-trip consistency

## Testing

Added two unit tests:
- `test_incident_dto_is_visible_from_db`: verifies `from_db_incident()` correctly maps `is_visible` for both True and False values
- `test_incident_dto_is_visible_to_db`: verifies `to_db_incident()` preserves `is_visible`

Closes #6279

### DIFF
--- a/keep/api/models/incident.py
+++ b/keep/api/models/incident.py
@@ -76,6 +76,7 @@ class IncidentDto(IncidentDtoIn):
 
     is_predicted: bool
     is_candidate: bool
+    is_visible: bool = True
 
     generated_summary: str | None
     ai_generated_name: str | None
@@ -190,6 +191,7 @@ class IncidentDto(IncidentDtoIn):
             generated_summary=db_incident.generated_summary,
             is_predicted=db_incident.is_predicted,
             is_candidate=db_incident.is_candidate,
+            is_visible=db_incident.is_visible,
             creation_time=db_incident.creation_time,
             start_time=db_incident.start_time,
             last_seen_time=db_incident.last_seen_time,
@@ -245,6 +247,7 @@ class IncidentDto(IncidentDtoIn):
             sources=self.alert_sources,
             is_predicted=self.is_predicted,
             is_candidate=self.is_candidate,
+            is_visible=self.is_visible,
             rule_fingerprint=self.rule_fingerprint,
             fingerprint=self.fingerprint,
             same_incident_in_the_past_id=self.same_incident_in_the_past_id,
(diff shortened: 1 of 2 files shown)


alpar:
Hi @shahargl, CI green — ready for review when you get a chance. Thanks!

shahargl (approved):
Lgtm

PR #6284, opened 2026-04-23
fix: handle None alert_results in _notify_alert when no if_condition

### LINKED ISSUE #6272: [🐛 Bug]:  `_notify_alert` crashes with `NoneType has no len()` when invoked via `enrich_alert` workflow action
## Summary

The `keep` provider's `_notify_alert` method crashes with
`TypeError: object of type 'NoneType' has no len()` whenever a workflow
uses the `enrich_alert` action. The workflow run is marked as error and
the enrichment is never applied.

## Environment

- Keep version: **<paste output of `pip show keep | grep -i version` inside the backend container>**
- Install: self-hosted Docker compose, SQLite
- Auth: `AUTH_TYPE=NO_AUTH`

## Reproducer

1. Create a workflow:
    ```yaml
    workflow:
      id: severity-enrichment
      disabled: false
      triggers:
      - type: alert
        filters:
        - key: source
          value: grafana
      actions:
      - name: enrich-severity
        provider:
          type: keep
          with:
            enrich_alert:
            - key: severity
              value: critical
    ```
2. Send any Grafana-source alert to Keep.
3. Workflow execution fails immediately; severity is not enriched.

## Observed error
Failed to run action enrich-severity: object of type 'NoneType' has no len()

Full stack trace:

File "/venv/lib/python3.13/site-packages/keep/providers/keep_provider/keep_provider.py", line 546, in _notify_alert
"Building alert DTOs", extra={"trigger_count": len(trigger_alerts)}
~~~^^^^^^^^^^^^^^^^
TypeError: object of type 'NoneType' has no len()



A secondary failure then occurs at line 549 (`for alert_result in trigger_alerts:`) with
`'NoneType' object is not iterable`.

## Root cause

In `keep_provider.py` around line 541:

```python
trigger_alerts = alert_results
When the action is purely an enrichment (no new alert triggered), alert_results
is None, which propagates to trigger_alerts. Subsequent len(trigger_alerts)
and iteration both fail.

Expected behavior
The enrichment is applied and the workflow completes successfully, even when
no new alert is produced by the action.

## Workaround

In-place patch (not persistent across container recreation):

```bash
sudo docker exec -u root keep-keep-backend-1 sed -i \
  -e 's/len(trigger_alerts)/len(trigger_alerts or [])/g' \
  -e 's/for alert_result in trigger_alerts:/for alert_result in (trigger_alerts or []):/' \
  /venv/lib/python3.13/site-packages/keep/providers/keep_provider/keep_provider.py
sudo docker restart keep-keep-backend-1
Confirmed working — enrichment applies correctly after patch.

Proposed fix
One-line change at the assignment site:


# before
trigger_alerts = alert_results

# after
trigger_alerts = alert_results or []
This makes the downstream len(trigger_alerts) and iteration work uniformly, no defensive checks needed elsewhere.

Happy to open a PR with this fix if accepted.

### PR DESCRIPTION
**Pull request by alpar**
## Summary

Fixes #6272 — `_notify_alert` crashes with `TypeError: object of type 'NoneType' has no len()` when invoked via `enrich_alert` workflow action.

## Root Cause

In `keep_provider.py` line 544, when no `if_condition` is provided:
```python
trigger_alerts = alert_results
```

For pure enrichment actions (like `enrich_alert`), `alert_results` is `None` because no new alert is triggered. Subsequent `len(trigger_alerts)` and iteration both fail.

## Fix

```python
trigger_alerts = alert_results or []
```

This makes downstream `len()` and iteration work uniformly without defensive checks elsewhere.

## Testing

Added `test_keep_provider_notify_alert_none.py` with two test cases:
1. `test_notify_alert_with_none_alert_results_no_crash` — verifies the crash is fixed
2. `test_notify_alert_with_none_alert_results_and_alert_param` — verifies the alert parameter path still works

## Checklist

- [x] The PR follows the contributing guidelines
- [x] I have added tests that prove my fix is effective
- [x] I have checked that the fix does not break existing functionality

### DIFF
--- a/keep/providers/keep_provider/keep_provider.py
+++ b/keep/providers/keep_provider/keep_provider.py
@@ -538,7 +538,7 @@ class KeepProvider(BaseProvider):
         # if no if_condition, trigger all alerts
         else:
             self.logger.info("No 'if' condition - triggering all alerts")
-            trigger_alerts = alert_results
+            trigger_alerts = alert_results or []
 
         # build the alert dtos
         alert_dtos = []
--- a/tests/test_keep_provider_notify_alert_none.py
+++ b/tests/test_keep_provider_notify_alert_none.py
@@ -1,0 +1,104 @@
+"""
+Test for Keep Provider _notify_alert crash with None alert_results.
+
+This test reproduces the issue described in https://github.com/keephq/keep/issues/6272
+where _notify_alert crashes with TypeError: object of type 'NoneType' has no len()
+when invoked via enrich_alert workflow action (no if_condition, alert_results is None).
+"""
+
+import uuid
+from unittest.mock import MagicMock, patch
+
+from keep.api.core.dependencies import SINGLE_TENANT_UUID
+from keep.contextmanager.contextmanager import ContextManager
+from keep.providers.models.provider_config import ProviderConfig
+from keep.providers.keep_provider.keep_provider import KeepProvider
+
+
+def test_notify_alert_with_none_alert_results_no_crash():
+    """
+    When alert_results is None and no if_condition is provided,
+    _notify_alert should handle it gracefully instead of crashing
+    with TypeError: object of type 'NoneType' has no len().
+
+    This is the enrich_alert scenario: the action is purely enrichment,
+    no new alert is triggered, so alert_results is None.
+    """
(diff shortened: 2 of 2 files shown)


shahargl (approved):
Lgtm

PR #6296, opened 2026-04-25
fix: Database migration race condition in docker with Redis

### LINKED ISSUE #6294: [🐛 Bug]: Race condition on database migration when deploying Keep with Redis
**Describe the bug**
When running Keep with Redis enabled, database migrations are executed during startup by the ARQ Gunicorn instance via keep/api/config.py, which calls migrate_db().

After starting the ARQ workers, the entrypoint script waits a fixed 5 seconds before launching the API Gunicorn workers with SKIP_DB_CREATION=true. This assumes that database migrations will always complete within that time window.

In practice, this assumption is incorrect. Migration duration is variable, and in some cases exceeds 5 seconds. When this happens, the API workers may start and attempt to query the database while migrations are still in progress or incomplete, causing the processes to crash due to missing or partially‑applied schema changes.

This creates a race condition between database migration and API startup, resulting in nondeterministic and flaky failures at startup.

**To Reproduce**

1. Deploy Keep with Redis and an external database like PostgreSQL.
2. Watch the logs of Keep-backend.
3. See that the tenant-table is queried by the API-workers before migration has finished.
4. The container stops with errors.

**Expected behavior**
Database creation and migration should be finished before API workers get to be able to query the database.

**Screenshots**
N/A

**Additional context**
Will PR a fix pulling the DB migration from the workers. This issue is a reference for the PR.

Walkablenormal:
Closes #6294

## 📑 Description

As stated in issue #6294, when deploying in Docker with Redis a race condition is triggered which only allows for a little over 5 seconds for database migration until the API workers start, and try to query the database. In this PR, database migration is moved up and out of the workers. Now it will run before the workers start. The migrate_db() function handles if the user has set SKIP_DB_CREATION to true.

## ✅ Checks

- [N] My code requires changes to the documentation
- [X] I have updated the documentation as required

### DIFF
--- a/keep/api/config.py
+++ b/keep/api/config.py
@@ -6,7 +6,7 @@ from keep.api.alert_deduplicator.deduplication_rules_provisioning import (
     provision_deduplication_rules_from_env,
 )
 from keep.api.api import AUTH_TYPE
-from keep.api.core.db_on_start import migrate_db, try_create_single_tenant
+from keep.api.core.db_on_start import try_create_single_tenant
 from keep.api.core.dependencies import SINGLE_TENANT_UUID
 from keep.api.core.tenant_configuration import TenantConfiguration
 from keep.api.routes.dashboard import provision_dashboards
--- a/keep/api/config.py
+++ b/keep/api/config.py
@@ -44,8 +44,6 @@ def on_starting(server=None):
     """This function is called by the gunicorn server when it starts"""
     logger.info("Keep server starting")
 
-    migrate_db()
-
     # Load this early and use preloading
     # https://www.joelsleppy.com/blog/gunicorn-application-preloading/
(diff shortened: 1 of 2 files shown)


Walkablenormal:
Hi,

Even with the failing tests, this seems to be safe to merge. The E2E test and unit test failing seems to be unrelated to this PR.

I'll take a look if I can fix these tests in a separate PR.

EDIT: Fixed unit test in #6298, if you want, you can/may merge that first and check if this PR correctly runs the unit tests.

@shahargl Can you look at this and merge please?

shahargl (changes requested):
I don't want to change default behaviour just due to this use case, you should either fix the problem itself (wait as long as needed) or add some env var to enable this behavior

PR #6392, opened 2026-05-02
fix: show zero-config providers in Connect Provider catalog

### LINKED ISSUE #6273: [🐛 Bug]: HTTP / bash / python / console providers hidden from Connect Provider UI catalog
## Summary

The `http` provider (and similarly `bash`, `python`, `console`) does not appear in the **Connect Provider** UI catalog, even though the provider code is present and functional on the backend. This makes it impossible to install these providers through the UI, blocking common use cases (workflow actions that call arbitrary HTTP endpoints, run scripts, etc.).

## Environment

- Keep build: commit `339151c9` on `main` (released 2026-04-14 as part of post-v0.51.0 development)
- Deployment: self-hosted docker compose, images `us-central1-docker.pkg.dev/keephq/keep/keep-api:latest` and `keep-ui:latest`
- `EE/` directory on the image only contains `identitymanager` — no EE provider gating

## Reproduction

1. Open Keep UI → **Providers** → **Connect Provider**
2. Search `http` in the catalog → **no results**
3. Repeat with `bash`, `python`, `console` → **no results**
4. On the backend container:
   ```bash
   docker exec keep-keep-backend-1 ls /venv/lib/python3.13/site-packages/keep/providers/ | grep -E 'http|bash|python|console'
   ```
   → `http_provider/`, `bash_provider/`, `python_provider/`, `console_provider/` are all present.

## Root cause

Frontend filter in [`keep-ui/app/(keep)/providers/page.client.tsx`](https://github.com/keephq/keep/blob/main/keep-ui/app/%28keep%29/providers/page.client.tsx) requires:

```ts
Object.keys(provider.config || {}).length > 0 || (provider.tags && provider.tags.includes("alert"))
```

The `http` provider:
- has no auth config (it's stateless — URL/method/body/headers are per-call workflow args)
- has no `PROVIDER_TAGS` declaration in [`http_provider.py`](https://github.com/keephq/keep/blob/main/keep/providers/http_provider/http_provider.py)
- backend tag-derivation logic in [`providers_factory.py`](https://github.com/keephq/keep/blob/main/keep/providers/providers_factory.py) (lines ~385–400) gives it `data` + `messaging` tags based on `_query` / `_notify` methods, but not `alert` (which requires `_get_alerts`)

→ Filter evaluates `false || false` → hidden.

Same applies to `bash`, `python`, `console` which also have no auth config and no alert-source behavior.

## Impact

Workflows that need to call REST APIs with dynamic URLs (e.g., auto-restart an infra resource based on alert labels) cannot install the provider via UI. Operators either abandon the workflow, use the wrong provider (`webhook` has a fixed URL), or discover the `KEEP_PROVIDERS` env var workaround by reading source code.

Even worse: the UI silently hides the provider with no indication that it exists. Search just returns empty results — no hint that you should use a different mechanism.

## Workaround

Provision via `KEEP_PROVIDERS` env var on `keep-backend`:

```bash
# docker-compose.yml keep-backend service:
environment:
  - KEEP_PROVIDERS=${KEEP_PROVIDERS}

# .env:
KEEP_PROVIDERS={"http-default":{"type":"http","authentication":{}}}
```

Then `docker compose up -d --force-recreate keep-backend`. Provider appears under **Installed Providers**.

## Suggested fix

Pick one:

**A. Add minimal `PROVIDER_TAGS` to the affected providers** — e.g., `PROVIDER_TAGS = ["action"]` (new tag) or `PROVIDER_TAGS = ["alert"]` (reuse existing) on `http_provider`, `bash_provider`, `python_provider`, `console_provider`. Minimal code change, doesn't touch the filter.

**B. Update the UI filter to include action-capable providers** — `Object.keys(config).length > 0 || tags.includes("alert") || (can_notify && can_query)`. Broader fix, catches any similar provider in the future.

**C. Add a "Utility" or "Action" tab** to the Connect Provider UI that exposes these stateless power-user providers explicitly, separate from the alert-source tabs.

Happy to open a PR for any of these if there's a preferred direction.

## Related

Another Keep asymmetry discovered same day: the `snowflake` provider is installed by default (Database category, `_query` for enrichment) but has no `_format_alert`, so `POST /alerts/event/snowflake` explodes with `NotImplementedError` while `POST /alerts/event/<unknown-name>` falls back gracefully to the `keep` provider. Users hit this landmine when routing Snowflake-originated alerts through Keep. Distinct issue but same root cause family: asymmetric provider gating.

aayushbaluni:
## Problem
HTTP, bash, python, and console providers are hidden from the Connect Provider UI because the catalog filter excludes providers with empty auth config and no `alert` tag.

## Change
Add shared `isConnectCatalogVisible()` and include providers that expose both `can_notify` and `can_query` (stateless workflow/action providers), in addition to providers with auth fields or alert ingestion.

Fixes #6273.

### DIFF
--- a/keep-ui/app/(keep)/providers/page.client.tsx
+++ b/keep-ui/app/(keep)/providers/page.client.tsx
@@ -1,5 +1,9 @@
 "use client";
-import { defaultProvider, Provider } from "@/shared/api/providers";
+import {
+  defaultProvider,
+  Provider,
+  isConnectCatalogVisible,
+} from "@/shared/api/providers";
 import ProvidersTiles from "./providers-tiles";
 import React, { useState, useEffect } from "react";
 import Loading from "@/app/(keep)/loading";
@@ -196,11 +200,8 @@ export default function ProvidersPage({
       searchCategories(provider)
   );
 
-  const displayableProviders = filteredProviders.filter(
-    (provider) =>
-      Object.keys(provider.config || {}).length > 0 ||
-      (provider.tags && provider.tags.includes("alert"))
-  );
+  const displayableProviders =
+    filteredProviders.filter(isConnectCatalogVisible);
 
   return (
     <>
--- a/keep-ui/app/(keep)/providers/providers-tiles.tsx
+++ b/keep-ui/app/(keep)/providers/providers-tiles.tsx
@@ -1,6 +1,6 @@
 "use client";
 import { Title } from "@tremor/react";
-import { Providers, Provider } from "@/shared/api/providers";
+import { Providers, Provider, isConnectCatalogVisible } from "@/shared/api/providers";
 import { useEffect, useState } from "react";
 import ProviderForm from "./provider-form";
 import ProviderTile from "./provider-tile";
@@ -85,11 +85,7 @@ const ProvidersTiles = ({
   };
 
   const sortedProviders = providers
-    .filter(
-      (provider) =>
-        Object.keys(provider.config || {}).length > 0 ||
-        (provider.tags && provider.tags.includes("alert"))
-    )
+    .filter(isConnectCatalogVisible)
     .sort(
       (a, b) =>
         // Put coming_soon providers at the end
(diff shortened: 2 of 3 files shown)


shahargl:
it is intended behaviour not to show them

Issue #3526, opened 2025-02-17
[🔌 Provider]: Solarwinds


Matvey-Kuk:
We're looking for a https://www.solarwinds.com/ provider for Keep.

Matvey-Kuk:
Hi @Ayush9026 ! Do you have experience with Solarwinds?

ezhil56x:
In queue, if reassigned

Matvey-Kuk:
@Ayush9026 thank you! Assigned to you.

ezhil56x:
@Ayush9026 Are you still working on this?

talboren:
@Ayush9026 since there's no activity for quite a while, I'm assigning @ezhil56x

anashakim:
/attempt #3526 I have experience with solarwinds, can I work on this?

talboren:
> /attempt [#3526](https://github.com/keephq/keep/issues/3526) I have experience with solarwinds, can I work on this?
> 
> Options

Seems like @ezhil56x is already on it

ezhil56x:
@anashakim I'm dropping this, couldn't make it. Feel free to take it!

Matvey-Kuk:
@anashakim assigned, thank you!

DMMutua:
hello @Matvey-Kuk I'd like to be re-assigned to this and /attempt #3526 since there's been inactivity for quite a long time. Will raise PR ASAP.

Matvey-Kuk:
Bringing @shahargl to the thread

shahargl:
@Matvey-Kuk thanks.

@DMMutua i can assign you, but currently we halted the bounty program so it’s only if you are up for a classic oss contribution

DMMutua:
@shahargl I'd like to work on it either way.

shahargl:
assigned @DMMutua

MekonMAC:
I'm interested in implementing the Solarwinds provider as an automated alternative. I have a robust implementation plan ready with tests. Please let me know if I can submit a PR for this.

tianzhuxhc-bit:
Hi maintainers, I'd like to work on this bounty.

Plan:
1) Reproduce and align scope quickly
2) Submit an initial PR with tests/docs
3) Iterate fast based on review feedback

I can post first progress within 12 hours.

brutusworker-arch:
I have implemented the SolarWinds provider and opened a PR to resolve this issue.

PR: https://github.com/keephq/keep/pull/6108

**What is implemented:**
[...]

PR #6404, opened 2026-05-05
feat(providers): add SolarWinds provider for pulling Orion active alerts

ayoubil:
## Summary

Adds a pull-based provider that queries SolarWinds Orion via the SolarWinds Information Service (SWIS) REST API and surfaces active alerts in Keep. Modeled after the existing Centreon provider (`keep/providers/centreon_provider/`) — small, self-contained, no extra runtime dependencies.

## Changes

- `keep/providers/solarwinds_provider/__init__.py` — package marker.
- `keep/providers/solarwinds_provider/solarwinds_provider.py` — provider class + auth dataclass (~240 LOC).
- `keep/providers/solarwinds_provider/alerts_mock.py` — four mock SWIS query payloads (node down, interface warning, acknowledged volume, informational discovery).
- `keep/providers/solarwinds_provider/README.md` — setup, severity mapping, SWIS reference.

## Behaviour

- **Auth**: HTTP Basic against the SWIS endpoint on the Orion server (port `17778` by default), using the same credentials as the Orion Web Console. `verify_ssl` flag for self-signed certs.
- **Query**: a single SWQL join across `Orion.AlertActive`, `Orion.AlertObjects` and `Orion.AlertConfigurations` so each row already carries the configured alert name, description, severity and the entity that triggered it — no follow-up per-alert calls needed.
- **Severity** mapping: Orion's numeric `0..4` (Informational → Critical) → Keep's `info / low / warning / high / critical`.
- **Status**: alerts that have been acknowledged in Orion become `AlertStatus.ACKNOWLEDGED`, otherwise `FIRING`.
- **Stable id**: `AlertObjectID` (unique per alert-config × entity in Orion), so the same active alert is deduped across pulls rather than being re-emitted as new on every cycle.
- **Read-only**: the provider only issues `SELECT` queries against `Orion.*` views; a read-only Orion account is sufficient.

## How to verify

1. Configure the provider with a SolarWinds Orion server URL, username, and password.
2. Trigger an alert in Orion (e.g. shut down a polled interface) and wait one polling cycle.
3. The active alert should appear in Keep with the correct severity, status, and entity caption. Acknowledging the alert in Orion should change its Keep status to `acknowledged` on the next pull.

The mock payloads in `alerts_mock.py` mirror the SWIS `Query` response shape and can be used to unit-test `_get_alerts` without a live Orion instance.

Closes #3526

### DIFF
--- a/keep/providers/solarwinds_provider/README.md
+++ b/keep/providers/solarwinds_provider/README.md
@@ -1,0 +1,66 @@
+# SolarWinds Provider
+
+Pulls active alerts from [SolarWinds Orion](https://www.solarwinds.com/) via the
+SolarWinds Information Service (SWIS) REST API.
+
+## Authentication
+
+The provider authenticates against SWIS with HTTP Basic auth — the same
+username and password you would use to log into the Orion Web Console. SWIS
+listens on port `17778` by default.
+
+| Field        | Required | Description                                                                 |
+|--------------|----------|-----------------------------------------------------------------------------|
+| `host_url`   | yes      | Base URL of the Orion server, e.g. `https://orion.example.com:17778`.       |
+| `username`   | yes      | Orion account username with read access to alerts.                           |
+| `password`   | yes      | Orion account password.                                                      |
+| `verify_ssl` | no       | Set to `false` if your Orion server uses a self-signed certificate.          |
+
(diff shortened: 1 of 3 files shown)


shahargl:
N/A

PR #6469, opened 2026-05-12
feat: add SNMP provider for Keep

CuboYe:
/claim #2112

Implement SNMP provider to receive SNMP traps/events via webhook.
- Follows existing provider pattern (icinga2, etc.)
- SnmpProvider class with pydantic auth config
- STATUS_MAP and SEVERITY_MAP for SNMP generic trap types
- _format_alert for webhook payload parsing
- _get_alerts for polling SNMP manager API
- Webhook installation support

Closes #2112


shahargl:
N/A

PR #6472, opened 2026-05-13
feat: SNMP trap provider (v1/v2c/v3 with USM auth/priv)

ykd007:
Closes #2112\n\nBeen meaning to get SNMP trap ingestion working for a while. This adds a new provider that listens for traps over UDP and turns them into Keep alerts — covering v1, v2c, and v3 with USM auth/priv.\n\n**What's in here:**\n\n- UDP trap listener that runs as a background thread, feeds a bounded queue\n- SNMPv1 and v2c: community string filtering, enterprise OID and varbind extraction\n- SNMPv3: USM with SHA/MD5 auth and AES/AES256/DES encryption config\n- Well-known trap OIDs mapped to human names + severity (linkDown, linkUp, coldStart, authFailure, etc.)\n- Raw BER decoder as fallback if pysnmp isn't available\n- Mock alerts for UI preview\n\nTested manually with  against a local listener. The v3 path requires pysnmp; if it's not installed the BER fallback handles v1/v2c cleanly.\n\nPart of the Algora bounty on this issue.

### DIFF
--- a/keep/providers/snmp_provider/alerts_mock.py
+++ b/keep/providers/snmp_provider/alerts_mock.py
@@ -1,0 +1,63 @@
+"""Mock SNMP trap alerts for Keep UI preview."""
+
+import uuid
+from datetime import datetime, timezone
+
+ALERTS_MOCK = [
+    {
+        "id": str(uuid.uuid4()),
+        "name": "linkDown",
+        "severity": "high",
+        "status": "firing",
+        "source": ["snmp"],
+        "message": "A network link has gone down",
+        "description": "enterprise=1.3.6.1.6.3.1.1.5.3; ifIndex=2",
+        "pushed": True,
+        "pdu_type": "SNMPv2-Trap",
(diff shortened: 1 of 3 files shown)


ykd007:
Pushed a fix commit addressing two correctness issues:

1. **BER v2c varbind parsing was a stub** — was returning a placeholder string instead of actually walking the VarBindList. Now fully decodes each OID/value pair, handles OCTET STRING (UTF-8), INTEGER, and raw hex for other types.

2. **SNMPv3 decode path was broken** — `protoModules[protoVersion3]` is not a valid key in pysnmp; the v3 path always fell through to BER. Fixed to use pyasn1 BER decoder with `rfc3412.Message` and `rfc3414.ScopedPDU` to extract varbinds directly from the scoped PDU.

Both the pysnmp and BER fallback paths now correctly return populated AlertDtos.

ykd007:

shahargl:
N/A

PR #6477, opened 2026-05-13
feat: add nagios provider

canburakyol:
Added Nagios provider supporting both Nagios XI API and Nagios Core webhooks. Closes #3960


shahargl:
N/A

PR #6482, opened 2026-05-14
fix(prometheus): add Thanos Querier compatibility mode

### LINKED ISSUE #6474: [➕ Feature]: Add thanos compatibility
It's not realy a problem, but the provider prometheus is not compatible with thanos.
When I setup thanos url I have this issue : 
```
prometheus scopes are invalid: { "connectivity": "'NoneType' object is not iterable" }
```
Thanos does not support api/v1/alerts.

It would be great if provider prometheus was compatible both with prometheus and thanos (same engine search).

fix provider prometheus to work with thanos

Develop new provider "thanos"

ykd007:
## Problem

Connecting Keep's Prometheus provider to a Thanos Querier fails with:

```
prometheus scopes are invalid: { "connectivity": "'NoneType' object is not iterable" }
```

This happens because Thanos Querier does not implement `/api/v1/alerts` (that endpoint only exists on Prometheus's built-in rules evaluator / Alertmanager). Keep's connectivity check and alert polling both hit that endpoint unconditionally.

Closes #6474

## Solution

Adds a `thanos_compatible: bool` flag (default `False`) to `PrometheusProviderAuthConfig`. When enabled:

- **Connectivity check** uses `/api/v1/rules?type=alert` instead of `/api/v1/alerts`
- **Alert polling** uses `/api/v1/rules?type=alert` which Thanos supports and which returns full alert data including annotations, state (`firing`/`pending`), `activeAt`, and labels
- Fingerprints are computed as MD5 of sorted label JSON (same approach used elsewhere in the codebase)
- Fully backward compatible — existing Prometheus configs with `thanos_compatible: false` (default) are unchanged

## Testing

Set `thanos_compatible: true` in provider auth config when connecting to a Thanos Querier URL.

### DIFF
--- a/docs/snippets/providers/anthropic-snippet-autogenerated.mdx
+++ b/docs/snippets/providers/anthropic-snippet-autogenerated.mdx
@@ -1,9 +1,11 @@
-{/* This snippet is automatically generated using scripts/docs_render_provider_snippets.py 
+{/* This snippet is automatically generated using scripts/docs_render_provider_snippets.py
 Do not edit it manually, as it will be overwritten */}
 
 ## Authentication
 This provider requires authentication.
 - **api_key**: Anthropic API Key (required: True, sensitive: True)
+- **model**: Claude model to use (required: False, sensitive: False)
+- **system_prompt**: System prompt that sets Claude's role for all requests in this provider. (required: False, sensitive: False)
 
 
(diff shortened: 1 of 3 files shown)


shahargl (dismissed):
lgtm

Issue #6426, opened 2026-05-06
[🔌 Provider]: LINE Notify notification provider


ayoubil:
## Summary

Add LINE Notify provider to Keep.

## Description

LINE Notify is LINE messaging notification API. Add a provider to send Keep alerts via LINE Notify, widely used in Japan and Southeast Asia.

API: https://notify-bot.line.me/doc/en/

## Provider Type

Notification (outbound)

shahargl:
N/A

PR #6504, opened 2026-05-16
feat: add LINE Notify notification provider (#6426)

qiuluoli:
Implements #6426: LINE Notify notification provider.

LINE has 200M+ users especially in Japan/Thailand/Taiwan. Simple POST API with access token.
- Text messages with 1000 char limit
- Optional sticker support
- LINE Notify API: https://notify-api.line.me/api/notify

/attempt #6426


shahargl:
N/A

PR #6508, opened 2026-05-18
fix: Handle test Contact Point alert from Grafana 12

### LINKED ISSUE #6506: [🐛 Bug]: Grafana 12 Contact Point test alert fails to be processed
### **Describe the bug**
After upgrading to Grafana 12, sending a test Contact Point to Keep fails with "Error processing event, contact Keep team for more information" and no alert is ingested.

**To Reproduce**

1. In Grafana 12, go to Alerting -> Contact points and open the Keep webhook.
2. Click "Test" -> "Send test notification".
3. Check Keep: no alert appears. Backend logs show a pydantic URL host invalid validation error from grafana_provider._format_alert.

**Expected behavior**
The test alert should be ingested and visible in Keep with all labels, annotations, and dashboard/panel/silence URLs preserved, just like with Grafana 11.

**Additional context**
Grafana 12 test alerts emit generatorURL as a query string only (e.g. "?orgId=1"), which Keep's AlertDto.url validator can't accept because it has no host, so the whole payload is dropped. Real (non-test) Grafana 12 alerts that carry an absolute generatorURL are unaffected.

piekarski-mateusz:
Closes #6506 

## 📑 Description
Fixes Grafana 12 test Contact Point payloads being rejected by Keep. Grafana 12 emits generatorURL as a bare query string (e.g. "?orgId=1"), which AlertDto.url can't validate because it has no host, so the whole alert was dropped. The Grafana provider now resolves relative generatorURL/imageURL against the payload's externalURL before constructing the AlertDto.

## ✅ Checks

## ℹ Additional Information
No breaking changes, no new dependencies. 
Real (non-test) Grafana alerts with absolute generatorURLs are unaffected.
Tested locally by POSTing the exact Grafana 12 test payload to /alerts/event/grafana: without the fix the alert is dropped with a pydantic URL host invalid error, with the fix the alert is parsed and url resolves to https://<externalURL host>/?orgId=1.

### DIFF
--- a/keep/providers/grafana_provider/grafana_provider.py
+++ b/keep/providers/grafana_provider/grafana_provider.py
@@ -9,6 +9,7 @@ import json
 import logging
 import re
 import time
+import urllib.parse
 
 import pydantic
 import requests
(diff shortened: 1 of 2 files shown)


shahargl (approved):
lgtm

PR #6512, opened 2026-05-19
fix: set is_predicted = true for topology-based incidents

### LINKED ISSUE #6511: [🐛 Bug]: Missing explicit `is_predicted=True` when creating topology based incident
**Describe the bug**

In [keep/topologies/topology_processor.py](https://github.com/keephq/keep/blob/main/keep/topologies/topology_processor.py#L352-L360), when creating an Incident with `incident_type="topology"` , the `is_predicted` flag is not set. In the ORM model (keep/api/models/db/incident.py), this field defaults to `False`. 

Because of this, topology incidents are incorrectly marked as unpredicted, even though they are, in fact, the result of correlation analysis based on service topology.

**To Reproduce**

1. Open keep/topologies/topology_processor.py.  
2. Check the Incident constructor call in` _create_application_based_incident`.  
3. Observe that `is_predicted` is absent, while `is_visible=True` and `is_candidate=False` are present.  
4. Check `keep/api/models/db/incident.py` — `is_predicted` defaults to `False`.

**Expected behavior**
When creating a topological incident, `is_predicted=True` must be explicitly specified, since the incident is generated algorithmically based on the topological dependency of services, and not by the direct triggering of a custom alert or rule.

**Screenshots**
-

**Additional context**

Version: 0.51.0 | Build: 0cfb50

PrometheRus:
Closes #6511

## 📑 Description
In [keep/topologies/topology_processor.py](https://github.com/keephq/keep/blob/main/keep/topologies/topology_processor.py#L352-L360), when creating an Incident with `incident_type="topology"` , the `is_predicted` flag is not set. In the ORM model (keep/api/models/db/incident.py), this field defaults to `False`.

When creating a topological incident, `is_predicted=True` must be explicitly specified, since the incident is generated algorithmically based on the topological dependency of services, and not by the direct triggering of a custom alert or rule.

## Fix
The fix would be straightforward — adding `is_predicted=True` to the Incident constructor in `topology_processor.py`:
```
incident = Incident(
    tenant_id=tenant_id,
    user_generated_name=f"Application incident: {application.name}",
    user_summary=f"Multiple services in application {application.name} are experiencing issues",
    incident_type="topology",
    incident_application=application.id,
    is_candidate=False,
    is_visible=True,
    is_predicted=True,  # Topology-based incidents are algorithmically generated
)
```

This aligns topology incidents with the existing convention for system-generated incidents.

## Testing

Added unit test:

*   `test_create_application_based_incident_sets_is_predicted`: verifies that `_create_application_based_incident()` correctly sets `is_predicted=True` for topology incidents, and validates that `is_visible=True`, `is_candidate=False`, and `incident_type="topology"` are properly assigned

### DIFF
--- a/keep/topologies/topology_processor.py
+++ b/keep/topologies/topology_processor.py
@@ -357,6 +357,7 @@ class TopologyProcessor:
                 incident_application=application.id,
                 is_candidate=False,  # Topology-based incidents are always confirmed
                 is_visible=True,  # Topology-based incidents are always confirmed
+                is_predicted=True,  # Topology-based incidents are algorithmically generated
             )
 
             # Persist incident to database before adding alerts
--- a/tests/test_topology.py
+++ b/tests/test_topology.py
@@ -17,8 +17,9 @@ from keep.topologies.topologies_service import (
     InvalidApplicationDataException,
     ServiceNotFoundException,
 )
+from keep.api.models.db.alert import Incident
+from keep.topologies.topology_processor import TopologyProcessor
 from tests.fixtures.client import setup_api_key, client, test_app  # noqa: F401
-
 
 VALID_API_KEY = "valid_api_key"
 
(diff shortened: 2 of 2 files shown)


shahargl (approved):
lgtm

PR #6526, opened 2026-05-25
fix: parse site24x7 LABELS as dict for AlertDto compatibility

### LINKED ISSUE #6527: [🐛 Bug]: site24x7 LABELS still not passed correctly to AlertDto after #6196
## Bug Description

Issue #6196 was previously reported and closed with PRs #6210 and #6201, but the labels 
are still not appearing on alerts in Keep.

## Root Cause

The fix in the previous PRs added extraction of `LABELS` from the webhook payload, but 
parsed it as a `list` of strings. `AlertDto.labels` is typed as `dict | None`, so passing 
a list causes pydantic to silently drop the value — resulting in `"labels": {}` on every alert.

## Fix

Parse the `LABELS` string (e.g. `"env:prod,team:sre"`) into a dict (`{"env": "prod", "team": "sre"}`).

This is addressed in PR #6526.

Jleee713:
Closes #6527 

## 📑 Description

Fix `Site24X7Provider._format_alert` incorrectly parsing the `LABELS` webhook field as a `list` instead of a `dict`.

`AlertDto.labels` is typed as `dict | None`, so passing a list caused labels to be silently dropped by pydantic on alert ingestion.

- [x] Parse `LABELS` string (e.g. `"env:prod,team:sre"`) into a `dict` (`{"env": "prod", "team": "sre"}`)
- [x] Handle pre-parsed `dict` payloads passthrough
- [x] Fall back to `{}` for empty/missing values

## ✅ Checks

## ℹ Additional Information

**Before:** `LABELS` was split into a list of strings and passed to `AlertDto.labels`, which expects a `dict` — causing labels to never appear on alerts.

**After:** `LABELS` is parsed as `key:value` pairs into a dict. Keys without a `:` separator are stored with an empty string value.

### DIFF
--- a/keep/providers/site24x7_provider/site24x7_provider.py
+++ b/keep/providers/site24x7_provider/site24x7_provider.py
@@ -234,11 +234,20 @@ class Site24X7Provider(BaseProvider):
         elif not isinstance(tags, list):
             tags = []
 
-        labels = event.get("LABELS", "")
-        if isinstance(labels, str) and labels:
-            labels = [lbl.strip() for lbl in labels.split(",") if lbl.strip()]
-        elif not isinstance(labels, list):
-            labels = []
+        labels_raw = event.get("LABELS", "")
+        if isinstance(labels_raw, dict):
+            labels = labels_raw
+        elif isinstance(labels_raw, str) and labels_raw:
+            labels = {}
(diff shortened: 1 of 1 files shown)


shahargl (approved):
lgtm

PR #6540, opened 2026-05-27
fix: resolve "window is not defined" SSR crash on direct URL access to alerts page

### LINKED ISSUE #6543: [🐛 Bug]: Alert details permanent link redirects to feed
**Describe the bug**
When user opens a browser and directly types the permanent link of an alert it shows up temporary the alert but automatically forward again to the feed page.

**To Reproduce**

1. Open browser
2. write the url: https://<host>/alerts/feed?alertPayloadFingerprint=fd2efbf28935f0b6
3. Shows the alert details and sudently disappears to show only the alerts feed 

**Expected behavior**
It's expected to have a direct url to alerts, so they can be referenced from external systems

ausias-armesto:
**Problem**
Closes #6543
Directly loading any alerts URL (e.g. /alerts/feed?alertPayloadFingerprint=…) caused a server-side crash:

```
ReferenceError: window is not defined
  at .next/server/chunks/4585.js
  at .next/server/app/(keep)/rules/page.js
 ```
 
The root cause was monaco-editor being imported at module level in two files. monaco-editor is a browser-only library that accesses window during its own module initialisation. In Next.js App Router, "use client" components are still pre-rendered on the server (SSR), so any module-level import that touches window crashes the Node.js runtime before the page can render.

**Changes**
- Removed the top-level `import * as monaco from "monaco-editor"` and the module-level `loader.config({ monaco })` call.
- Moved both into a `useEffect` as a dynamic `import("monaco-editor")`, so they only ever execute in the browser.
shared/ui/MonacoCELEditor/monaco-cel-editor.tsx
- Changed `import { editor, Token } from "monaco-editor"` → `import type { editor, Token } from "monaco-editor"`.
- Replaced `window.location.search` and `window.location.pathname` inside `resetUrlAfterModal` with Next.js's `usePathname()` and `useSearchParams()` hooks.
- Adds `usePathname` to the import and wires it into the useCallback dependency array.
- Implement a full permanent link approach by adding a new route `/alerts/fingerprint/<fp>` page.

### DIFF
--- a/keep-ui/app/(keep)/alerts/[id]/ui/__tests__/alerts-fingerprint.test.tsx
+++ b/keep-ui/app/(keep)/alerts/[id]/ui/__tests__/alerts-fingerprint.test.tsx
@@ -24,6 +24,7 @@ import Alerts from "../alerts";
 jest.mock("next/navigation", () => ({
   useRouter: jest.fn(),
   useSearchParams: jest.fn(),
+  usePathname: jest.fn().mockReturnValue("/alerts/feed"),
 }));
 
 // ─── Mock data hooks ─────────────────────────────────────────────────────────
--- a/keep-ui/app/(keep)/alerts/[id]/ui/alerts.tsx
+++ b/keep-ui/app/(keep)/alerts/[id]/ui/alerts.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
-import { useRouter, useSearchParams } from "next/navigation";
+import { usePathname, useRouter, useSearchParams } from "next/navigation";
 import { type AlertDto, type AlertsQuery } from "@/entities/alerts/model";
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Review by shahargl — CHANGES_REQUESTED** [review]
I can’t merge without linking this to some ticket c can you open a bug and link it?

**Comment by shahargl** [comment]
@ausias-armesto - frontend tests are failng
```
  ● Alerts — fingerprint modal fix (resolvedFingerprintRef) › resets the ref and opens modal correctly when navigating to a different fingerprint

    TypeError: (0 , _navigation.usePathname) is not a function

      62 |   );
      63 |
    > 64 |   const pathname = usePathname();
         |                               ^
      65 |   const searchParams = useSearchParams();
      66 |   // hooks for the note and ticket modals
      67 |   const [noteModalAlert, setNoteModalAlert] = useState<AlertDto | null>();
```

ausias-armesto:
@shahargl Seems that the frontend tests are passing now

PR #6541, opened 2026-05-28
fix: Simpler approach to load all incident params into labels

### LINKED ISSUE #6542: [🐛 Bug]: Site24x7 incident parameters not automatically loading in labels
Describe the bug
When a Site24x7 monitor triggers a DOWN/TROUBLE/CRITICAL alert, the labels field on the resulting AlertDto is always empty, even though Site24x7 sends incident parameters ($STATUS, $MONITORTYPE, $INCIDENT_DETAILS, $FAILED_LOCATIONS, etc.) in the webhook payload.

To Reproduce

Configure a Site24x7 provider with a webhook pointing to Keep
Trigger a monitor alert (e.g. take a monitored URL down)
Inspect the received alert in Keep — labels is {}
Expected behavior
All incident parameters sent by Site24x7 in the webhook payload (e.g. MONITORTYPE, INCIDENT_DETAILS, FAILED_LOCATIONS, MONITOR_GROUPNAME, OUTAGE_TIME_UNIX_FORMAT) should be populated in the alert's labels field.

Root cause
_format_alert looked for a LABELS key via event.get("LABELS", ""). Site24x7 never sends a LABELS field — incident parameters are top-level keys in the JSON body, so the lookup always returns an empty string and labels is never populated.

Additional context
Fixed in PR #6541  by replacing the dead key lookup with {k: str(v) for k, v in event.items() if v is not None}.

Jleee713:
📑 Description
closes #6542

Fix _format_alert in the Site24x7 provider so that webhook incident parameters are correctly surfaced as alert labels.

Previously the code looked for a LABELS key in the webhook payload (event.get("LABELS", "")), which Site24x7 never sends. Incident parameters (STATUS, MONITORTYPE, INCIDENT_DETAILS, FAILED_LOCATIONS, etc.) are delivered as top-level fields in the JSON body when send_incident_parameters=True, so labels was always empty.

Changes:

_parse_labels helper extracts labels from the webhook payload via two paths:

Default: all top-level incident parameters are captured as labels — works automatically with send_incident_parameters=True (set by both setup_webhook and manual webhook configuration)
Backward compat: if a user previously configured a custom LABELS parameter in Site24x7's webhook UI (key:value,key:value format), that string is still parsed correctly
Extracted into a separate _parse_labels method to keep _format_alert within cognitive complexity limits

 Labels now populated from Site24x7 incident parameters on monitor DOWN/TROUBLE/CRITICAL/UP events

 Backward compatible with users who configured a custom LABELS parameter

 Existing AlertDto fields (name, url, description, etc.) unchanged

✅ Checks
 My pull request adheres to the code style of this project
 My code requires changes to the documentation
 I have updated the documentation as required
 All the tests have passed
ℹ Additional Information
Before (labels always empty):

labels_raw = event.get("LABELS", "")  # Site24x7 never sends this key
...
labels = {}
After (labels populated from incident parameters, with backward compat):

labels_raw = event.get("LABELS")
if labels_raw is None:
    # Default: incident params arrive as top-level fields
    labels = {k: str(v) for k, v in event.items() if v is not None}
else:
    # Backward compat: custom LABELS parameter configured in Site24x7 UI
    labels = parse_key_value_string(labels_raw)


shahargl (changes requested):
For backward compatibility, keep the original logic and just add the latter (instead of removing), unless you have a very good reason why the old logic is not needed

Jleee713:
The original logic only works if user configures custom labels parameters in Site24x7 for each monitor. The only use-case for that is if user does not want all incident parameters in webhook, however we can just pass all incident parameters with option to filter on keephq side. 

Either way, this automatically populates the incident parameters in labels without repeated configuration from site24x7.

shahargl:
@Jleee713 but isn't  Site24x7 is doing that with `setup_webhook`, so whoever installs it via setup webhook won't have labels now?

Jleee713:
Hey @shahargl, good catch — **setup_webhook** sets **send_incident_parameters: True** which delivers all incident params as top-level JSON fields automatically, so both manual and auto webhook setup paths work correctly.

Updated the implementation with two paths:

- Default: captures all top-level incident parameters (STATUS, MONITORTYPE, FAILED_LOCATIONS, etc.) as labels — works out of the box with send_incident_parameters: True

- Backward compat: if a user previously configured a custom LABELS parameter in Site24x7's webhook UI, that key:value,key:value string is still parsed correctly

Also extracted the logic into _parse_labels to keep _format_alert within complexity limits.

shahargl (approved):
lgtm

PR #6548, opened 2026-05-28
fix: deserialize JSON string body in HTTP provider before sending request

### LINKED ISSUE #6547: [🐛 Bug]: HTTP provider sends double-encoded JSON when workflow body uses {{ alert }} template
**Describe the bug**
The HTTP provider is not handling as JSON the content pass on the body content

**To Reproduce**
When a workflow action uses the HTTP provider with a template like:
```
    - name: http-request
      provider:
        type: http
        with:
          url: "http://host:port"
          method: POST
          headers:
            Content-Type: "application/json"
          body: "{{ alert }}"

```
, Keep's template engine renders {{ alert }} into a JSON string (e.g. '{"id": "123", ...}'). That string is then passed to requests.post(json=body), which serialises it again — sending a JSON-encoded string ("\"...\"") to the target endpoint instead of a JSON object ({...}). The receiving endpoint gets the wrong content type and payload shape.

ausias-armesto:
Close #6547 

When a workflow renders body: "{{ alert }}", the template produces a JSON string instead of a dict. Passing that string to requests.post(json=body) double-encodes it, sending a JSON string to the target instead of a JSON object.

The fix mirrors the existing handling already in place for headers: if body arrives as a string and is valid JSON, it is deserialized back to a dict before being handed to requests. Non-JSON strings are left unchanged for backwards compatibility.

Adds unit tests for the HTTP provider.

### DIFF
--- a/docs/mint.json
+++ b/docs/mint.json
@@ -300,7 +300,8 @@
             "deployment/provision/overview",
             "deployment/provision/provider",
             "deployment/provision/workflow",
-            "deployment/provision/dashboard"
+            "deployment/provision/dashboard",
+            "deployment/provision/mapping"
           ]
         },
         "deployment/secret-store",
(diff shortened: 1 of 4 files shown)


ausias-armesto:
@shahargl Can you have a look at this pull request. It would be nice to have it.

PR #6551, opened 2026-05-31
fix(api): prevent background task garbage collection in lifespan startup

### LINKED ISSUE #6552: Bug: background tasks created in api.py lifespan can be garbage-collected mid-execution
## Describe the bug

\`keep/api/api.py\` creates two background tasks via \`asyncio.create_task(...)\` without holding a strong reference to the returned Task objects. Per the Python asyncio docs, the event loop only keeps **weak** references to running tasks, so a fire-and-forget task can be garbage-collected mid-execution and silently cancelled — no traceback, just missing work.

> Important: Save a reference to the result of [\`create_task()\`], to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn't referenced elsewhere may be garbage collected at any time, even before it's done.
> — https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

## Affected sites

Both on \`main\` at commit \`2217c15\`:

1. \`startup()\` line 183 — the maintenance-window watcher:
    \`\`\`python
    asyncio.create_task(process_watcher_task.async_process_watcher())
    \`\`\`
    Zero strong references. The Task can be GC'd before it ever runs the watcher.

2. \`lifespan()\` line 234 — the debug pending-task logger:
    \`\`\`python
    asyncio.create_task(check_pending_tasks(background_tasks))
    \`\`\`
    The coroutine receives \`background_tasks\` as an argument but the Task itself is never added to that set, so it lacks a strong reference too.

## Impact

- The maintenance-window watcher (#1) is registered unconditionally when \`SCHEDULER\` is enabled and no debug task is configured. If GC reclaims it, maintenance windows stop being processed silently.
- The pending-task logger (#2) only runs under \`KEEP_DEBUG_TASKS\`, so the blast radius for #2 is limited to debug builds, but the bug class is identical.

The probability of GC reclamation is non-deterministic and load-dependent, which is why this can lurk in production for a long time — but the [stdlib docs explicitly warn against the pattern](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task).

## Proposed fix

Add each task to the \`background_tasks\` set that \`lifespan()\` already maintains for request-context use, and register a done-callback so the task removes itself on completion. This is the canonical pattern from the asyncio docs.

PR: #6551

alexzhu0:
Closes #6552

**Problem**

Two `asyncio.create_task(...)` calls in `keep/api/api.py` create background tasks without holding a strong reference to the returned `Task` object. The Python asyncio docs warn that the event loop only keeps a *weak* reference to a task, so a fire-and-forget task can be garbage-collected mid-execution and silently cancelled — no traceback, just missing work.

> Important: Save a reference to the result of [`create_task()`], to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn't referenced elsewhere may be garbage collected at any time, even before it's done.
> — https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

**Affected sites**

1. `startup()` line 183 — the maintenance-window watcher:
    ```python
    asyncio.create_task(process_watcher_task.async_process_watcher())
    ```
    Zero strong references. The Task can be collected before it ever runs `async_process_watcher`, and when it does the watcher just disappears.

2. `lifespan()` line 234 — the debug pending-task logger:
    ```python
    asyncio.create_task(check_pending_tasks(background_tasks))
    ```
    `check_pending_tasks` is passed the `background_tasks` set as an argument, but the Task wrapping the coroutine is never added to that set. So this task also lacks a strong reference and is GC-eligible.

**Root cause**

CPython garbage-collects coroutine Tasks the same way it collects any other object. The asyncio event loop only keeps weak references to running tasks (`asyncio.all_tasks()` is implemented via a `WeakSet`). If no other code holds a strong reference, the GC is free to reclaim the Task at any GC pass, which cancels it.

**Changes**

`keep/api/api.py` — one file, +10 / -5 lines:

- `startup()` now accepts an optional `background_tasks: set | None = None` parameter; when present the watcher task is added to it and registers a done-callback to discard itself on completion (`task.add_done_callback(background_tasks.discard)`).
- `lifespan()` passes its existing `background_tasks` set into `startup()`, and now also adds the debug `check_pending_tasks` task to that same set with the same done-callback. The set was already being created here for request-context use; this just makes the lifespan tasks share it.
- The `set` type annotation is tightened to `set[asyncio.Task]` for clarity.

Backward compatible: any other caller of `startup()` continues to work because the new parameter defaults to `None`.

**Testing**

Manual reasoning + asyncio docs reference; the GC-cancellation behaviour is established Python asyncio semantics and the fix matches the [pattern recommended in the stdlib docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task) (hold strong reference in a set, discard via done-callback).

I considered adding a unit test but a test that reliably triggers `gc.collect()` mid-task in pytest would be fragile and the bug class is well-documented. Happy to add a regression test that simply asserts the watcher Task ends up in `app.state` if maintainers prefer.

### DIFF
--- a/keep/api/api.py
+++ b/keep/api/api.py
@@ -117,7 +117,7 @@ async def check_pending_tasks(background_tasks: set):
         await asyncio.sleep(1)
 
 
-async def startup():
+async def startup(background_tasks: set | None = None):
     """
     This runs for every worker on startup.
     Read more about lifespan here: https://fastapi.tiangolo.com/advanced/events/#lifespan
@@ -180,7 +180,10 @@ async def startup():
             except Exception:
                 logger.exception("Failed to start the maintenance windows")
         else:
-            asyncio.create_task(process_watcher_task.async_process_watcher())
+            task = asyncio.create_task(process_watcher_task.async_process_watcher())
+            if background_tasks is not None:
+                background_tasks.add(task)
+                task.add_done_callback(background_tasks.discard)
             logger.info(
                 "Added task",
                 extra={
@@ -227,14 +230,16 @@ async def lifespan(app: FastAPI):
     """
     app.state.limiter = limiter
     # create a set of background tasks
-    background_tasks = set()
+    background_tasks: set[asyncio.Task] = set()
     # if debug tasks are enabled, create a task to check for pending tasks
     if KEEP_DEBUG_TASKS:
         logger.info("Starting background task to check for pending tasks")
-        asyncio.create_task(check_pending_tasks(background_tasks))
+        debug_task = asyncio.create_task(check_pending_tasks(background_tasks))
+        background_tasks.add(debug_task)
+        debug_task.add_done_callback(background_tasks.discard)
 
     # Startup
-    await startup()
+    await startup(background_tasks=background_tasks)
 
     # yield the background tasks, this is available for the app to use in request context
     yield {"background_tasks": background_tasks}


alexzhu0:
Hi! This PR has been ready for review for 9 days with all checks passing. Happy to address any feedback or make adjustments.

alexzhu0:
Closing this for now — it's been open ~2 weeks without review bandwidth, and I'd rather not leave a stale PR lingering on the queue.

To be clear, nothing is wrong with the change: the two `asyncio.create_task(...)` sites in `keep/api/api.py` (the maintenance-window watcher at startup and the debug pending-task logger in `lifespan`) are still fire-and-forget at HEAD, so the GC-cancellation risk described in #6552 remains real, and the fix here (thread `background_tasks` through `startup()` + `add_done_callback(discard)`) still applies cleanly.

I'm leaving #6552 open since it documents the underlying bug independently. If a maintainer wants this picked up later, I'm happy to reopen or rebase — just ping me. Thanks!

PR #6556, opened 2026-06-04
fix(MonacoCel): move dynamic import to module level to prevent render-cycle interference

### LINKED ISSUE #6555: [🐛 Bug]: Infinite re-render loop on alerts feed  (MonacoCel dynamic import inside useEffect)
**Describe the bug**
After #6540, navigating to the alerts Feed page causes an infinite re-render loop that makes the UI completely unresponsive. The regression does not appear in the 0.52.1 release.

**Root cause**

#6540 moved the `monaco-editor` import from module-level (static) into a dynamic `import()` call **inside** `useEffect`:

```ts
// 🔴 after #6540 — dynamic import inside useEffect
useEffect(() => {
  import("monaco-editor")
    .then((monaco) => {
      loader.config({ monaco });
      return loader.init();
    })
    .then(() => setIsLoaded(true));
}, []);
```

Placing an async chunk load inside `useEffect` interferes with React's render cycle: the in-flight Promise keeps the component in its `null` (loading) state across multiple render passes, which — combined with the session loading transition in `useWebsocket` (`channelName` going from `"private-undefined"` to `"private-keep"`) — cascades into an unbounded re-render loop that saturates the main thread.

**To Reproduce**

1. Build Keep from `main`
2. Deploy with `AUTH_TYPE=OAUTH2PROXY`
3. Navigate to the alerts Feed page
4. Observe ~30,000 `useAlertPolling: Initializing` entries in the browser console and a frozen UI

**Expected behavior**
The alerts feed loads normally and the Monaco CEL editor initialises without impacting the render cycle.

ausias-armesto:
Closes #6555 — regression introduced by #6540.

## 📑 Description
#6540 moved `import("monaco-editor")` from module-level into a `useEffect` to fix an SSR crash ("window is not defined"). While the SSR fix is correct, placing the async chunk load **inside** `useEffect` causes it to interact with React's render cycle: the unresolved Promise keeps `MonacoCelBase` in its `null` loading state across multiple render passes. Combined with `useWebsocket`'s `channelName` transitioning during OAUTH2PROXY session load, this cascades into ~30,000 re-renders that freeze the UI.

## ✅ Checks

## Why this is correct

- **SSR-safe**: `typeof window !== "undefined"` prevents the import from running during server-side rendering — same guarantee as the original fix in #6540.
- **No render-cycle interference**: the dynamic import is started once, outside React, so it never blocks or re-triggers the render pipeline.
- **Performance**: Monaco starts loading as early as possible (module evaluation) rather than waiting for the component to mount.
- **Singleton**: the Promise is module-level, so even if multiple `MonacoCelBase` instances mount, Monaco is only configured once.

## Testing

1. Build from a commit `main`
2. Deploy with `AUTH_TYPE=OAUTH2PROXY`
3. Navigate to the alerts Feed page — previously caused ~30,000 re-renders and a frozen UI; now loads normally

### DIFF
--- a/keep-ui/shared/ui/MonacoCELEditor/MonacoCel.tsx
+++ b/keep-ui/shared/ui/MonacoCELEditor/MonacoCel.tsx
@@ -8,6 +8,20 @@ interface MonacoCelProps extends EditorProps {
   onMonacoLoadFailure?: (error: Error) => void;
 }
 
+// Monaco Editor - imported as an npm package instead of loading from the
+// CDN to support air-gapped environments.
+// https://github.com/suren-atoyan/monaco-react?tab=readme-ov-file#use-monaco-editor-as-an-npm-package
+//
+// The dynamic import is started at module evaluation (not inside useEffect) so
+// that it does not interfere with React's render cycle.  A typeof-window guard
+// makes it SSR-safe without causing "window is not defined" crashes.
+const monacoConfigPromise: Promise<void> | null =
+  typeof window !== "undefined"
+    ? import("monaco-editor").then((monaco) => {
+        loader.config({ monaco });
+      })
+    : null;
+
 export function MonacoCelBase(props: MonacoCelProps) {
   const [isLoaded, setIsLoaded] = useState(false);
   const onMonacoLoadedRef = useRef<MonacoCelProps["onMonacoLoaded"] | null>(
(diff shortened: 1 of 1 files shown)


shahargl (approved):
lgtm

PR #6560, opened 2026-06-08
feat: add `keep.dict_set` and `dict_merge` workflow helpers

### LINKED ISSUE #6488: [➕ Feature]: Add keep.dict_set helper for dict augmentation in workflow expressions
  Workflow body fields that need to carry a dict (e.g. an HTTP provider's body.commonAnnotations) cannot be built from existing data + a few overrides.
   The expression evaluator in keep/iohandler/iohandler.py is a hand-rolled AST walker, not Python eval(), and parse() only returns a dict when the
  entire field value is a single top-level keep.X(...) call whose arguments are literals or other keep.X(...) calls. Anything else (dict-spread,
  attribute access, vars, multiple top-level expressions) gets stringified via str(val) and ends up in the payload as a Python-repr blob.

  In practice this means there is no way to take an existing dict (e.g. keep.dictget(keep.json_loads('{{ alert }}'), 'annotations', {})) and add a
  single extra key. The only workarounds are (a) enumerate every key in the target dict by hand, or (b) drop the field entirely.

  Concrete trigger: keep/providers/prometheus_provider/prometheus_provider.py L193 pops description out of annotations into AlertDto.description. A
  workflow that forwards alerts back to an AM-compatible downstream (Grafana OnCall, another Alertmanager, etc.) can't reconstruct the full annotations
   dict because there is no way to dict | {"description": alert.description} inside a workflow expression. We have to enumerate summary, description,
  message, dashboard_url, runbook_url, … explicitly and maintain the list as new annotations appear.

  Add a small dict helper to keep/functions/__init__.py:
```
  def dict_set(d: dict, key: str, value) -> dict:
      """Return a new dict equal to `d` with `key` set to `value`."""
      return {**d, key: value}

  Optionally a variadic merge for the general case:

  def dict_merge(*dicts: dict) -> dict:
      """Merge dicts left-to-right; later keys override earlier."""
      out = {}
      for d in dicts:
          out.update(d)
      return out
```

  Both compose with existing helpers and stay inside the AST walker's "one Call with literal/Call args" constraint. Forwarding annotations with
  description re-added becomes one expression:

  commonAnnotations: "keep.dict_set(keep.dictget(keep.json_loads('{{ alert }}'), 'annotations', {}), 'description', keep.dictget(keep.json_loads('{{
  alert }}'), 'description', ''))"

  1. Enumerate every annotation explicitly. Works, but loses generic-forwarding: any new annotation on the source alert requires a workflow change. We
  currently do this — keep.dictget(keep.dictget(keep.json_loads('{{ alert }}'), 'annotations', {}), '<key>', '') repeated per field.
  2. Dict-spread inside a keep.X(...) call, e.g. keep.json_loads(keep.json_dumps({**keep.dictget(...), "description": ...})). Fails because
  _parse_token calls ast.literal_eval on dict-literal Call args; **spread and nested Call are not literals → ValueError.
  3. Using vars / aliases to hold the base dict. vars.X is stored without rendering (step.py L167-168) so it appears literally. aliases.X is rendered
  but stringified via str(dict) (Python repr, not JSON) when interpolated; inside a keep.X(...) call the AST walker treats aliases.X as ast.Attribute
  and crashes on arg.id.
  4. Patch the provider (e.g. .pop("description") → .get). Breaks the cross-provider normalization invariant (every provider has its own "main
  description" field — Sentry culprit, GCP summary, SignalFx displayBody, Prometheus description — Keep canonicalizes them into AlertDto.description).
  Not a real alternative.

  dict_set is the smallest change that preserves all current behavior and unblocks dict augmentation in workflow expressions. It's also useful far
  beyond this specific pop: any workflow that forwards payloads to a downstream system with a couple of extra fields runs into the same wall.

**Additional context**
  - Relevant source: keep/iohandler/iohandler.py — extract_keep_functions (L118-174), parse (L189-291), _parse_token (L293-410).
  - The .pop("description") lives at keep/providers/prometheus_provider/prometheus_provider.py L193-195 (and similarly in grafana_provider.py L841).
  - Real-world use case: forwarding Prometheus security alerts to Grafana OnCall's AlertManager integration through a Keep workflow. With dict_set we
  can ship a single-line generic forwarder. Without it, the workflow has to be updated every time a new annotation is introduced upstream.

Ishant5436:
Hey, saw issue #6488 and implemented dict_set and dict_merge in AST helpers to enable dynamic dictionary augmentation within workflow expressions. Wrote some unit tests for the edge cases too.

Closes #6488

### DIFF
--- a/keep/functions/__init__.py
+++ b/keep/functions/__init__.py
@@ -604,3 +604,56 @@ def dictget(data: str | dict, key: str, default: any = None) -> any:
         return default
 
     return data.get(key, default)
+
+
+def dict_set(data: str | dict, key: str, value: any) -> dict:
+    """
+    Sets a key-value pair in a dictionary, returning a new dictionary.
+    
+    Args:
+        data (str | dict): The dictionary to update. Can be a JSON string or dict.
+        key (str): The key to set.
+        value (any): The value to set.
+        
+    Returns:
+        dict: A new dictionary with the key set to the value.
+    """
+    if isinstance(data, str):
+        try:
+            data = json.loads(data)
+        except Exception:
+            data = {}
+
+    if not isinstance(data, dict):
+        data = {}
+
+    import copy
+    dict_copy = copy.deepcopy(data)
+    dict_copy[key] = value
+    return dict_copy
+
+
+def dict_merge(*args) -> dict:
+    """
+    Merges multiple dictionaries into one, returning a new dictionary.
+    Dictionaries passed later override earlier ones.
+    
+    Args:
+        *args: Dictionaries or JSON strings to merge.
+        
+    Returns:
+        dict: A new dictionary containing all merged keys and values.
+    """
+    result = {}
+    for data in args:
+        if isinstance(data, str):
+            try:
+                data = json.loads(data)
+            except Exception:
+                continue
+
+        if isinstance(data, dict):
+            result.update(data)
+            
+    return result
+
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -1217,3 +1217,48 @@ def test_from_timestamp_with_timezone():
     timestamp = dt.timestamp()
     result = functions.from_timestamp(timestamp, "Europe/Berlin")
     assert result == dt
+
+
+def test_dict_set():
+    """
+    Test dict_set function
+    """
+    # dict input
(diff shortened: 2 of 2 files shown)


shahargl (approved):
lgtm

PR #6566, opened 2026-06-09
fix: prevent AUTH0 test timeout caused by OIDC discovery HTTP call

### LINKED ISSUE #6565: [🐛 Bug]: AUTH0 tests time out due to real OIDC discovery HTTP call
**Describe the bug**

Tests using `AUTH_TYPE=AUTH0` fail with `Timeout >20.0s` because `auth0_authverifier.py` makes a real HTTP request to the Auth0 OIDC discovery endpoint at module import time. When the `test_app` fixture reloads `keep.api.api`, it cascades into re-importing the auth verifier — triggering the network call outside any mock context.

**Fix**

Wrap the module reload and `get_app()` call in a `patch("requests.get")` context when `AUTH_TYPE=AUTH0`, so the OIDC discovery call is intercepted and returns immediately.

ausias-armesto:
## Summary

Closes #6565

`auth0_authverifier.py` makes a real HTTP request at module import time when `AUTH0_DOMAIN` is set. The `test_app` fixture's module reload cascade re-imports the verifier outside any mock, causing a 10s network hang that exceeds the 20s per-test timeout.

**Fix:** move the module reload block and `get_app()` call inside a `patch("requests.get")` context when `AUTH_TYPE=AUTH0`, so the OIDC discovery call returns immediately without hitting the network.

## Files changed

- `tests/fixtures/client.py` — mock OIDC discovery during app startup for AUTH0 test fixtures

### DIFF
--- a/tests/fixtures/client.py
+++ b/tests/fixtures/client.py
@@ -1,12 +1,28 @@
 import hashlib
 import importlib
 import sys
+from contextlib import nullcontext
+from unittest.mock import MagicMock, patch
 
 import pytest
 from fastapi.testclient import TestClient
 
 from keep.api.core.dependencies import SINGLE_TENANT_UUID
 from keep.api.models.db.tenant import TenantApiKey
+
+
+def _mock_oidc_discovery():
+    """Return a context manager that stubs OIDC discovery HTTP calls.
+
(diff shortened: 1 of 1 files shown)


shahargl (approved):
Lgtm

PR #6574, opened 2026-06-11
feat(okta): group-to-role mapping, user auto-provisioning and userinfo enrichment

### LINKED ISSUE #6573: feat: Okta group-to-role mapping and user auto-provisioning
## Summary

The current Okta integration authenticates users correctly but lacks two features required for production use:

1. **Group-to-role mapping** — there is no way to map Okta group membership to Keep roles (admin, noc, webhook). All users fall back to the default role.
2. **User auto-provisioning** — users are not created in Keep's local DB on first login, so they don't appear in the users list.

### Environment variables
| Variable | Description |
|---|---|
| `OKTA_ADMIN_GROUPS` | Comma-separated Okta groups mapped to `admin` role |
| `OKTA_NOC_GROUPS` | Comma-separated Okta groups mapped to `noc` role |
| `OKTA_WEBHOOK_GROUPS` | Comma-separated Okta groups mapped to `webhook` role |
| `OKTA_USERINFO_URL` | Override the userinfo endpoint URL (optional) |

### Behaviour
- On each login, Keep calls Okta's `/v1/userinfo` endpoint to retrieve group memberships (group claims are not included in access tokens by default)
- The highest-priority matching group determines the user's role (`admin` > `noc` > `webhook`)
- Users are automatically created in Keep's DB on first login and their role/last-login are updated on subsequent logins
- The `OktaIdentityManager.get_users()` method now reads from Keep's local DB instead of returning an empty list

an0nym21:
## Summary

- Add \`OKTA_ADMIN_GROUPS\`, \`OKTA_NOC_GROUPS\`, \`OKTA_WEBHOOK_GROUPS\` env vars to map Okta group membership to Keep roles
- Use the Okta **ID token** (instead of access token) as the bearer credential sent to the Keep API, so that app-level group claims are visible to the backend
- Request the \`groups\` scope during OIDC authorization to trigger Okta's app-level group claim inclusion in the ID token
- Accept both \`OKTA_CLIENT_ID\` and \`OKTA_AUDIENCE\` as valid token audiences (ID tokens use the client ID as audience)
- Auto-provision users in Keep's DB on first login; update role and last-login on subsequent logins
- Fix \`OktaIdentityManager.get_users()\` which previously returned an empty list
- Add \`OKTA_USERINFO_URL\` env var to override the userinfo endpoint URL
- Document new env vars in \`docs/deployment/authentication/okta-auth.mdx\`

## How it works

Okta app-level Group Claims (configured under **Sign On → OpenID Connect ID Token**) are embedded in the **ID token** only — they do not appear in the access token or the \`/v1/userinfo\` response.

This PR switches the Okta auth flow to forward the ID token to the Keep API (matching the approach already used for Auth0 and OneLogin). The backend verifies the ID token signature via JWKS, reads the \`groups\` claim directly from the JWT payload, and resolves the highest-priority matching Keep role.

Role priority: \`admin\` > \`noc\` > \`webhook\`

## Configuration example

```env
OKTA_ADMIN_GROUPS=Keep_Admin
OKTA_NOC_GROUPS=Keep_User
OKTA_WEBHOOK_GROUPS=Keep_Webhook
```

> **Note:** the Okta app must have a Group Claim configured under Sign On → OpenID Connect ID Token with filter e.g. "Starts with \`Keep_\`".

Fixes #6573

### DIFF
--- a/docs/deployment/authentication/okta-auth.mdx
+++ b/docs/deployment/authentication/okta-auth.mdx
@@ -26,6 +26,9 @@ Keep supports Okta as an authentication provider, enabling:
 | `OKTA_AUDIENCE` | Expected audience claim in the token. Falls back to `OKTA_CLIENT_ID` if not set | No |
 | `OKTA_JWKS_URL` | Explicit JWKS URL. If not set, derived from `OKTA_ISSUER` | No |
 | `OKTA_API_TOKEN` | Okta API token for management operations | No |
+| `OKTA_ADMIN_GROUPS` | Comma-separated Okta group names that map to the `admin` role | No |
+| `OKTA_NOC_GROUPS` | Comma-separated Okta group names that map to the `noc` role | No |
+| `OKTA_WEBHOOK_GROUPS` | Comma-separated Okta group names that map to the `webhook` role | No |
 
 ### Frontend Environment Variables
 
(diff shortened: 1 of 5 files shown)


an0nym21:
Hi 👋 — I've updated the PR description to accurately reflect the final implementation (the key insight was that Okta app-level group claims are only available in the ID token, not the access token or userinfo endpoint).

Could someone take a look and let me know if anything is missing before this can be merged? Specifically:

- Is the approach of forwarding the **ID token** to the API acceptable? (Auth0 and OneLogin already do this in the codebase)
- Any concerns about the audience verification change (accepting both `OKTA_CLIENT_ID` and `OKTA_AUDIENCE`)?
- Is the documentation in `okta-auth.mdx` sufficient?

Thanks!

PR #6576, opened 2026-06-11
fix(ui): exclude public static assets from auth middleware to fix image loading

### LINKED ISSUE #6575: fix: local images not loading — middleware redirects internal /_next/image fetches to signin
## Bug

In self-hosted deployments, local images served via `/_next/image` fail with:

```
The requested resource isn't a valid image.
```

Example: `/_next/image?url=%2Fkeep.png&w=64&q=75`

## Root cause

When Next.js image optimization processes a local image, it fetches the source file internally via `fetchInternalImage`. This internal request goes through the full Next.js middleware pipeline.

`/keep.png` (and other public static assets) are **not excluded** from the middleware matcher. Since the internal request carries no auth session, the middleware redirects it to `/signin`, returning an HTML page instead of image data. Next.js then fails to detect a valid image content-type and throws the error.

Note: `keep.svg` was already excluded from the matcher but `keep.png` was not.

## Fix

Add public static asset extensions (`.png`, `.jpg`, `.jpeg`, `.gif`, `.webp`, `.ico`) to the middleware matcher exclusion list so they bypass authentication.

an0nym21:
## Summary

Local images served via `/_next/image` were failing with "The requested resource isn't a valid image" in self-hosted deployments.

## Root cause

When Next.js image optimization (`/_next/image`) fetches a local image internally via `fetchInternalImage`, the request goes through the full middleware pipeline. Public assets like `/keep.png` were not excluded from the middleware matcher, so the middleware redirected them to `/signin` — returning HTML instead of image data, which Next.js cannot process as an image.

`keep.svg` was already in the exclusion list but `keep.png` was not.

## Fix

Add common static image extensions (`.png`, `.jpg`, `.jpeg`, `.gif`, `.webp`, `.ico`) to the middleware matcher negative lookahead so public assets bypass authentication entirely.

## Test

Before: `/_next/image?url=%2Fkeep.png&w=64&q=75` → "The requested resource isn't a valid image."
After: image loads correctly.

Fixes #6575

### DIFF
--- a/docker/Dockerfile.ui
+++ b/docker/Dockerfile.ui
@@ -55,6 +55,12 @@ COPY --from=builder /app/public ./public
 # https://nextjs.org/docs/advanced-features/output-file-tracing
 COPY --from=builder --chown=nextjs:nodejs /app/.next/standalone ./
 COPY --from=builder --chown=nextjs:nodejs /app/.next/static ./.next/static
+
+# sharp is required for Next.js image optimization in standalone mode.
+# The standalone output does not include native modules automatically.
+# sharp v0.33+ uses @img/* packages for its native bindings.
+COPY --from=deps --chown=nextjs:nodejs /app/node_modules/sharp ./node_modules/sharp
+COPY --from=deps --chown=nextjs:nodejs /app/node_modules/@img ./node_modules/@img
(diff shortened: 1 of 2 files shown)


shahargl (approved):
lgtm

PR #6583, opened 2026-06-16
fix: parse PUSHER_USE_SSL and PUSHER_DISABLED env booleans correctly

### LINKED ISSUE #6582: [🐛 Bug]: PUSHER_USE_SSL is incorrectly parsed

```
    "1",
    "true",
    "yes",
    "on",
)

pusher = Pusher(
    host=pusher_host,
    app_id=pusher_app_id,
    key=pusher_app_key,
    secret=pusher_app_secret,
    ssl=pusher_use_ssl,
)
```

SueJianjian:
## Summary
- parse PUSHER_USE_SSL and PUSHER_DISABLED using explicit boolean coercion instead of Python truthiness
- preserve existing behavior for already-boolean values
- add focused regression coverage for string env inputs

Closes #6582

## Testing
- .venv313\\Scripts\\python.exe -m pytest tests/test_dependencies.py -q

### DIFF
--- a/keep/api/core/dependencies.py
+++ b/keep/api/core/dependencies.py
@@ -23,6 +23,14 @@ if PUSHER_ROOT_CA:
     pusher_requests.CERT_PATH = PUSHER_ROOT_CA
 
 
+def parse_env_bool(value: str | bool | None, default: bool = False) -> bool:
+    if value is None:
+        return default
+    if isinstance(value, bool):
+        return value
+    return value.strip().lower() in {"1", "true", "yes", "on"}
+
+
 async def extract_generic_body(request: Request) -> dict | bytes | FormData:
     """
     Extracts the body of the request based on the content type.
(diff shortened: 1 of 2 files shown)


shahargl (changes requested):
I think this is a bit over engineering, you can use `Config` class for that

SueJianjian:
Updated this to use the existing Config bool parsing instead of a custom helper, and switched the tests to cover get_pusher_client() behavior directly. Could you take another look when you have a moment?

SueJianjian:
Hi @shahargl, I updated this to use the existing Config bool parsing path and adjusted the tests to cover get_pusher_client() behavior directly. Could you please take another look when you have a moment?

SueJianjian:
Quick follow-up for review: the requested Config-based boolean parsing is in place on the current head, with focused get_pusher_client() tests and all checks passing. @shahargl, could you re-review when convenient?

shahargl (approved):
lgtm

PR #6584, opened 2026-06-16
fix: parse PUSHER_USE_SSL as a boolean env flag

### LINKED ISSUE #6582: [🐛 Bug]: PUSHER_USE_SSL is incorrectly parsed

SueJianjian:
## Summary

Closes #6582

- fixes `PUSHER_USE_SSL` parsing in `keep/api/core/dependencies.py`
- treats string env values like `false`, `0`, `no`, and `off` as `False`
- reuses the same parsing helper for `PUSHER_DISABLED` to avoid the same class of bug
- adds narrow regression coverage for env flag parsing

## Why

`PUSHER_USE_SSL` currently uses:

```python
ssl=False if os.environ.get(PUSHER_USE_SSL, False) is False else True
```

When the app is configured through Docker Compose, `PUSHER_USE_SSL: false` is passed as the string `false`, which is not the boolean `False`. That makes the current expression evaluate to `True`, so SSL is enabled unexpectedly.

## Change

- add `_env_flag(name, default=False)` for simple env-based boolean parsing
- switch `PUSHER_USE_SSL` to `ssl=_env_flag(PUSHER_USE_SSL)`
- switch `PUSHER_DISABLED` to `_env_flag(PUSHER_DISABLED)`

## Validation

Local lightweight verification with Python 3.13 and minimal runtime deps:

- `PUSHER_USE_SSL=false -> _ssl=False`
- `PUSHER_USE_SSL=true -> _ssl=True`
- `_env_flag` also correctly handles `0/1` and `false/true`

I also added a narrow regression test file:

- `tests/test_pusher_dependencies.py`

## Notes

- I kept the scope limited to the Pusher env parsing path instead of broadening this into a repo-wide boolean parsing refactor.

### DIFF
--- a/keep/api/core/dependencies.py
+++ b/keep/api/core/dependencies.py
@@ -23,6 +23,15 @@ if PUSHER_ROOT_CA:
     pusher_requests.CERT_PATH = PUSHER_ROOT_CA
 
 
+def _env_flag(name: str, default: bool = False) -> bool:
+    value = os.environ.get(name)
+    if value is None:
+        return default
+    if isinstance(value, str):
+        return value.strip().lower() in {"1", "true", "yes", "on"}
+    return bool(value)
+
+
 async def extract_generic_body(request: Request) -> dict | bytes | FormData:
     """
     Extracts the body of the request based on the content type.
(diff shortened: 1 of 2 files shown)


SueJianjian:
Linked the PR to #6582 in the description now.

shahargl:
just use `Config`

PR #6589, opened 2026-06-21
fix: Prevent alerts facets sidebar from freezing with high facet cardinality (#6577)

### LINKED ISSUE #6577: [🐛 Bug]: Frontend 0.53.0 hangs on Alerts page due to excessive facet rendering (Regression from 0.52.0)
**Describe the bug**
After upgrading keep-ui from v0.52.0 to v0.53.0, the Alerts page become completely unresponsive ("Just a second, getting your data..." spinner freezes indefinitely).
The browser's main thread is blocked by JavaScript execution during the hydration/rendering of the facets sidebar.
Downgrading keep-ui back to v0.52.0 while keeping keep-core at v0.53.0 immediately resolves the issue.
This confirms a frontend-only regression introduced in v0.53.0.

**To Reproduce**
1. Deploy KeepHQ with keep-core:0.53.0 and keep-ui:0.53.0.
2. Ingest alerts from Prometheus/Alertmanager that contain a high cardinality of unique labels (e.g., Kubernetes pods, shards, instances across multiple clusters). In our case, we have ~268k alerts generating 200+ unique facet fields.
3. Navigate to /alerts/feed.
4. Observe that the page never finishes loading. The browser tab becomes unresponsive.

**Expected behavior**
The page should load within a few seconds. Facets should either be paginated, limited to a configurable whitelist, or rendered asynchronously without blocking the main thread.

**Screenshots**
I have attached Firefox Performance Profiles for both versions.
The most critical finding is the massive memory regression in v0.53.0 during page hydration:

What the screenshots show:
| Version | Memory Usage | Status | Evidence |
| ----- | ----- | ----- | ----- |
| v0.52.0 | ~100 MB | Healthy | Stable memory footprint during facet rendering. |
| v0.53.0 | >1 GB | Critical | Uncontrolled memory allocation spike. 10x increase compared to working version. |

1. Memory Timeline: Hovering over the timeline in the profiler clearly shows v0.52.0 staying flat at ~100MB, while v0.53.0 shoots up past 1GB within seconds of loading /alerts/feed.
2. Allocation Call Tree: In v0.53.0, the memory allocations are concentrated in React component creation functions (VoidFunction, ik) inside chunk 4bd1b696...js.
This suggests that for every unique label key (~200+), the new version is allocating heavy objects (possibly memoized selectors, state copies, or unbounded arrays) that are never garbage collected.
3. Main Thread Blocking: The memory pressure correlates directly with the main thread freeze.
The browser spends all its time managing this massive object graph instead of painting the UI.

Severity Note: This is not just a UX issue.
A 1GB+ memory spike for a single page load will cause tab crashes on machines with limited RAM and forces aggressive GC cycles that degrade performance for all other tabs.

**Environment Details**
* Keep Version (Working): keep-core:0.53.0 + keep-ui:0.52.0
* Keep Version (Broken): keep-core:0.53.0 + keep-ui:0.53.0
* Browser: Chrome 148, Firefox 151
* Data Volume: ~268,000 alerts, 200+ unique facet fields
* Deployment: Docker Compose behind Nginx + oauth2-proxy (HTTP/2 enabled)

**Workaround**
We attempted to mitigate this using the ALERT_SIDEBAR_FIELDS environment variable to whitelist only essential fields (service,severity,environment,alertname,status,fingerprint), but the UI continued to hang even with only 30 alerts in the database.
This suggests that in v0.53.0, the component responsible for collecting/rendering facets ignores the whitelist configuration or performs an expensive pre-render scan of all available fields before applying filters.

**Additional environment Optimizations**
To ensure the issue was not caused by backend/database performance, we have heavily optimized our infrastructure:
* Database Indexes: We created specialized indexes on the alert table to handle high-cardinality labels efficiently:
  * GIN index on event column (jsonb_path_ops) for fast JSON key extraction.
  * Composite B-tree index on (tenant_id, status) with partial condition for firing alerts.
  * These indexes reduced facet collection query time from seconds to milliseconds.
* Backend Performance: With these indexes, the /backend/alerts/facets/fields endpoint returns the list of 200+ fields in <200ms. The bottleneck is confirmed to be purely client-side rendering.
* Infrastructure: Nginx + HTTP/2, oauth2-proxy with increased buffers, PostgreSQL tuned for write-heavy workloads.
* Workaround: Currently running keep-ui:0.52.0 with keep-core:0.53.0. This combination works perfectly with the same database and indexes, confirming the regression is isolated to the frontend codebase in v0.53.0.

Possible Root Cause
It appears that v0.53.0 changed how the facets sidebar is initialized. Instead of lazily loading or respecting the ALERT_SIDEBAR_FIELDS whitelist during the initial render, it seems to synchronously process and create React components for every unique key found in the alert metadata, causing a complete UI freeze when label cardinality is high.

shahargl:
## Summary

Fixes #6577 — the Alerts page hangs and spikes browser memory past 1GB when there is high facet cardinality (200+ unique facet fields). The reporter's profiler shows React component-creation allocations concentrated per unique label key, growing an object graph that is never GC'd, blocking the main thread.

### On the "regression from 0.52.0"

I traced the real release boundaries (0.52.0 = #6510, 0.53.0 = #6558). The `features/filter` facets **source code is byte-identical between 0.52.0 and 0.53.0**, and there were no `package.json`/lockfile/Next.js dependency changes in that range. So this is **not** a specific source-line regression — it is a **latent scalability flaw in the facets panel** that the 0.53.0 build exposed (the reporter confirms swapping only `keep-ui` 0.53→0.52 fixes it with the same backend; the profiler points at a minified webpack chunk, consistent with a production-build/bundle difference rather than a source change). Regardless of which build crossed the freeze threshold, the allocation source the profiler identified is the always-present behavior fixed here.

### Root cause (the allocations the profiler points at)

The server-side facets panel does several things that scale poorly with facet count:

- **Every facet renders expanded eagerly and loads options on mount.** One `Facet` component + memoized selectors + render closures are instantiated per facet, all open. The backend already marks user-defined facets as `is_lazy`, but the frontend ignored the flag.
- **Option queries built for all facets at once** via an O(n²) cross-filter loop.
- **Whole-object Zustand subscriptions**: each `Facet` subscribed to the entire `facetsState`/`facetOptionsLoadingState`, so any toggle re-rendered all 200+ facets.
- **`JSON.parse(JSON.stringify(...))` deep clone** of the entire options tree on every ~5s poll — the growing, never-GC'd object graph.

### Changes

- **Respect `is_lazy`**: lazy facets start **collapsed** and only fetch options when expanded (or when they have a restored selection from URL params). A new `activeFacetIds` store concept tracks which facets should load options, preventing the initial bulk-load of 200+ facets.
- **Scope option queries to active facets** in `useQueriesHandler`.
- **Per-facet store subscriptions**: each `Facet` subscribes only to its own slice, eliminating the 200× re-render cascade.
- **Remove the deep clone** in `useFacetOptions` — shallow per-facet merge instead.
- Bumps version `0.53.0` → `0.54.0`.

Note: `ALERT_SIDEBAR_FIELDS` (the workaround attempted in the issue) only controls the alert *detail* sidebar, not the facets filter panel, which is why it had no effect.

## Test plan

- [x] `cd keep-ui && npx jest features/filter` — 48 passed (incl. new tests for lazy activation, expand-to-activate, clear-filters, and excluding inactive facets from option queries)
- [x] `tsc --noEmit` clean for `features/filter` (remaining repo errors are pre-existing/unrelated)
- [ ] Manual: load Alerts page with 200+ facets → page responsive, lazy facets collapsed, options load on expand, memory stays flat
- [ ] Manual: facet selections restored from a shared URL still auto-expand and show the active filter

### DIFF
--- a/.github/workflows/release-workflow-schema.yml
+++ b/.github/workflows/release-workflow-schema.yml
@@ -88,87 +88,93 @@ jobs:
           name: workflow-schema
           path: workflow-yaml-json-schema.json
 
-  release-schema:
-    runs-on: ubuntu-latest
-    needs: generate-schema
-    if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.head.repo.fork }}
-
-    steps:
-      - name: Download schema artifact
-        uses: actions/download-artifact@v4
-        with:
-          name: workflow-schema
-          path: .
-      - name: Checkout schema repository
-        uses: actions/checkout@v4
-        with:
-          repository: ${{ env.SCHEMA_REPO_NAME }}
-          token: ${{ secrets.SCHEMA_REPO_PAT }}
-          path: schema-repo
-
-      - name: Set target branch variable
-        id: set_branch
-        run: |
-          if [ "${{ github.event_name }}" = "pull_request" ]; then
-            echo "branch=${{ github.head_ref }}" >> $GITHUB_OUTPUT
-          else
-            echo "branch=${{ github.ref_name }}" >> $GITHUB_OUTPUT
-          fi
-
-      - name: Create or switch to target branch in schema repo
-        working-directory: schema-repo
-        run: |
-          git fetch origin
-          if git show-ref --verify --quiet refs/heads/${{ steps.set_branch.outputs.branch }}; then
-            git checkout ${{ steps.set_branch.outputs.branch }}
-          else
-            git checkout -b ${{ steps.set_branch.outputs.branch }}
-          fi
-
-      - name: Copy schema to target repository
-        run: |
-          cp workflow-yaml-json-schema.json schema-repo/schema.json
-
-          # Update schema with version info
-          jq --arg version "${{ needs.generate-schema.outputs.version }}" \
-             --arg id "https://raw.githubusercontent.com/${{ env.SCHEMA_REPO_NAME }}/v${{ needs.generate-schema.outputs.version }}/schema.json" \
-             '. + {version: $version, "$id": $id}' \
-             schema-repo/schema.json > schema-repo/schema.tmp.json
-
-          mv schema-repo/schema.tmp.json schema-repo/schema.json
-
-      - name: Check if schema changed
-        id: check_changes
-        working-directory: schema-repo
-        run: |
-          git add schema.json
-          if git diff --cached --quiet schema.json; then
-            echo "changed=false" >> $GITHUB_OUTPUT
-          else
-            echo "changed=true" >> $GITHUB_OUTPUT
-          fi
-
-      - name: Commit and push schema
-        if: steps.check_changes.outputs.changed == 'true'
-        working-directory: schema-repo
-        run: |
-          git config user.name "Keep Schema Bot"
-          git config user.email "no-reply@keephq.dev"
-          git commit -m "Release schema v${{ needs.generate-schema.outputs.version }}"
-          git push origin ${{ steps.set_branch.outputs.branch }}
-          if [ "${{ steps.set_branch.outputs.branch }}" = "main" ]; then
-            git tag "v${{ needs.generate-schema.outputs.version }}"
-            git push origin "v${{ needs.generate-schema.outputs.version }}"
-          fi
-
-      - name: Create GitHub Release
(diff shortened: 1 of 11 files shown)


davidchagalidze:
thanks @shahargl

PR #6591, opened 2026-06-23
feat: include fingerprints in poll-alerts payload and patch visible alert rows

### LINKED ISSUE #6592: [➕ Feature]: Include fingerprints in poll-alerts payload ( Pusher ) to reduce full alert table refetch
## Problem

`poll-alerts` Pusher events currently trigger a broad `/alerts/query` refresh with a recomputed `lastReceived` time window, even when only a single alert changed.

## Proposal

- Send `{"fingerprints": [...]}` in existing `poll-alerts` payloads
- Parse fingerprint metadata on the frontend
- Patch visible alert rows via `GET /alerts/{fingerprint}` instead of refetching the full table
- Keep existing refresh fallback when fingerprints are missing or empty

## Expected outcome

Faster UI updates for alerts already visible in the table, with backward-compatible behavior for legacy payloads.

harisrujanc:
Closes #6592

## Summary

- Include alert fingerprints in existing `poll-alerts` Pusher payloads.
- Parse fingerprint metadata in `useAlertPolling` while keeping the existing `data` refresh token.
- Patch visible alert rows in the alerts table via `GET /alerts/{fingerprint}` instead of triggering a full time-window refetch.
- Preserve existing polling fallback when fingerprints are missing or empty.

## Why

`poll-alerts` currently triggers a broad `/alerts/query` refresh with a recomputed `lastReceived` window. For updates to alerts already on screen, fingerprint metadata lets the UI patch only those rows.

## Test plan

- [x] `npm run test -- utils/hooks/__tests__/useAlertPolling.test.ts widgets/alerts-table/ui/__tests__/useAlertsTableData.test.ts`

### DIFF
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -299,6 +299,10 @@ jobs:
       - name: Dump logs
         if: always()
         run: |
+          if [ ! -f tests/e2e_tests/docker-compose-modified.yml ]; then
+            echo "docker-compose-modified.yml not found; skipping log collection"
+            exit 0
+          fi
           docker compose -p keep --project-directory . -f tests/e2e_tests/docker-compose-modified.yml logs keep-backend > backend_logs-${{ inputs.db-type }}.txt
           docker compose -p keep --project-directory . -f tests/e2e_tests/docker-compose-modified.yml logs keep-frontend > frontend_logs-${{ inputs.db-type }}.txt
           docker compose -p keep --project-directory . -f tests/e2e_tests/docker-compose-modified.yml logs keep-backend-db-auth > backend_logs-${{ inputs.db-type }}-db-auth.txt
@@ -330,4 +334,8 @@ jobs:
       - name: Tear down environment
         if: always()
         run: |
-          docker compose -p keep --project-directory . -f tests/e2e_tests/docker-compose-modified.yml down
+          if [ -f tests/e2e_tests/docker-compose-modified.yml ]; then
+            docker compose -p keep --project-directory . -f tests/e2e_tests/docker-compose-modified.yml down
+          else
+            echo "docker-compose-modified.yml not found; skipping tear down"
+          fi
--- a/keep-ui/utils/hooks/__tests__/useAlertPolling.test.ts
+++ b/keep-ui/utils/hooks/__tests__/useAlertPolling.test.ts
@@ -1,0 +1,28 @@
+import { parsePollAlertsPayload } from "../useAlertPolling";
+
+describe("parsePollAlertsPayload", () => {
+  it("returns empty array for missing payload", () => {
+    expect(parsePollAlertsPayload(null)).toEqual([]);
+    expect(parsePollAlertsPayload(undefined)).toEqual([]);
+  });
+
+  it("parses fingerprints from object payload", () => {
+    expect(
+      parsePollAlertsPayload({ fingerprints: ["fp-1", "fp-2"] })
+    ).toEqual(["fp-1", "fp-2"]);
+  });
+
+  it("parses fingerprints from legacy string payload", () => {
+    expect(parsePollAlertsPayload('{"fingerprints":["fp-1"]}')).toEqual([
+      "fp-1",
+    ]);
+  });
+
+  it("filters invalid fingerprint values", () => {
+    expect(
+      parsePollAlertsPayload({
+        fingerprints: ["fp-1", "", 2, null, "fp-2"],
+      })
+    ).toEqual(["fp-1", "fp-2"]);
+  });
+});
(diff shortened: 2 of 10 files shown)


shahargl (changes requested):
Thanks for this — the intent (patch on-screen rows instead of a full time-window refetch) is great and the code is clean with good unit coverage. However, there are a couple of blocking issues, the biggest being that this "optimization" actually pessimizes the backend, plus some behavioral regressions in the live feed. Requesting changes.

## Blocking

### 1. `GET /alerts/{fingerprint}` is a full-list scan — patching N rows = N full scans
The endpoint this PR relies on (`keep/api/routes/alerts.py::get_alert`, ~line 781) is implemented as a full fetch + Python filter:

```python
all_alerts = get_all_alerts(authenticated_entity=authenticated_entity)  # get_last_alerts(limit=1000) + convert_db_alerts_to_dto_alerts
alert = list(filter(lambda a: a.fingerprint == fingerprint, all_alerts))
```

`get_all_alerts` fetches up to 1000 last alerts and converts them all to DTOs on every call. The frontend (`useAlertsTableData.ts::patchVisibleAlerts`) then fires one such request **per fingerprint, in parallel** via `Promise.all(fingerprintsToPatch.map(... api.get(`/alerts/${fingerprint}`)))`. So a `poll-alerts` event for e.g. 20 ingested alerts triggers 20 concurrent full-list fetch+convert operations — potentially far heavier on the backend than the single `/alerts/query` it replaces.

**Ask:** use a real point lookup (`get_last_alert_by_fingerprint` already exists in `db.py`) or add a single batched endpoint (e.g. `GET /alerts?fingerprints=...`) and patch in one round-trip.

### 2. New alerts are silently dropped from the live view
The patch path (`useAlertsTableData.ts::patchVisibleAlerts`) only updates fingerprints that are already visible:
```ts
const fingerprintsToPatch = fingerprints.filter((fp) => visibleFingerprints.has(fp));
```
When a brand-new alert is ingested, its fingerprint isn't on screen, so it's filtered out and never appears until some other trigger causes a full refetch. The previous behavior (`setShouldRefreshDate(true)`) surfaced new alerts. This is a functional regression for the primary "alerts streaming in" case — the feed will look stale.

**Ask:** keep the full-refetch fallback when any polled fingerprint is not currently visible (i.e. a new alert), and only fast-path when all polled fingerprints are already on screen.

## Should fix

### 3. Sort order / counts drift
`mergeAlertsByFingerprint` replaces rows in place via a `Map` keyed by fingerprint, preserving the old position. With sort = `lastReceived`, a freshly-updated alert should move but stays put. Also `totalCount` and facet counts aren't updated on the patch path, so the header/facets can disagree with the rows.

### 4. Resolved/dismissed alerts won't be removed
If an enrichment moves an alert out of the current filter (e.g. dismissed), `GET /alerts/{fingerprint}` still returns it and `mergeAlertsByFingerprint` keeps it. The patch path doesn't re-evaluate the row against `filterCel`. Either re-evaluate or scope the feature explicitly to "update existing rows" + keep refetch for everything else.

### 5. Pusher payload size
`process_event_task.py::__handle_formatted_events` now sends `{"fingerprints": [...all ingested fingerprints...]}`. Pusher has a ~10KB message limit; a large batch ingest (each fingerprint is a 64-char hash → ~250 fingerprints) can exceed it. Please cap/chunk, or fall back to the token-only signal above a threshold.

## Nits / notes
- Backward-compat during rolling deploy: payload changed from the literal `"{}"` to an object. `parsePollAlertsPayload` handles object/legacy-string/missing (nice), but please confirm an old frontend still bound to `poll-alerts` tolerates receiving an object (it ignored the arg before, so likely fine).
- Nice cleanup removing the `console.log`s, and `parsePollAlertsPayload` is well tested.

## Before merge
- Replace per-fingerprint full-scan with a point/batched lookup.
- Preserve full-refetch fallback for not-yet-visible (new) fingerprints.
- Handle removal + count/sort updates, or explicitly scope to existing-row updates.
- Cap/chunk the Pusher fingerprint list.
- Wait for the pending integration/DB CI jobs to go green.

harisrujanc:
Good catch, fixing the ones mentioned:

- **#1:** Dropping per-fingerprint `GET`, adding `POST /alerts/batch` (`fingerprint IN (...)` DB query). One round-trip regardless of N.
- **#2:** Fast-patch only runs when every polled fingerprint is already on screen; any new fingerprint falls through to the existing full-refresh path.
- **#3:** `alertsToReturn` re-sorted after merge; `setFacetsPanelRefreshToken` called on each patch.
- **#4:** Fingerprints absent from the batch response are removed from the row set.
- **#5:** Capping at 100 fingerprints; larger batches send `{"fingerprints": []}` which hits the full-refresh fallback.

old frontends ignored the `poll-alerts` payload so receiving an object is safe.

shahargl (approved):
lgtm

PR #4265, opened 2025-03-27
docs: adding GKE specific workaround to the cilium provider documentation

aksenk:
## 📑 Description
If you are using a GKE cluster, you cannot connect Keep to the Google-managed hubble-relay directly because:
- hubble-relay operates only in secure mode,
- hubble-relay requires client certificate authentication.

However, Keep does not currently support these features.

To work around this, you can add an NGINX Pod that listens on a plaintext HTTP port and proxies requests to hubble-relay secure port using hubble-relay certificates.

## ✅ Checks

- [x] My pull request adheres to the code style of this project
- [x] My code requires changes to the documentation
- [x] I have updated the documentation as required
- [x] All the tests have passed

## ℹ Additional Information
I created feature request [issue](https://github.com/keephq/keep/issues/4264) for root cause.

PR #6606, opened 2026-06-28
feat(cilium): support secure TLS connection to the hubble relay

### LINKED ISSUE #4264: [➕ Feature]: Cilium provider secure connection
**Problem**

If you are using cilium with hubble-relay secure connection (for example in a GKE cluster and google-managed cilium), you cannot connect Keep hubble-relay directly because:
- hubble-relay operates only in secure mode,
- hubble-relay requires client certificate authentication.

However, Keep does not currently support these features.

**Resolution**

Adding support:
- hubble-relay secure connection
- hubble-relay client certificate authentication 

**Current workaround**
#4265

aditya-786:
### Problem

The Cilium provider connects to the Hubble relay with `grpc.insecure_channel`, so the connection is plaintext only. As raised in #4264, the documented GKE approach (#4265) is an insecure workaround rather than a real fix, and there is currently no way to connect to a Hubble relay that requires TLS or mutual TLS.

### Change

Add optional TLS to the Cilium provider:

- `use_tls` (switch) connects to the relay over TLS.
- `ca_certificate` (PEM) verifies the relay server; when omitted, the system trust store is used.
- `client_certificate` + `client_key` (PEM) enable mutual TLS.

Default behavior is unchanged: `use_tls` defaults to `false`, which keeps the existing `grpc.insecure_channel` path. The channel construction is extracted into a small grpc-only helper `build_cilium_channel` (`keep/providers/cilium_provider/secure_channel.py`), and `pull_topology` now calls it.

### Design note

The cert fields take PEM content via `file` config fields (the same convention as the GKE service-account field), which keeps the helper IO-free and fits Keep's secret handling. Happy to switch these to file paths instead if you prefer.

### Tests

`tests/test_cilium_secure_channel.py` covers the insecure default, mutual TLS, server-only TLS, and TLS using the system trust store, asserting the exact gRPC credential calls:

```
$ pytest tests/test_cilium_secure_channel.py -q
4 passed
```

The test imports only the standalone helper (not the provider, which pulls the wider app), so it runs without the full backend. Verified fail-before/pass-after.

### DIFF
--- a/docs/providers/documentation/cilium-provider.mdx
+++ b/docs/providers/documentation/cilium-provider.mdx
@@ -40,6 +40,10 @@ The Cilium provider leverages Hubble's network flow data to automatically discov
 | Parameter | Description | Example |
 |-----------|-------------|----------|
 | `cilium_base_endpoint` | The base endpoint of the Cilium Hubble relay | `localhost:4245` |
+| `use_tls` | Connect to the Hubble relay over TLS | `true` |
+| `ca_certificate` | CA certificate (PEM) used to verify the Hubble relay server | `-----BEGIN CERTIFICATE-----` |
+| `client_certificate` | Client certificate (PEM) for mutual TLS | `-----BEGIN CERTIFICATE-----` |
+| `client_key` | Client private key (PEM) for mutual TLS | `-----BEGIN PRIVATE KEY-----` |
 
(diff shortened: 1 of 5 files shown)


shahargl (approved):
lgtm

PR #2397, opened 2024-11-05
chore(provider): zabbix from mediatype to script

talboren:

PR #6612, opened 2026-07-03
fix(zabbix): populate hostname on pulled and pushed alerts

### LINKED ISSUE #6613: Zabbix alerts missing host name
## Bug

Zabbix alerts often arrive without a host name attached, in both the polling and webhook-push paths of the Zabbix provider.

- **Polling (`_get_alerts` in `keep/providers/zabbix_provider/zabbix_provider.py`)**: the `problem.get` API call never requests host data (no `selectHosts` param), so pulled alerts have no host information at all.
- **Webhook push (`_format_alert`)**: the installed Zabbix script sends the host as `host_name` / `host_ip` (matching the parameter names configured in `setup_webhook`), but the code reads `hostName` / `hostIp` instead — these keys never exist in the payload. The primary source, `service` (`{HOST.HOST}`), can be empty for problems tied to multiple hosts, and the fallback then silently returns `None` due to this key mismatch.

## Fix

See linked PR.

washosk:
Fixes #6613

## Summary

Zabbix alerts were missing the host name in two separate ways:

- **Polling (`_get_alerts`)**: the `problem.get` call never asked Zabbix for host data, so pulled alerts had no host information at all.
- **Webhook push (`_format_alert`)**: the installed Zabbix script sends the host as `host_name` / `host_ip` (matching the parameter names configured in `setup_webhook`), but the code read `hostName` / `hostIp` instead. The primary source, `service` (`{HOST.HOST}`), can be empty for problems tied to multiple hosts, and the fallback silently returned `None` due to this key mismatch.

## Changes

- `_get_alerts`: `problem.get` does **not** support `selectHosts` (verified against a live instance — it errors with `Invalid parameter: unexpected parameter selectHosts`). Fetch hosts via a separate batched `event.get` call on the same eventids and join them back in by eventid, mapping the host's visible name (falling back to its technical name) onto `service`/`hostname` on the `AlertDto`. Adds the `event.get` scope.
- `_format_alert`: read `host_name`/`host_ip` (matching the actual webhook payload keys), falling back through `hostName`/`hostIp` and the older `HOST.NAME`/`HOST.IP` macro keys for backward compatibility with webhook installs set up before the mediatype→script migration (#2397).

**Note (added while re-testing against a live instance):** also fixes an unrelated pre-existing bug in the same `_get_alerts` loop: `problem.get` never actually returns a `status` text field (that's a UI concept, not an API field), so `STATUS_MAP.get(problem.pop("status", ""), AlertStatus.FIRING)` always fell through to `FIRING` regardless of the problem's real state. Status is now derived from the `acknowledged`/`suppressed` flags `problem.get` does return, the same pattern already used by other providers (Datadog, PagerDuty, Grafana, ...) to report these states from a pull. `problem.get` only returns still-open problems, so `resolved`/`ok` remain unreachable here, same as before.

## Test plan

- [x] Verified against a live production Zabbix instance: `problem.get` + `event.get` join now returns real hostnames for real open problems (confirmed several, e.g. host `ke-mba-icolo02-fw01a` for an OPNsense DHCP problem).
- [x] Verified `suppressed`/`acknowledged` flags on real problems now map to `SUPPRESSED`/`ACKNOWLEDGED` instead of always `FIRING`.
- [ ] Verify pushed webhook alerts for a multi-host trigger problem now include a host name
- [ ] No existing tests cover this provider; consider adding unit tests for `_get_alerts` and `_format_alert` covering these cases


shahargl (changes requested):
Can you just add the new fields and not remove the old one? For backward compatibility /older zabbix versions

talboren (dismissed):
lgtm

washosk:
Good catch  pushed a fix that keeps hostName/hostIp (and the even older HOST.NAME/HOST.IP macro keys from the pre-script mediatype webhook) as fallbacks alongside host_name/host_ip, so older installs that haven't re-run the webhook setup keep working.

washosk:
Correction to the earlier commit: problem.get does not support selectHosts (verified against a live instance — Zabbix rejects it with "Invalid parameter: unexpected parameter selectHosts", which broke every poll after deploying this branch). selectHosts is an event.get parameter, not problem.get.

Pushed a fix: _get_alerts now does a second batched event.get call on the fetched eventids (with selectHosts) and joins the hostname back in, instead of asking problem.get for something it doesn't support. This also adds the event.get scope requirement.

washosk:
Note: rolled the status bug I mentioned above into this same PR instead of a separate issue, since it's the same _get_alerts loop.

problem.get never returns a "status" text field (that's a UI-only concept), so the old STATUS_MAP.get(problem.pop("status", ""), AlertStatus.FIRING) always fell through to FIRING no matter the real state. Now deriving status from the acknowledged/suppressed flags problem.get actually returns — same approach Datadog/PagerDuty/Grafana providers already use to report these states on pull. Verified against real production problems: a suppressed=1 problem now correctly comes back SUPPRESSED instead of FIRING.

PR #6614, opened 2026-07-03
feat(jiraonprem): allow enabling TLS certificate verification

aditya-786:
## Summary

The jira-on-prem provider hardcodes `verify=False` on every HTTP call to the Jira server, so there is no way to enable TLS certificate verification even when the server presents a trusted certificate. Every request (auth, scope validation, issue create/update, board and sprint queries) silently skips verification, which exposes the personal access token and ticket data to man-in-the-middle interception.

This adds an optional `verify` switch to the provider's authentication config and threads it through all requests. It defaults to `false` to preserve the current behavior, so self-signed on-prem servers keep working unchanged; operators whose Jira presents a trusted certificate can now opt into verification.

## Test plan

`tests/test_jira_provider.py`:

- `test_jiraonprem_ssl_verification_disabled_by_default`: with no `verify` set, requests are issued with `verify=False`.
- `test_jiraonprem_ssl_verification_enabled_via_config`: with `verify: true`, requests are issued with `verify=True`.

The provider docs snippet was regenerated; `python scripts/docs_render_provider_snippets.py --validate` passes.


talboren (approved):
lgtm

Issue #6524, opened 2026-05-22
[➕ Feature]: Add a way delete alerts (database is growing forever)


zerint:
**Is your feature request related to a problem? Please describe.**
We were testing the use of the tool in the past couple of weeks. It seems like there is no way to completely delete alerts/workflow execution logs/etc in a way, where they are also removed from the database. Right now, it seems like the database is growing forever. This is not something, that's possible to maintain on a long term.

**Describe the solution you'd like**
Implement a way to delete alerts and other data that has the ability to grow the database significantly.

**Describe alternatives you've considered**
Deleting data from the database manually. This is wrong for obvious reasons. What happens if migrations update the structure in the future? What if something gets corrupted?

**Additional context**
We've been using KeepHQ for about a month and our postgres database is at ~20GB already.
(Regardless of how many alerts are sent in different environments, eventually every instance would grow out of a reasonable sized database.)

PR #6616, opened 2026-07-05
feat: time-based alert retention policy

### LINKED ISSUE #4802: [➕ Feature]: Time-based automatic alert deletion (retention policy)
Support automatic deletion of alerts based on age and status. Currently, alerts remain in the database indefinitely unless removed manually. It would be useful to define retention policies — for example:

- Delete `resolved` alerts after 7 days  
- Keep `firing` alerts for up to 30 days  

Allow configuring per-status retention policies (via config or UI). For example:  
- "Resolved alerts older than X days → auto-delete"  
- "Firing alerts older than Y days → retain"  
- Apply defaults if no config is given

**Additional context**
Useful in high-velocity environments to prevent unbounded alert growth. Could also help reduce DB size and improve system performance over time.

aditya-786:
## Summary

Closes #4802

Adds a time-based alert retention policy so the alert tables stop growing forever (#6524 reports a ~20GB postgres database after a month). When `KEEP_ALERT_RETENTION_DAYS` is set to a positive value, Keep periodically deletes alerts older than that many days, together with their incident links. The feature is disabled by default, so nothing changes for existing deployments.

This is a native, database-agnostic version of the MySQL cleanup CronJob referenced in the issue (helm-charts `db-cleanup-cronjob.yaml`), which does not work for postgres or sqlite deployments.

| Env var | Purpose | Default |
| --- | --- | --- |
| `KEEP_ALERT_RETENTION_DAYS` | Delete alerts older than this number of days (0 disables the feature) | 0 |
| `KEEP_ALERT_RETENTION_INTERVAL` | Seconds between retention runs | 86400 |
| `KEEP_ALERT_RETENTION_BATCH_SIZE` | Alerts deleted per batch | 1000 |

## How it works

- `delete_alerts_by_retention` in `keep/api/core/db.py` deletes in batches and commits per batch, so there are no long transactions and an interrupted run resumes cleanly on the next cycle. Deletion order per batch keeps foreign keys satisfied: `lastalerttoincident`, then `lastalert`, then `alerttoincident`, then `alert`.
- `LastAlert` rows are selected by `alert_id` rather than by timestamp: a fingerprint that is still active keeps its `LastAlert` (which points at a fresh alert) while its old `alert` rows are purged, and a fully stale fingerprint is removed entirely.
- `keep/api/tasks/process_retention_task.py` mirrors `process_watcher_task.py`: with Redis it runs through arq guarded by an NX lock, otherwise it runs as an asyncio loop guarded by a file lock. The arq cron is only registered when retention is enabled, since a destructive job should be opt-in.
- Per-status retention (the second part of the issue) needs alert status to be queryable at purge time and is left for a follow-up; this PR delivers the age-based policy the referenced cronjob already implements.

Two deliberate divergences from the cronjob, to keep behavior safe and the scope reviewable:

- `alerttogroup` is not touched because the table was dropped from the schema in 2024 and has no model anymore.
- Fingerprint-keyed tables (`alertenrichment`, `alertaudit`, `alertraw`, `alertfield`) are not purged. Enrichments are keyed by fingerprint and also serve incident and alert-instance enrichment, so deleting them based on alert age is not safe without more care.

## Tests

`tests/test_alert_retention.py`, running on the standard `db_session` fixture:

- purges old alerts across multiple batches (`batch_size=2` over 3 alerts) and removes their `LastAlert`, `AlertToIncident` and `LastAlertToIncident` rows while fresh alerts survive
- a fingerprint with old history but fresh activity keeps its `LastAlert` and latest alert while old rows are purged
- soft-deleted `LastAlertToIncident` rows are cleaned up for purged fingerprints
- no-op when nothing is older than the cutoff


aditya-786:
The patch coverage number is a measurement artifact rather than missing tests. The unit tests job runs `coverage run -m pytest -n auto`, and coverage.py does not follow pytest-xdist worker processes, so only lines executed in the controller process (imports and definitions) get recorded. That is why `delete_alerts_by_retention` shows 17 missing lines even though `tests/test_alert_retention.py` calls it directly in all 4 tests.

Reproduced both ways locally: with `-n auto` the report shows the same 17 missing lines, without it the new function reports 18 of 19 lines covered, the remaining one being the decorator line. If useful I can open a separate PR to capture worker coverage (pytest-cov, or COVERAGE_PROCESS_START with parallel mode), which would also make the project number reflect reality.

rsicart:
@aditya-786 @shahargl any updates on this topic? Looks really useful ^^.

PR #6622, opened 2026-07-08
fix(chore): slow bulk alert status changes and prevent duplicate submissions

### LINKED ISSUE #6623: [🐛 Bug]: Bulk "Change Status" action is slow and can be submitted twice
## Problem

When selecting multiple alerts in the Feed and using "Change Status" to bulk-update them, the operation takes noticeably long to complete. Because there's no loading indicator, impatient users click "Change Status" a second time before the first request finishes, which fires the request twice and results in two success toasts/disclaimers.

## Steps to reproduce

- Select multiple alerts in the Feed.
- Click "Change status", pick a new status, click "Change Status" in the dialog.
- Notice the delay before the dialog closes.
- Click "Change Status" twice in quick succession — two success toasts appear.

## Root cause

- Frontend: the modal's submit button has no in-flight/disabled state, so nothing prevents a second click while the first request is still pending (alert-change-status-modal.tsx).
- Backend: WorkflowManager.insert_events re-fetches and re-parses all of the tenant's workflows from the DB for every single alert being enriched (O(alerts × workflows)), which dominates request latency for larger batches.
- Backend: batch_enrich_alerts re-checks incident resolution and commits once per incident per alert, redundantly repeating work when multiple selected alerts belong to the same incident.

## Expected behavior

- The submit button should disable/show a loading state while the request is in flight, preventing duplicate submissions.
- Bulk status changes should scale with the number of alerts/workflows without redundant DB fetches, parsing, or commits.

ausias-armesto:
Fixes #6623 

- Add loading/disabled state to the "Change Status" modal button so an impatient double-click can't fire the request twice (which caused duplicate success toasts).
- Add missing "select a status" validation to the batch status-change handler.
- Fix O(alerts × workflows) blowup in WorkflowManager.insert_events: workflows were being re-fetched and re-parsed from the DB for every single alert being enriched. Now fetched/parsed once and reused across all events in the batch.
- Dedupe per-incident resolution checks in batch_enrich_alerts: previously ran an aggregate query and a session.commit() once per incident per alert, redundantly re-checking the same incident multiple times when several selected alerts belonged to it. Now each unique incident is checked once with a single commit at the end.


shahargl (changes requested):
can you fix tests?

shahargl (approved):
lgtm

PR #6632, opened 2026-07-14
feat(provisioning): provision correlation rules from KEEP_CORRELATION_RULES

### LINKED ISSUE #6563: Add Helm-native provisioning for correlation rules (parity with deduplication/mapping rules)
## Summary

The startup-time resource provisioner (`keep/api/config.py:provision_resources`) configures **providers**, **workflows**, **dashboards**, **deduplication rules**, and **mapping rules** from Helm values, but does NOT cover **correlation rules**. The only ways to create one today are the UI, the Python SDK, or `POST /rules` against the backend.

This issue requests symmetric, GitOps-friendly provisioning of correlation rules — `provision_correlation_rules_from_env` — so they can be declared in `values.yaml` next to the workflows that the same rules group.

## Why this matters

GitOps deployments (Helm + ArgoCD) want the entire alert→incident→ticket pipeline defined in code:

```yaml
keep:
  backend:
    provision:
      providers: { ... }      # ✅ provisioned
      workflows: [ ... ]      # ✅ provisioned
      dashboards: [ ... ]     # ✅ provisioned
      deduplicationRules:     # ✅ provisioned via env
      mappingRules:           # ✅ provisioned via env
      correlationRules: [ ... ]  # ❌ currently NOT provisioned
```

Today, declaring a `rules:` block under `keep.backend.provision` is silently dropped. Operators have to manually `POST /rules` after every cluster bootstrap — error-prone and breaks reproducibility.

## Proposed implementation

Mirror the existing `provision_deduplication_rules_from_env` and `provision_mapping_rules_from_env` shapes:

- Add a new env var (e.g. `KEEP_CORRELATION_RULES`) that accepts a JSON array of rule definitions matching the `POST /rules` schema (`ruleName`, `celQuery`, `sqlQuery`, `timeframeInSeconds`, `groupingCriteria`, `createOn`, `resolveOn`, `incidentNameTemplate`, `incidentPrefix`, `threshold`, `requireApprove`).
- In `provision_resources()`, call a new `provision_correlation_rules_from_env()` after the existing `provision_mapping_rules_from_env()` call.
- The new function should be idempotent: on each startup, upsert by `ruleName` (or by a stable ID exposed in the env spec) so rolling restarts don't create duplicates.

In the Helm chart, expose the same `correlationRules` key shape as `deduplicationRules` (already templated in `templates/backend-deployment.yaml`).

## Footgun to surface in docs

`cel-python` (the evaluator used for correlation rule `celQuery`) does NOT enable the regex `matches()` extension. Posting a rule with `name.matches('^Foo')` returns HTTP 200 but the next `GET /rules` surfaces `Method 'matches' not implemented` and the rule remains in a broken state. The same applies to workflow trigger CEL. Documenting this in the correlation-rules page (and ideally validating CEL at provision time) would prevent silent ingestion of broken rules.

## Use case backing this request

We deploy Keep via Helm + ArgoCD on EKS. We want a single `values.yaml` to define:

- The workflow that opens Jira tickets on critical alerts (`severity == "critical" && !name.startsWith("Data")`).
- The correlation rule that groups the same set of critical alerts into incidents, using the same CEL filter (for parity).

Today the workflow lives in Git; the rule lives only in the DB and we have to maintain a runbook (`POST /rules` curl) so the CEL parity isn't broken by a Git-only change. Native provisioning closes this gap.

## Related code

- `keep/api/config.py:provision_resources` — entry point that already covers the four other resource types.
- `keep/rulesengine/rulesengine.py` — `RulesEngine.create_or_update_rule` would be the natural upsert target.
- `examples/workflows/incident_example.yml` — workflow side, useful as a doc anchor for the parity story.

aditya-786:

The startup provisioner (`keep/api/config.py:provision_resources`) configures providers, workflows, dashboards, deduplication rules, and mapping rules from Helm values, but correlation (rules-engine) rules were left out. Today the only ways to create one are the UI, the SDK, or `POST /rules`, so a GitOps deployment cannot declare the alert grouping rule next to the workflow that uses the same CEL filter, and operators have to `POST /rules` after every cluster bootstrap.

This adds `provision_correlation_rules_from_env`, giving correlation rules the same env-driven provisioning the other resources already have. Closes #6563.

## What changed

- **`keep/api/bl/correlation_rules_provisioning.py`** (new): `provision_correlation_rules_from_env(tenant_id)`, mirroring `provision_deduplication_rules_from_env`. `KEEP_CORRELATION_RULES` holds a JSON array (or a path to a `.json` file, dispatched by the same regex the dedup provisioner uses) of rule specs in the `POST /rules` schema (`ruleName`, `celQuery`, `sqlQuery`, `timeframeInSeconds`, `timeUnit`, `groupingCriteria`, `createOn`, `resolveOn`, `incidentNameTemplate`, `incidentPrefix`, `threshold`, `requireApprove`).
- On every startup the DB is reconciled to the env: upsert by `ruleName`, delete provisioned rules that are no longer present, and never touch UI-created rules (`is_provisioned=False`). Unsetting the env var deprovisions everything the provisioner owns. Rolling restarts are idempotent.
- Each `celQuery` is compiled at provision time; a spec that is missing a required field or has an unparseable CEL is logged and skipped, so one bad rule does not block the others.
- **`is_provisioned` column on `Rule`** + Alembic migration `b3f1c9a24d70` (on the current head `67ff7efffed4`), tracking which rules the provisioner owns. Mirrors the flag added to `MappingRule`.
- **`db.create_rule`** gains an `is_provisioned=False` parameter (same shape `create_deduplication_rule` already has).
- **`provision_resources()`** calls the new function right after `provision_mapping_rules_from_env`.

Example `values.yaml`:

```yaml
keep:
  backend:
    env:
      - name: KEEP_CORRELATION_RULES
        value: |
          [
            {
              "ruleName": "critical-alerts-correlation",
              "celQuery": "severity == \"critical\" && !name.startsWith(\"Data\")",
              "sqlQuery": { "sql": "severity = :severity", "params": { "severity": "critical" } },
              "timeframeInSeconds": 600,
              "timeUnit": "seconds",
              "createOn": "any",
              "resolveOn": "never",
              "threshold": 1
            }
          ]
```

## Notes on scope

- The Helm chart lives in the separate `keephq/helm-charts` repo, so exposing a `correlationRules` key in the chart templates is a companion change there; this PR is the backend that reads the env var (same split as the existing dedup/mapping provisioning).
- On the `matches()` footgun the issue raises: the pinned `cel-python` in this repo does evaluate `name.matches('^Foo')` (returns a bool), so that specific case does not reproduce here. Provision-time validation therefore compiles the CEL to catch genuine parse errors rather than attempting a trial evaluation, which would false-negative on any rule referencing fields absent from a synthetic payload.

## Checks

- Testing Strategy
  - [x] Unit tests

`tests/test_correlation_rules_provisioning.py` (11 tests, mirroring the mapping/dedup provisioning tests): create, multiple, idempotent, update-in-place, deprovision-on-removal, deprovision-on-unset, leave-UI-rules-untouched, skip-invalid-CEL, skip-missing-field, load-from-.json-file, and clean no-op. All pass locally; the rules-engine suite (`tests/test_rules_engine.py`) stays green.

### DIFF
--- a/keep/api/bl/correlation_rules_provisioning.py
+++ b/keep/api/bl/correlation_rules_provisioning.py
@@ -1,0 +1,210 @@
+import json
+import logging
+import re
+
+import celpy
+
+import keep.api.core.db as db
+from keep.api.core.config import config
+from keep.api.models.db.rule import CreateIncidentOn, ResolveOn
+
+logger = logging.getLogger(__name__)
+
+KEEP_CORRELATION_RULES_ENV_VAR = "KEEP_CORRELATION_RULES"
+SYSTEM_ACTOR = "system"
+
+_REQUIRED_FIELDS = ("ruleName", "celQuery", "sqlQuery", "timeframeInSeconds", "timeUnit")
+_CEL_ENV = celpy.Environment()
+
+
+def provision_correlation_rules_from_env(tenant_id: str):
+    """Provision correlation (rules-engine) rules from ``KEEP_CORRELATION_RULES``.
+
+    Mirrors ``provision_deduplication_rules_from_env``: the env var holds a JSON
+    array (or a path to a ``.json`` file) of rule specs matching the ``POST /rules``
+    schema. On every startup the DB is reconciled to the env:
+
+      - Provisioned rules whose ``ruleName`` is no longer in the env are deleted.
+      - A provisioned rule with a matching name is updated in place.
+      - A new provisioned rule is created (``is_provisioned=True``).
+      - UI-created rules (``is_provisioned=False``) are never touched.
+
+    If the env var is unset, any currently-provisioned rules are deprovisioned.
+    A spec that fails validation (missing field or an unparseable ``celQuery``)
+    is logged and skipped so one bad rule does not block the others.
+    """
+    rules_to_provision = get_correlation_rules_to_provision()
+
+    provisioned_rules = [rule for rule in db.get_rules(tenant_id) if rule.is_provisioned]
+
+    if not rules_to_provision:
+        if provisioned_rules:
+            logger.info(
+                "%s unset; deprovisioning %d existing correlation rule(s)",
+                KEEP_CORRELATION_RULES_ENV_VAR,
+                len(provisioned_rules),
+            )
+            for rule in provisioned_rules:
+                db.delete_rule(tenant_id=tenant_id, rule_id=str(rule.id))
+        else:
+            logger.info(
+                "No correlation rules to provision and none currently provisioned"
+            )
+        return
+
+    provisioned_rules_by_name = {rule.name: rule for rule in provisioned_rules}
+
+    # delete provisioned rules that are no longer in the env
+    for rule in provisioned_rules:
+        if str(rule.name) not in rules_to_provision:
+            logger.info(
+                "Correlation rule with name '%s' is not in the env, deleting from DB",
+                rule.name,
+            )
+            db.delete_rule(tenant_id=tenant_id, rule_id=str(rule.id))
+
+    for rule_name, rule_to_provision in rules_to_provision.items():
+        try:
+            _validate_rule_to_provision(rule_name, rule_to_provision)
+        except ValueError as exc:
+            logger.warning(
+                "Skipping invalid correlation rule '%s': %s", rule_name, exc
+            )
+            continue
+
+        existing_rule = provisioned_rules_by_name.get(rule_name)
+        if existing_rule is not None:
+            logger.info(
+                "Correlation rule with name '%s' already exists, updating in DB",
+                rule_name,
+            )
+            _update_correlation_rule(
+                tenant_id, str(existing_rule.id), rule_name, rule_to_provision
+            )
+            continue
+
+        logger.info(
+            "Correlation rule with name '%s' does not exist, creating in DB",
+            rule_name,
+        )
+        _create_correlation_rule(tenant_id, rule_name, rule_to_provision)
+
+
+def get_correlation_rules_to_provision() -> dict[str, dict]:
+    """Read correlation rules from ``KEEP_CORRELATION_RULES`` as a dict keyed by name.
+
+    The env var is either an absolute/relative path to a ``.json`` file or a JSON
+    string. Its content is a JSON array of rule specs matching the ``POST /rules``
+    schema (a single object is also accepted). Specs without a ``ruleName`` are
+    skipped.
+    """
+    rules_from_env_var = config(key=KEEP_CORRELATION_RULES_ENV_VAR, default=None)
+
+    if not rules_from_env_var:
+        return None
+
+    if re.compile(r"^(\/|\.\/|\.\.\/).*\.json$").match(rules_from_env_var):
+        with open(file=rules_from_env_var, mode="r", encoding="utf8") as file:
+            try:
+                parsed = json.loads(file.read())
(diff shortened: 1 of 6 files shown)


apercivall:
We reviewed this PR, ran its tests, and deployed it on 0.54.3 in a self-hosted single-tenant install, where it does what it says — correlation rules that previously existed only as a database row are now declared in configuration and reconciled on every start. Thanks for writing it.

Some findings from reviewing and running it, in case they help it move.

**The red checks are not this PR.** All three failures are `tests/e2e_tests/incidents_alerts_tests/test_filtering_sort_search_on_alerts.py` — a Playwright strict-mode locator violation (`[data-testid='facet']` matching two elements) and a 10s `wait_for_selector` timeout. Unrelated to provisioning; `unit-tests` and `integration-tests` pass. A rebase would probably clear them.

**`groupDescription` is applied at creation and silently dropped on update.** `_create_correlation_rule` passes it; `_update_correlation_rule` cannot, because `db.update_rule` has no such parameter — and the `PUT /rules` route never sends one either, so this is pre-existing and the PR inherits it. The effect is that editing `groupDescription` in `KEEP_CORRELATION_RULES` applies the first time the rule is created and is ignored ever after, which is the failure mode provisioning is meant to remove.

What worked for us was giving `update_rule` a `group_description` parameter defaulting to a sentinel rather than `None`:

```python
_UNCHANGED = object()

def update_rule(..., assignee=None, group_description=_UNCHANGED):
    ...
    if group_description is not _UNCHANGED:
        rule.group_description = group_description
```

The sentinel matters: with a plain `=None` default, the route — which does not pass the field — would blank the description on every edit made in the UI.

**A `threshold` of 0 aborts startup provisioning for every rule after it.** `_validate_rule_to_provision` does not check it, `Rule.threshold` carries `CheckConstraint("threshold>0")`, and the failing commit happens outside the per-rule `try`, so it propagates out of `provision_resources()`. The `POST /rules` handler rejects 0 with its own check; `RuleCreateDto` does not. Suggest:

```python
threshold = rule_to_provision.get("threshold", 1)
if isinstance(threshold, bool) or not isinstance(threshold, int) or threshold <= 0:
    raise ValueError("'threshold' must be a positive integer")
```

`bool` excluded explicitly since `True` is an `int` in Python and would pass a naive `> 0`.

**Docstring vs. behaviour on paths.** `get_correlation_rules_to_provision` documents "an absolute/relative path to a `.json` file", but the regex — inherited from the deduplication provisioner — only matches `/`, `./` and `../` prefixes. A bare `config/rules.json` falls through to `json.loads` and raises during startup. Worth either widening the check or narrowing the docstring; the same is true of the deduplication provisioner it was copied from.

**One adoption note for anyone deploying this**, which might be worth a line in the PR description: the provisioner only manages rules with `is_provisioned = true`, so an existing UI-created rule is not adopted by a matching `ruleName` — you get a second rule matching the same alerts. And the column does not exist until the migration has run, so the flag cannot be set beforehand. We are rolling it out in two steps for that reason: deploy with the env var unset (verified a clean no-op — `No correlation rules to provision and none currently provisioned`), then set the flag on the pre-existing row so the provisioner updates it in place rather than creating a twin, then add the declaration.

The migration applied cleanly on the current single head (`67ff7efffed4`) against PostgreSQL, and the full non-integration suite passes with the two changes above plus tests for them.

Issue #3931, opened 2025-03-08
[➕ Feature]: Physical Topology


mikeshaw57:
**Is your feature request related to a problem? Please describe.**

Our primary use case of Keep is for device alert correlation of a modest size physical network. Mostly routers, switches, etc.
We employ multiple types of monitoring tools, including event based (i.e. syslog/graylog) and polling-type (nagios, SNMP, etc.)
I would like Keep to use knowledge from NetBox to enrich alerts with metadata, represent relations in topology views, and provide context to AI for advanced correlation.
Reducing alert noise for larger incidents (i.e. rack power loss) from multiple monitoring tools, specifically polling alerts, has been difficult. I.e. a Nagios config might have generated "monitor services" for a device and each port/interface which is triggered every 5min. If a rack looses power (rate, but it's happened), this might trigger 100's of alerts from nagios for every device and monitored interface in the rack. Correlating these into a single incident would be ideal.

**Describe the solution you'd like**

I would like Keep to use knowledge from NetBox to enrich alerts with metadata, represent relations in topology views, apply topology correlation, and provide context to AI for advanced correlation.
Primarily to apply topology correlation.

We can currently enrich alerts with workflows to obtain information from NetBox. This is very useful.

Topology Correlation, according to the docs "helps correlate alerts based on your infrastructure’s topology" and "It automatically analyzes incoming alerts and their relationship to your infrastructure topology, creating incidents when multiple related services or components of an application are affected". I'd like topology correlation to creating incidents when multiple related NetBox DCIM objects are affected AND provide node graphs in the topology viewer for alerts/incidents. I.e. if multiple alerts (nagios, graylog, etc) are received for a specific device and multiple interfaces in a short timeframe, correlate them to a single incident. Or if many devices and device-interface alerts are received that correspond to a shared rack, correlate these. Likewise for a devices/interfaces under a site/location.

Having the NetBox plugin (or a more advanced one) that could feed the Topology Processor would be highly valuable in correlating events, especially for polling-based alerts. 
Additionally, representing this visually in a topology viewer would be highly insightful for engineers/operators. I.e. rack-x: [device-a...device-n]; device-a-interface-x -> line-axby -> device-b-interface-y

I could generate CSV or YAML to represent this, but our infrastructure and netbox data change fewquently, so a live connection would be preferred.
I don't think Keep's "Service Topology" is a best fit for this. Perhaps an additional "Infrastructure Topology" or being able to create multiple categorized topologies. 
[...]

PR #6635, opened 2026-07-17
feat(netbox): pull DCIM topology into the service topology map

aditya-786:

Towards #3931. Users monitoring physical networks (routers, switches) get alerts from tools like Nagios and Graylog, and when a shared failure hits (the issue describes a rack power loss) hundreds of alerts arrive with no way to correlate them by the physical infrastructure they belong to. NetBox is the source of truth for that infrastructure, but the Keep NetBox provider only received webhook alerts.

This upgrades the NetBox provider to a topology provider, so Keep pulls the DCIM inventory and feeds it to the existing topology map and topology based correlation:

- **Devices become services**: name (the attribute monitoring tools report alerts with), primary IP, role, manufacturer, site and tags, using the existing `TopologyService` columns. No migration needed.
- **Cables become dependencies**: a cable between two known devices creates a service dependency, so the topology map draws the physical connections. Both the current (`a_terminations`/`b_terminations`) and pre-3.3 (`termination_a`/`termination_b`) cable formats are supported.
- **Racks optionally become applications** (off by default, opt-in via `group_racks_as_applications`): devices sharing a rack are grouped into an application named after the rack. With the topology processor enabled (`KEEP_TOPOLOGY_PROCESSOR=true`), alerts from many devices in the same rack can then be correlated into a single incident, which is the rack power loss scenario from the issue. Application semantics for a physical container is a design choice, which is why it sits behind a flag.

Rack applications use a stable id derived from tenant, provider and rack, so re-pulls update the application in place instead of creating a new one on every pull.

## Backward compatibility

The API access fields are all optional and existing webhook-only installs are unaffected:

- With no URL or token configured, `pull_topology` is a silent no-op and scope validation passes, so the topology polling loop does not produce errors for legacy installs.
- The webhook alert path (`_format_alert`) is unchanged.
- On configured installs an API failure raises, so a transient NetBox outage never wipes previously pulled topology (`process_topology` only rewrites on non-empty data).

## Scale

Pulling paginates the NetBox API (`limit`/`offset`, 1000 per page) with a hard cap (`max_devices`, default 5000, truncation logged) and excludes `config_context` from device payloads to keep responses small. Every request carries a timeout, and the SSL verify switch matches the other on-prem providers.

## Out of scope (kept for follow-ups)

- No changes to the topology processor or correlation logic.
- No new "infrastructure topology" category or UI work; the existing topology map renders the pulled data as is.
- No AI context integration.

## Checks

- Testing Strategy
  - [x] Unit tests

`tests/providers/netbox_provider/test_netbox_topology.py` (17 tests): device and cable mapping (including the legacy cable format and dangling cable ends), rack grouping on and off with stable namespaced application ids, webhook-only no-op with no HTTP calls, pagination, `max_devices` cap, API error raising, verify wiring, the unchanged webhook alert format, and an end-to-end test through `process_topology` proving a re-pull is idempotent (services recreated, the rack application updated in place). Existing topology tests (`tests/test_topology.py`) stay green, and the docs snippet was regenerated with the provider docs updated.

### DIFF
--- a/docs/providers/documentation/netbox-provider.mdx
+++ b/docs/providers/documentation/netbox-provider.mdx
@@ -81,6 +81,16 @@ X-API-KEY: your-api-key
 
 Now, you have successfully connected NetBox to Keep. You will start receiving the events in Keep based on the event rules you have created.
 
+## Pulling topology from NetBox
+
+Besides receiving alerts, Keep can pull your physical infrastructure from the NetBox DCIM API and show it on the service topology map. Devices become services (with their primary IP, role, manufacturer and site), and cables between devices become dependencies.
+
+To enable it, set the `NetBox instance URL` and a `NetBox API token` in the provider configuration. A read-only token is enough. Without these fields the provider keeps working in webhook-only mode.
+
+Optionally, enable `Group devices of the same rack into an application` so that devices sharing a rack are grouped into an application named after the rack. Combined with Keep's topology-based correlation, this lets alerts from many devices in the same rack (for example after a rack power loss) be correlated into a single incident.
+
+The number of pulled objects is limited by `Maximum number of devices to pull` (default 5000).
+
 ## Useful Links
 
 - [NetBox](https://netboxlabs.com/)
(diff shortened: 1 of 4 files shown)


shahargl (approved):
lgtm

PR #6638, opened 2026-07-19
fix(db): refresh user after role update to avoid DetachedInstanceError

### LINKED ISSUE #4296: [🐛 Bug]: Error "Failed to update user" when editing user
**Describe the bug**
I am using DB Authentication. When trying to change a user’s role, I get the error: “Failed to update user.”

**To Reproduce**

1. Go to 'settings'
2. Сreate user
3. Click on the user and change the role
4. Click on the save button

**Expected behavior**
The user was successfully saved with the new role

**Screenshots**
Image

mvanhorn:
## Summary

Under DB authentication, editing a user's role in Settings fails with "Failed to update user". `update_user_role` committed the role change but never refreshed the ORM object, so the returned `User` was detached with expired attributes and building the response raised `DetachedInstanceError` -> HTTP 500. This mirrors the session handling its sibling `update_user_password` already uses.

## Background

roma55443 reported the "Failed to update user" error in #4296. A later 501 "Updating users is not supported by this identity manager" was a separate symptom already fixed by #6609 (which added `DbIdentityManager.update_user`); the role-change 500 is the live remaining defect. In `keep/api/core/db.py`, `update_user_role` runs inside `with Session(engine)` (default `expire_on_commit=True`) and returns `user` after `session.commit()`, but unlike `update_user_password` in the same file it never calls `session.refresh(user)`. Once the session block closes, the object is detached and its attributes are expired, so `update_user` raises `DetachedInstanceError` when it reads `updated_user.username` to build the response.

The fix adds `session.refresh(user)` after the role-changing commit and an early `return None` for a missing user, matching `update_user_password` exactly. Added regression tests to `tests/test_change_password.py` covering a changed role, an unchanged role, and a missing user, alongside the existing DB-auth update tests.

Closes #4296

### DIFF
--- a/keep/api/core/db.py
+++ b/keep/api/core/db.py
@@ -2139,10 +2139,13 @@ def update_user_role(tenant_id, username, role):
             .where(User.tenant_id == tenant_id)
             .where(User.username == username)
         ).first()
-        if user and user.role != role:
+        if not user:
+            return None
+        if user.role != role:
             user.role = role
             session.add(user)
             session.commit()
+            session.refresh(user)
     return user
 
 
--- a/tests/e2e_tests/incidents_alerts_tests/test_filtering_sort_search_on_alerts.py
+++ b/tests/e2e_tests/incidents_alerts_tests/test_filtering_sort_search_on_alerts.py
@@ -66,7 +66,9 @@ def init_test(browser: Page, alerts, max_retries=3):
             else:
                 raise e
 
-    browser.wait_for_selector("[data-testid='facet-value']", timeout=10000)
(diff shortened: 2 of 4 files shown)


shahargl:
hey @mvanhorn , thanks for adding this PR. mind signing CLA?

also - are you the founder of Lyft? :O

mvanhorn:
Thanks for the quick merge! CLA is signed - the assistant badge flipped green shortly after your comment.

mvanhorn:
Thanks once more for the merge itself, @shahargl - refreshing the user after role updates puts DetachedInstanceError to rest.

PR #6646, opened 2026-07-25
feat(discord): feature-complete webhook provider with embeds, base64 file attachments, edit, rate-limit retry, host restriction, and guardrails

### LINKED ISSUE #6645: [🔌 Provider]: Feature-complete Discord messaging: embeds, file attachments, full webhook payload support
## Describe the current state

The Discord provider hardcodes the JSON body to only `content` and `components`. The Discord Execute Webhook API supports a much richer payload, but none of it is surfaced:

- **JSON-only params** (could be passed through but aren't): `embeds`, `username`, `avatar_url`, `allowed_mentions`, `flags`, `thread_name`, `applied_tags`, `poll`, `attachments` (metadata)
- **Query string params** (not supported): `wait`, `thread_id`, `with_components`
- **File attachments** (requires `multipart/form-data`, not JSON; needs special handling): `files[n]` + `payload_json`
- **Message editing** (requires a different endpoint: `PATCH .../messages/{id}`)
- **No webhook validation, no rate-limit handling, no host restriction on `webhook_url`**

Ref: https://discord.com/developers/docs/resources/webhook#execute-webhook

## Proposed changes

1. **Generic JSON passthrough**: Accept `**kwargs` in `_notify()` and merge into the JSON body. Covers all JSON-only params (`embeds`, `username`, `avatar_url`, `allowed_mentions`, etc.) without explicit parameters. `tts` is deliberately excluded (dropped with a warning) - marginal value for incident notifications, not worth the surface area.

2. **File attachment support via base64**: Add a `files` parameter accepting a list of `bytes`, `(filename, content, mime_type)` tuples, or `{"base64": ..., "filename": ...}` dicts. When files are present, switch to `multipart/form-data`: JSON body goes into `payload_json`, file content into `files[0]`, `files[1]`, etc.
   - **No filesystem path form is supported, by design.** Discord notifications execute in the same shared, long-running process as every other tenant's workflows. A "read this path" parameter driven by workflow templating (which can embed externally-influenced alert/event data) would be an arbitrary local file read/exfiltration primitive. Base64 keeps the file entirely in-memory and flows through Keep's existing step-output/templating mechanism (`{{ steps.<name>.results.<key> }}`) - e.g. a chart-rendering step returns `{"chart_base64": "..."}` and a later Discord step attaches it directly.

3. **Query string params**: Extract `wait`, `thread_id`, and `with_components` from kwargs and pass as URL query params.

4. **Message editing**: A `message_id` param switches the request to `PATCH .../messages/{id}` instead of posting a new message (validated as a Discord snowflake - digits only - before being interpolated into the URL). Obtain the id from a prior `wait: true` send via `result["message_id"]`. (Message deletion was considered and deliberately left out - editing to mark a message resolved/false-positive preserves the incident timeline better than deleting it, which is what actually matters for postmortems.)

5. **`validate_scopes()`**: Confirms the webhook is real and reachable via a `GET` request at configuration time, before any workflow depends on it - catching typos/revoked webhooks before a real incident, not during one.

6. **Rate-limit handling**: Retries (bounded, 3 attempts) on `429`, honoring Discord's `Retry-After` value, clamped to a sane maximum (30s) so a malformed or hostile response can't block a worker thread indefinitely.

7. **Client-side guardrails**: Validates Discord's documented hard limits before sending (content ≤2000 chars, ≤10 embeds, ≤10 files, ≤10 MiB per file) with a clear error instead of a round-trip 400.

8. **`webhook_url` host restriction**: Only `discord.com`/`discordapp.com` accepted - previously any HTTPS URL passed validation, which combined with generic JSON/attachment passthrough made this provider usable as a generic outbound-request primitive to an arbitrary destination.

9. **Backward compatible**: Calls using only `content` and/or `components` are unaffected.

## Additional context

The provider already works for basic text and interactive component messages. This became a broader hardening pass once the file-attachment design was scrutinized for security implications in a shared, multi-tenant execution model - the scope grew from "surface the full webhook payload" to also include the guardrails above.

aries1980:
## Summary

Feature-completes the Discord webhook provider against the Execute Webhook + Edit Message APIs:

- **Generic JSON passthrough** — `embeds`, `username`, `avatar_url`, `allowed_mentions`, `flags`, `thread_name`, `applied_tags`, `poll`, `attachments` forwarded as-is
- **File attachments via base64** — `files` param via multipart/form-data; no filesystem paths supported (security rationale in code/docs); accepts `bytes`, `(filename, content, content_type)` tuples, or `{"base64": "...", "filename": "..."}` dicts
- **Message editing** — `message_id` param switches to `PATCH .../messages/{id}`; snowflake-validated
- **Rate-limit retry** — bounded 3-attempt retry on 429 with Retry-After (clamped to 30s max)
- **`webhook_url` host restriction** — only `discord.com`/`discordapp.com` accepted
- **`validate_scopes()`** — GET on the webhook URL to confirm it is real and reachable
- **Client-side guardrails** — content ≤2000 chars, ≤10 embeds, ≤10 files, ≤10 MiB per file
- **`tts` and `delete` deliberately dropped** — marginal value, not worth surface area

### Documentation updated
- Provider MDX page, auto-generated snippet, README, example workflow

Closes #6645


aries1980:
For the `ruff`-related errors, I created a mega PR: https://github.com/keephq/keep/pull/6649

aries1980:
@shahargl : Hi, after the rebase, the issues went away, I think it is ready for review.

PR #6654, opened 2026-07-28
ci: read the PR description from env instead of interpolating it into the shell

### LINKED ISSUE #6655: [🐛 Bug]: PR description is interpolated into the shell in auto-resolve-keep.yml
**Describe the bug**

In `.github/workflows/auto-resolve-keep.yml`, the pull request description is interpolated directly
into a shell script:

```yaml
run: |
  PR_DESC="${{ github.event.pull_request.body }}"
  INCIDENT_ID=$(echo "$PR_DESC" | grep -ioP 'close keep incident:\s*\K[a-f0-9-]+' || true)
```

GitHub Actions expands `${{ ... }}` into the script *text* before bash ever runs, so the description
becomes part of the script rather than a string assigned to `PR_DESC`. A pull request body is free text
chosen by whoever opens the pull request, and `$(...)` and backticks still evaluate inside double
quotes — so a body containing a command substitution would run on the runner instead of being searched
for `close keep incident: <id>`.

The `grep -ioP ... [a-f0-9-]+` pattern is tight about what it accepts as an ID, which is good. It just
runs *after* the value has already become part of the command.

**To Reproduce**

I have not run this against your repository — it would mean executing a command on your runner — so
this is read from the workflow file rather than demonstrated. To confirm it safely on a fork:

1. Fork the repository.
2. Open a pull request whose description contains a harmless command substitution, for example
   `$(echo marker-12345)`.
3. Open the `auto-resolve-keep` job log and look at the rendered `run:` block for that step.
4. If `marker-12345` appears in the script or its output, the body was evaluated rather than treated
   as data.

**Expected behavior**

The description should be handled as data. Binding it to an environment variable does that, because
`env:` values are passed to the process rather than substituted into the script text:

```yaml
env:
  PR_BODY: ${{ github.event.pull_request.body }}
run: |
  PR_DESC="$PR_BODY"
```

**Additional context**

On scope, so this is not overstated: the workflow uses the `pull_request` trigger, so a fork's pull
request gets a read-only token and no secrets. This is hardening rather than a live path to your
credentials, and I would not want it read as more urgent than it is.

Fix proposed in #6654 — both `grep` extractions and both `$GITHUB_OUTPUT` writes are untouched, so
incident IDs and alert fingerprints are found exactly as they are today. Two lines added, one changed.

Opening this because @shahargl asked for an issue to link the pull request to.

*Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow
myself.*

kobihikri:
Hi, and thanks for Keep.

In `.github/workflows/auto-resolve-keep.yml`, the PR description is read into the shell by interpolation:

```yaml
run: |
  PR_DESC="${{ github.event.pull_request.body }}"
  INCIDENT_ID=$(echo "$PR_DESC" | grep -ioP 'close keep incident:\s*\K[a-f0-9-]+' || true)
```

Actions expands `${{ ... }}` into the script text before bash runs, so the description becomes part of the script rather than a string assigned to `PR_DESC`. A PR body is free text chosen by whoever opens the PR, and `$(...)` and backticks still run inside double quotes — so a body containing `$(...)` would execute on the runner instead of being searched for `close keep incident: <id>`.

The `grep -ioP ... [a-f0-9-]+` pattern is nicely tight about what it will accept as an ID, which is good — it just runs after the value has already been part of the command.

The change binds the body to an environment variable:

```yaml
env:
  PR_BODY: ${{ github.event.pull_request.body }}
run: |
  PR_DESC="$PR_BODY"
```

Both `grep` extractions and both `$GITHUB_OUTPUT` writes are untouched, so incident IDs and alert fingerprints are found exactly as they are today. Two lines added, one changed.

On scope: this is a `pull_request` trigger, so a fork PR gets a read-only token and no secrets — hardening rather than a live exploit.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow myself.

Fixes #6655

### DIFF
--- a/.github/workflows/auto-resolve-keep.yml
+++ b/.github/workflows/auto-resolve-keep.yml
@@ -28,8 +28,10 @@ jobs:
       - name: Extract Keep ID from PR description
         if: github.event_name == 'pull_request'
         id: extract_id
+        env:
+          PR_BODY: ${{ github.event.pull_request.body }}
         run: |
-          PR_DESC="${{ github.event.pull_request.body }}"
+          PR_DESC="$PR_BODY"
           INCIDENT_ID=$(echo "$PR_DESC" | grep -ioP 'close keep incident:\s*\K[a-f0-9-]+' || true)
           ALERT_FINGERPRINT=$(echo "$PR_DESC" | grep -ioP 'close keep alert:\s*\K[a-f0-9-]+' || true)
           echo "incident_id=$INCIDENT_ID" >> $GITHUB_OUTPUT


shahargl:
@kobihikri thanks for the PR, can u sign the CLA and also open an issue and link to it so I can apprvove?

kobihikri:
Thanks @shahargl. The CLA is already signed — CLAassistant confirmed it on the PR yesterday, so I think that one is covered.

I have opened #6655 describing the issue and linked it from the PR description with `Fixes #6655`, which should also satisfy the *Validate PR to Issue link* check.

One thing I put in the issue and want to repeat here so it is not oversold: this workflow uses the `pull_request` trigger, so a fork's PR gets a read-only token and no secrets. It is hardening rather than a live path to your credentials.

shahargl (approved):
lgtm

PR #6659, opened 2026-07-30
fix: handle quoted alert JSON in workflow functions

### LINKED ISSUE #5302: [🐛 Bug]: Keep keep.dictget() function parsing issue
**Describe the bug**
When a workflow using `keep.dictget('{{alert}}', 'some_field', 'unknown')` function is run on the alert with `description="Hello ('Hola') everyone. How are you all doing?"`, the workflow fails. Tried using double quotes around the dictget params but it didn't work.

**To Reproduce**

1. Add an alert with the mentioned description
2. Add a workflow with the following configuration
steps:
    - name: console-step
      provider:
        type: console
        config: "{{ providers.default-console }}"
        with:
          message: keep.dictget('{{alert}}', 'some_field', 'unknown')
3. Execute the workflow on the inserted alert

**Expected behavior**
The workflow should run successfully.

**Screenshots**

Image

Image

Root cause:
The issue is originating from here: https://github.com/keephq/keep/blob/main/keep/iohandler/iohandler.py#L96
The substring ') is causing the issue because when it processes the quote, keep considers it as terminating character.

aditya-786:
## Summary

- preserve rendered alert JSON as one quoted workflow-function argument
- apply quote escapes using token-relative offsets for prefixed and repeated calls
- cover single and double delimiters without changing non-JSON string parsing

## Testing

- `SKIP_DOCKER=true pytest tests/test_iohandler.py tests/test_functions.py -o addopts='' -q` (190 passed)
- `ruff check keep/iohandler/iohandler.py tests/test_iohandler.py`

Closes #5302

### DIFF
--- a/keep/iohandler/iohandler.py
+++ b/keep/iohandler/iohandler.py
@@ -129,16 +129,27 @@ class IOHandler:
                     escape_next = False
                     quote_char = ""
                     escapes = {}
+                    embedded_json_end = None
+                    string_content_started = False
                     while i < len(text) and (parent_count > 0 or in_string):
                         if text[i] == "\\" and in_string and not escape_next:
                             escape_next = True
                             i += 1
                             continue
+                        elif (
+                            in_string
+                            and embedded_json_end is not None
+                            and i < embedded_json_end
+                        ):
+                            if text[i] == quote_char and not escape_next:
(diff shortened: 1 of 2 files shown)


shahargl (approved):
lgtm

PR #6661, opened 2026-08-02
fix: widen alert table column resize hit target

### LINKED ISSUE #6660: Column resize handle in alerts table has a near-invisible hit target
In the alerts table (e.g. Presets view), dragging a column header's edge to resize it rarely works. The resize handle is a permanently 2px-wide element that only grows on its own `:hover` — since a user can't hover it without already being on the 2px sliver, most drag attempts land just outside it and silently do nothing.

ausias-armesto:
Widens the column-resize hit target in the alerts table header from a hover-dependent 2px sliver to a constant 8px area centered on the column boundary, and adds `onTouchStart` alongside the existing `onMouseDown` for touch parity.

Fixes #6660

**Test plan:** drag a column border in the alerts table (e.g. Presets → Feed) and confirm the width changes live.

### DIFF
--- a/keep-ui/widgets/alerts-table/ui/alert-table-headers.tsx
+++ b/keep-ui/widgets/alerts-table/ui/alert-table-headers.tsx
@@ -406,13 +406,15 @@ const DraggableHeaderCell = ({
       {column.getIsPinned() === false && (
         <div
           className={clsx(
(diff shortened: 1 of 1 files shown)


shahargl (approved):
Lgtm

PR #6664, opened 2026-08-03
fix: rightmost column cannot be moved in preset view despite saved config (Fixes #5118)

### LINKED ISSUE #5118: [🐛 Bug]: Rightmost Column in Preset Cannot Be Moved Despite "Configuration Saved" Message
**Describe the bug**
In the preset view, the rightmost column cannot be repositioned using either drag-and-drop or the column menu options. Even though the UI shows "Column configuration is saved", the column remains in place. Additionally, the column just before the rightmost one incorrectly lacks a "Move column right" option, as if it were already the last column—despite another column being present to its right.

**To Reproduce**

1. Open a saved preset that includes at least three columns.
2. Try to rearrange the rightmost column:
   - Click the menu (three dots) on that column header.
   - Choose "Move column left".
   - Or attempt to drag the column.
3. Observe:
   - The toast message "Column configuration is saved" appears.
   - The column does not move visually in the UI.
4. Try the same steps on the second-to-last column:
   - Notice it has no "Move column right" option even though there's a column after it.
5. Repeat similar actions in the central feed view — the issue does not occur there.

**Expected behavior**
The rightmost column should be movable like all other columns, and the column next to it should correctly reflect available actions. The "configuration saved" message should only appear when changes are actually applied

**Screenshots**

**Additional context**
version 0.45.9

waterWang:
## Background

When using the preset view, the rightmost data column (before the `alertMenu` pinned column) cannot be moved. Two symptoms were reported:

1. **Rightmost column**: clicking "Move column left" shows "Column configuration saved" toast, but the column does not move visually.
2. **Second-to-last column**: incorrectly lacks a "Move column right" option.

## Root Cause

Two separate bugs were found:

### Bug 1: `isRightmostColumn()` used wrong column source

The function checked `table.getVisibleLeafColumns()` (the rendered column order) instead of the `columnOrder` prop (the actual state). In preset views, these can differ, causing the wrong column to be identified as the rightmost.

**Fix**: Filter `columnOrder` by visible columns (excluding `alertMenu`) to determine the rightmost data column.

### Bug 2: Local state not updated after backend API call

In `updateMultipleColumnConfigs()`, after a successful backend API call (`updateColumnConfig`), the local state was not updated. The UI relied solely on SWR's `mutate()` to re-fetch the config from the backend, creating a delay where the UI appeared to not update even though the backend had saved the new order.

**Fix**: After a successful backend API call, immediately update the corresponding local state fields (`localColumnOrder`, `localColumnVisibility`, etc.) so the UI reflects the change instantly.

## Changes

- `keep-ui/widgets/alerts-table/ui/alert-table-headers.tsx`: `isRightmostColumn()` now uses `columnOrder` prop instead of `table.getVisibleLeafColumns()`
- `keep-ui/entities/presets/model/usePresetColumnState.ts`: Local state is updated immediately after a successful backend API call

## Testing

1. Open a saved preset with at least 3 columns
2. Click the header menu of the rightmost data column (before `alertMenu`)
3. Verify "Move column right" is disabled (correct behavior)
4. Click "Move column left" - the column should move left immediately
5. Verify the second-to-last column now shows "Move column right" as enabled

### DIFF
--- a/keep-ui/entities/presets/model/usePresetColumnState.ts
+++ b/keep-ui/entities/presets/model/usePresetColumnState.ts
@@ -166,7 +166,26 @@ export const usePresetColumnState = ({
         };
 
         try {
-          return await updateColumnConfig(batchedUpdate);
+          const result = await updateColumnConfig(batchedUpdate);
+          // After successful backend update, also update local state immediately
+          // so the UI reflects the change without waiting for mutate() to re-fetch.
+          // This prevents the "configuration saved but column didn't move" bug.
+          if (updates.columnOrder !== undefined) {
+            setLocalColumnOrder(updates.columnOrder);
+          }
+          if (updates.columnVisibility !== undefined) {
+            setLocalColumnVisibility(updates.columnVisibility);
+          }
+          if (updates.columnRenameMapping !== undefined) {
+            setLocalColumnRenameMapping(updates.columnRenameMapping);
+          }
+          if (updates.columnTimeFormats !== undefined) {
+            setLocalColumnTimeFormats(updates.columnTimeFormats);
+          }
+          if (updates.columnListFormats !== undefined) {
+            setLocalColumnListFormats(updates.columnListFormats);
+          }
+          return result;
         } catch (err) {
           // If backend update fails, fall back to local storage
           console.warn(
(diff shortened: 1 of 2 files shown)


shahargl:
@waterWang looks like never-tested code, 100% ai generated, are you using Keep?

PR #6668, opened 2026-08-05
fix(victorialogs): return empty headers for NoAuth instead of None

### LINKED ISSUE #6669: VictoriaLogs provider: _query crashes with AttributeError when authentication_type is NoAuth
## Describe the bug

`VictorialogsProvider.generate_auth_headers()` returns a dict for `Basic` and `Bearer`, but has no `return` statement on the `NoAuth` path, so it implicitly returns `None`. Every branch of `_query()` then calls:

```python
headers = self.generate_auth_headers()
headers.update({"AccountID": account_id, "ProjectID": project_id})
```

on that value, which raises immediately.

As a result `NoAuth` — both the default and one of the three options offered in the provider UI — makes the provider unusable: every query type (`query`, `hits`, `stats_query`, `stats_query_range`) fails the same way. The provider installs and appears healthy; the failure only surfaces when a query actually runs, e.g. from a workflow step.

## To Reproduce

1. Install the VictoriaLogs provider pointing at a VictoriaLogs instance that requires no authentication, leaving `Authentication Type` as `NoAuth`.
2. Run any query through it:

```python
provider._query(queryType="query", query="*", limit="3")
```

## Expected behavior

The query runs and no `Authorization` header is sent.

## Actual behavior

```
File ".../keep/providers/victorialogs_provider/victorialogs_provider.py", line 220, in _query
    headers.update({"AccountID": account_id, "ProjectID": project_id})
    ^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'update'
```

## Suggested fix

Return an empty dict as the fallback in `generate_auth_headers()`, so the `NoAuth` path yields `{}` instead of `None`.

## Workaround

Select `Basic` and supply arbitrary non-empty credentials — an unauthenticated VictoriaLogs ignores the `Authorization` header.

## Version

Observed on keep 0.52.1 (`keep-api` image); the same code is present on `main`.

artemrootman:
## Problem

`VictorialogsProvider.generate_auth_headers()` returns a dict for `Basic` and `Bearer`, but has no `return` on the `NoAuth` path, so it implicitly returns `None`. Every branch of `_query()` then does:

```python
headers = self.generate_auth_headers()
headers.update({"AccountID": account_id, "ProjectID": project_id})
```

which raises `AttributeError: 'NoneType' object has no attribute 'update'`.

`NoAuth` is both the default and a documented option in the provider UI, so a VictoriaLogs instance that needs no credentials cannot be queried at all — every query type (`query`, `hits`, `stats_query`, `stats_query_range`) fails the same way. The provider installs and looks healthy; the failure only surfaces when a query actually runs, e.g. from a workflow step.

The current workaround is to select `Basic` and supply arbitrary credentials, since an unauthenticated instance ignores the `Authorization` header.

## Fix

Return `{}` as the fallback, so `NoAuth` yields an empty mapping instead of `None`. `Basic` and `Bearer` behaviour is unchanged.

## Tests

Adds `tests/providers/victorialogs_provider/test_victorialogs_noauth.py`:

- `generate_auth_headers()` returns `{}` for explicit `NoAuth` and for the default (unset) authentication type;
- `Basic` and `Bearer` still produce their expected headers;
- `_query()` no longer raises for `query`, `hits` and `stats_query` with `NoAuth`, and sends no `Authorization` header.

`black` and `isort` are clean on both files.

Fixes #6669

### DIFF
--- a/keep/providers/victorialogs_provider/victorialogs_provider.py
+++ b/keep/providers/victorialogs_provider/victorialogs_provider.py
@@ -181,6 +181,10 @@ class VictorialogsProvider(BaseProvider):
                 headers["X-Scope-OrgID"] = self.authentication_config.x_scope_orgid
             return headers
 
+        # NoAuth: no headers to add, but callers do headers.update(...) on the
+        # result, so an empty dict must be returned rather than None.
+        return {}
+
     def _convert_to_json(self, response: str) -> dict:
         """
         Convert the response string to JSON.
--- a/tests/providers/victorialogs_provider/test_victorialogs_noauth.py
+++ b/tests/providers/victorialogs_provider/test_victorialogs_noauth.py
@@ -1,0 +1,84 @@
+"""
+Tests for the VictoriaLogs provider authentication headers.
+
+generate_auth_headers() had no return statement on the NoAuth path, so it
+returned None and every _query() call failed with
+"AttributeError: 'NoneType' object has no attribute 'update'".
+"""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from keep.contextmanager.contextmanager import ContextManager
+from keep.providers.models.provider_config import ProviderConfig
+from keep.providers.victorialogs_provider.victorialogs_provider import (
+    VictorialogsProvider,
+)
+
+HOST_URL = "http://victorialogs.example.com:9428"
+
+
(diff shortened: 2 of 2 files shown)


shahargl (approved):
lgtm

PR #6671, opened 2026-08-05
fix(litellm): fail loudly on null content and accept fenced JSON

### LINKED ISSUE #6670: LiteLLM provider: silent {"response": null} on reasoning models, and structured output fails on fenced JSON
## Describe the bug

Two problems in `LitellmProvider._query`, both of which make a workflow step give a wrong result rather than an error.

**1. `content: null` is returned as a successful result.** The provider does:

```python
generated_text = result["choices"][0]["message"]["content"]
```

A reasoning model that spends its whole `max_tokens` budget on reasoning tokens replies with `content: null` and `finish_reason: "length"`. The provider passes that straight through as `{"response": None}`, the step is marked **successful**, and whatever consumes it sees an empty answer with nothing in the logs to explain why. Reproduced with two different reasoning models at `max_tokens: 400` and again at `3000`.

**2. Structured output rejects valid JSON in a code fence.** With `structured_output_format` set, the reply goes through a bare `json.loads()`. Models very commonly answer with the JSON wrapped in a ```json fence, which raises `Failed to parse generated text as JSON` and fails the step even though the payload itself is exactly the requested schema.

## Expected behavior

1. A `null` content should raise a clear error naming `finish_reason`, so the cause (budget consumed by reasoning) is obvious.
2. A fenced JSON reply should be unwrapped before parsing.

## Additional observation

Separately, `requests.post(..., timeout=60)` does not bound the call in practice: with a large `max_tokens` a request stayed open for several minutes and the workflow execution sat in `in_progress` the whole time. Not addressed in the attached PR, since a proper fix needs a total-time budget rather than a socket timeout.

## Version

Observed on 0.52.1 and 0.54.2; the code is unchanged on `main`.

artemrootman:
## Problem

See #6670. `LitellmProvider._query` returns `{"response": None}` as a success when a reasoning model exhausts `max_tokens` on reasoning tokens, and fails a step outright when a model wraps otherwise-valid structured output in a ```json fence.

## Fix

- Raise `ProviderException` when `content` is `null`, naming `finish_reason` and pointing at the `max_tokens`/reasoning cause instead of silently yielding an empty answer.
- Add `_strip_code_fence()` and apply it before `json.loads()` for `structured_output_format`, so a fenced reply parses. Plain JSON and plain text are untouched.

## Tests

Adds `tests/providers/litellm_provider/test_litellm_response_parsing.py`: fence unwrapping (with and without the `json` tag, and with surrounding whitespace), plain JSON left as is, structured output accepting a fenced reply, `null` content raising with `finish_reason` in the message, and a plain-text reply still working. 7 tests, all passing.

`black` is clean on both files. `isort` reports the provider's imports as unsorted, but that is pre-existing on `main` and untouched here.

Fixes #6670

### DIFF
--- a/keep/providers/litellm_provider/litellm_provider.py
+++ b/keep/providers/litellm_provider/litellm_provider.py
@@ -60,6 +60,19 @@ class LitellmProvider(BaseProvider):
         """Format the prompt as a chat message."""
         return [{"role": "user", "content": prompt}]
 
+    @staticmethod
+    def _strip_code_fence(text: str) -> str:
+        """Unwrap a ```json ... ``` fence, which models add unprompted."""
+        stripped = text.strip()
+        if not stripped.startswith("```"):
+            return text
+        stripped = stripped[3:]
+        if stripped.lower().startswith("json"):
+            stripped = stripped[4:]
+        if stripped.endswith("```"):
+            stripped = stripped[:-3]
+        return stripped.strip()
+
     def _query(
         self,
         prompt: str,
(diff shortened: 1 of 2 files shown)


artemrootman:
The `unit-tests` failure looks unrelated to this change. The run reports `1067 passed, 84 skipped, 1 error`, and the single error is at **teardown** of `tests/test_workflowstore.py::test_unprovision_workflows[test_app0]` with `sqlite3.OperationalError: database table is locked`, preceded by `no such table: workflowexecutionlog` noise from the workflow-log flusher. This PR only touches response parsing in `LitellmProvider._query`; it does not import or exercise the workflow store, and the log never references the added test file — its 7 tests are among the 1067 that passed.

For reference, `main` at a465d44 is green on the same job, and it already contains `tests/providers/victorialogs_provider/test_victorialogs_noauth.py`, which builds a provider through `ContextManager` in exactly the same way as the tests added here.

Could you re-run the job? I can't trigger it from a fork.

shahargl (approved):
lgtm

PR #6673, opened 2026-08-05
feat(grafana): query any datasource via /api/ds/query

### LINKED ISSUE #6672: Grafana provider cannot query datasources
## Problem

`GrafanaProvider` can manage alerts, install webhooks and pull topology, but it has no `_query` at all, so a workflow cannot read data from a datasource that Grafana already knows about.

That is a large gap in practice. A Grafana instance typically has every backing store already configured — Prometheus/VictoriaMetrics, ClickHouse, SQL databases, log stores — with credentials, permissions and audit already handled centrally. Today each of those needs its own Keep provider with its own credentials, and for datasources that have no Keep provider there is no path at all.

## Proposal

Add `_query()` on top of Grafana's `/api/ds/query`, taking a `datasource_uid` plus either a convenience argument (`expr` for Prometheus-style, `raw_sql` for SQL) or a full `query` object for datasource-specific fields, and a Grafana time range.

Since `/api/ds/query` answers with columnar data frames (`schema.fields` naming the columns, `data.values` holding one array per column), the provider should flatten them into plain row dicts so workflow templates can use the result directly.

Verified shapes against a real Grafana:

```jsonc
// prometheus-type datasource
{"schema": {"fields": [{"name": "Time"}, {"name": "Value"}]},
 "data": {"values": [[1785959838794], [1611]]}}

// clickhouse datasource
{"schema": {"fields": [{"name": "provider_id"}, {"name": "c"}]},
 "data": {"values": [["a", "b"], [3884296, 1849312]]}}
```

Flattened: `[{"provider_id": "a", "c": 3884296}, {"provider_id": "b", "c": 1849312}]`.

artemrootman:
## Problem

See #6672. `GrafanaProvider` has no `_query`, so workflows cannot read from datasources that Grafana already has configured — even though those datasources usually cover every backing store an operator cares about, with credentials and permissions already centralised there.

## Fix

Adds `_query()` on top of `/api/ds/query`:

- `datasource_uid` (required), plus `expr` for Prometheus-style datasources, `raw_sql` for SQL ones, or a full `query` dict for anything datasource-specific — the dict is merged last so it can override the convenience arguments;
- `start`/`end` accept Grafana's time syntax, absolute or relative (`now-1h`), with `instant` and `max_data_points` for the Prometheus case;
- errors surface as `ProviderException`, both for a non-2xx response and for a per-refId `error` in the body, instead of returning something empty.

Adds `_frames_to_rows()`, which flattens Grafana's columnar data frames (`schema.fields` names the columns, `data.values` holds one array per column) into a list of row dicts, so a workflow template can consume the result directly rather than re-implementing frame parsing.

## Tests

`tests/providers/grafana_provider/test_grafana_datasource_query.py` — 10 tests: frame flattening for SQL and Prometheus shapes, multiple frames concatenated, empty/malformed frames returning `[]` rather than raising, payload construction (uid, time range, `rawSql`), an explicit `query` dict overriding defaults, missing `datasource_uid`, an HTTP error and a datasource-level error. All passing; `black` clean on both files.

Beyond the unit tests, the implementation was exercised against a live Grafana over three datasource types — a Prometheus-compatible one and two ClickHouse instances — and returned correctly flattened rows in each case.

Fixes #6672

### DIFF
--- a/docs/snippets/providers/grafana-snippet-autogenerated.mdx
+++ b/docs/snippets/providers/grafana-snippet-autogenerated.mdx
@@ -1,4 +1,4 @@
-{/* This snippet is automatically generated using scripts/docs_render_provider_snippets.py 
+{/* This snippet is automatically generated using scripts/docs_render_provider_snippets.py
 Do not edit it manually, as it will be overwritten */}
 
 ## Authentication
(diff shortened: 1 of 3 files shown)


shahargl (approved):
lgtm

PR #6678, opened 2026-08-06
feat(litellm): return token usage and cost with the response

### LINKED ISSUE #6677: LiteLLM provider discards token usage, so a workflow cannot report what a run cost
## Problem

`LitellmProvider._query` parses the completion and returns only the text:

```python
return {
    "response": generated_text,
}
```

The `usage` block from the API response — `prompt_tokens`, `completion_tokens`, `total_tokens` and, on providers that report it, `cost` — is dropped on the floor.

That information is exactly what an operator wants next to a generated summary: how many tokens this particular analysis consumed and what it cost. With the provider as it stands there is no way to get it from a workflow, and the only workaround is to bypass the provider entirely and call the API from a `python` step, which then has to handle the endpoint, the key and the error paths itself.

## Proposal

Return the usage fields alongside the response, defaulting to `None` when the backend does not report them (not every OpenAI-compatible server does), plus the model actually used, which can differ from the one requested when a proxy routes the call.

artemrootman:
## Problem

See #6677. `LitellmProvider._query` returns only `{"response": ...}` and discards the `usage` block, so a workflow cannot tell how many tokens an analysis consumed or what it cost. The only way to get that today is to skip the provider and call the API from a `python` step.

## Fix

Return the usage fields next to the response:

- `prompt_tokens`, `completion_tokens`, `total_tokens`, `cost` — read through `.get()` so a backend that reports no usage yields `None` rather than raising;
- `model` as reported by the API, which can differ from the requested one when a proxy routes the call.

`response` keeps its meaning, so existing workflows are unaffected.

## Tests

Two added to `tests/providers/litellm_provider/test_litellm_response_parsing.py`: usage and model surfaced from a normal response, and a response without a `usage` block leaving the fields `None` while the text still comes through. 9 tests in the file, all passing; `black` clean.

Fixes #6677

### DIFF
--- a/keep/providers/litellm_provider/litellm_provider.py
+++ b/keep/providers/litellm_provider/litellm_provider.py
@@ -139,8 +139,16 @@ class LitellmProvider(BaseProvider):
                         f"Failed to parse generated text as JSON: {generated_text}. Model not following the structured output format. Response: {result}"
                     )
 
+            # Surface usage so a workflow can report what a run cost. Absent
+            # on some OpenAI-compatible backends, hence the .get() chain.
+            usage = result.get("usage") or {}
             return {
                 "response": generated_text,
+                "model": result.get("model", model),
+                "prompt_tokens": usage.get("prompt_tokens"),
+                "completion_tokens": usage.get("completion_tokens"),
+                "total_tokens": usage.get("total_tokens"),
+                "cost": usage.get("cost"),
             }
 
         except requests.exceptions.RequestException as e:
--- a/tests/providers/litellm_provider/test_litellm_response_parsing.py
+++ b/tests/providers/litellm_provider/test_litellm_response_parsing.py
@@ -34,16 +34,17 @@ def _build_provider() -> LitellmProvider:
     return LitellmProvider(ContextManager(tenant_id="test"), "litellm-test", config)
 
 
-def _response(content, finish_reason="stop"):
+def _response(content, finish_reason="stop", usage=None, model=None):
     response = MagicMock()
     response.raise_for_status = MagicMock()
-    response.json = MagicMock(
-        return_value={
-            "choices": [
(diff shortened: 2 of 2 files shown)


artemrootman:
The `unit-tests` failure is unrelated to this change: the run reports `1078 passed, 84 skipped, 1 error`, and the error is at **setup** of `tests/test_auth.py::test_deleted_api_key_authentication[MULTI_TENANT]`, with `HTTPSConnection(host='https', port=443)` — a real network call. This PR only touches the return value of `LitellmProvider._query`; its two new tests are among the 1078 that passed.

The cause looks reproducible rather than random, if it helps:

1. `tests/fixtures/client.py` sets `AUTH0_DOMAIN` to `https://auth0domain.com` — with the scheme included — for the `MULTI_TENANT` string parameter.
2. `_mock_oidc_discovery()` is only applied when `is_auth0` is true, which is set from `request.param.get("AUTH_TYPE") == "AUTH0"` and therefore only for the dict-style parameters. The `MULTI_TENANT` string path gets `nullcontext()`.
3. `ee/identitymanager/identity_managers/auth0/auth0_authverifier.py` then runs at import time and builds the JWKS URL as `f"https://{auth_domain}/.well-known/jwks.json"`, which with the scheme already present becomes `https://https://auth0domain.com/...`. urllib3 parses the host out of that as `https`, hence the connection attempt in the traceback.

So whenever that module is imported unmocked in this parameterisation, the setup depends on an outbound HTTPS request, which is why it surfaces intermittently across workers. Applying the discovery mock for `MULTI_TENANT` as well, or storing `AUTH0_DOMAIN` without the scheme, would remove the network dependency.

Happy to send that as a separate PR if it is useful. Could you re-run the job in the meantime? I cannot trigger it from a fork.

shahargl (approved):
lgtm

PR #6679, opened 2026-08-07
fix(maintenance_windows): handle invalid CEL syntax gracefully

### LINKED ISSUE #6683: [🐛 Bug]: Unhandled CELParseError in maintenance windows crashes event processing and drops alerts
**Describe the bug**
If a user configures a Maintenance Window with invalid CEL syntax (such as using a single = for assignment instead of == for comparison), it crashes the entire event processing pipeline. The celpy library throws a CELParseError which is currently unhandled in keep/api/bl/maintenance_windows_bl.py. This exception bubbles all the way up to process_event_task.py, causing the background worker to fail and completely drop the incoming alert. A user error in a maintenance rule should not halt global event processing.

**To Reproduce**
Steps to reproduce the behavior (Note: The Web UI prevents this, but it is achievable via the API):

1. Make a request to the KeepHQ API to create a new Maintenance Window.
2. In the API payload, define a condition using invalid CEL syntax (e.g., fingerprint = "5ce80b0bc9f9e84fb53d..." instead of ==).
3. Send a test alert/event to KeepHQ.
4.  Check the background worker logs and observe the exception. The alert is completely dropped and never processed.

**Expected behavior**
KeepHQ should gracefully handle syntax and evaluation errors in user-defined rules, even if they bypass UI validation via the API. The check_if_alert_in_maintenance_windows or evaluate_cel function should catch the CELParseError, log a clear error message, and return False (assuming the alert is not in maintenance mode) so the pipeline can continue processing the event.

**Screenshots**
Not applicable for the UI, but here is the resulting crash from the worker logs:

**Plaintext**
```
celpy.celparser.CELParseError: fingerprint = "5ce80b0bc9f9e84fb53d2780fe7b6409db191
            ^
Expected one of: 
	 DOT
	 "||"
     ...
	* "=="
```
**Additional context**
The crash occurs at this specific execution path:

- keep/api/tasks/process_event_task.py inside __handle_formatted_events
- keep/api/bl/maintenance_windows_bl.py inside check_if_alert_in_maintenance_windows
- keep/api/bl/maintenance_windows_bl.py inside evaluate_cel calling environment.compile(cel)

Adding a standard try...except block around the compilation and evaluation logic in evaluate_cel that catches Exception (or specifically celpy.celparser.CELParseError) and returns False will resolve this vulnerability.

aaleksandrov-top:
## 📑 Description
Wrap CEL compilation and evaluation in a single try-except block to catch all exceptions, including invalid CEL syntax errors. Previously, invalid CEL expressions would crash the event processing pipeline.

Example error that is now handled:
  maintenance_window.cel_query = "alert.severity = 'high'"  # Invalid: = instead of ==
  → CELParseError: syntax error in expression

Now invalid rules are logged and safely skipped, returning False so alerts continue processing instead of being dropped.

## ✅ Checks
- ✅ My pull request adheres to the code style of this project
- ✅ All the tests have passed

Closed #6683

### DIFF
--- a/keep/api/bl/maintenance_windows_bl.py
+++ b/keep/api/bl/maintenance_windows_bl.py
@@ -120,20 +120,19 @@ class MaintenanceWindowsBl:
     @staticmethod
     def evaluate_cel(maintenance_window: MaintenanceWindowRule, alert: AlertDto | Alert, environment: celpy.Environment, logger, logger_extra_info: dict) -> bool:
 
-        cel = preprocess_cel_expression(maintenance_window.cel_query)
-        ast = environment.compile(cel)
-        prgm = environment.program(ast)
-
-        if isinstance(alert, AlertDto):
-            payload = alert.dict()
-        else:
-            payload = alert.event
-        # todo: fix this in the future
-        payload["source"] = payload["source"][0]
-
-        activation = celpy.json_to_cel(json.loads(json.dumps(payload, default=str)))
-
         try:
+            cel = preprocess_cel_expression(maintenance_window.cel_query)
+            ast = environment.compile(cel)
+            prgm = environment.program(ast)
+
+            if isinstance(alert, AlertDto):
+                payload = alert.dict()
+            else:
+                payload = alert.event
+            # todo: fix this in the future
+            payload["source"] = payload["source"][0]
+
+            activation = celpy.json_to_cel(json.loads(json.dumps(payload, default=str)))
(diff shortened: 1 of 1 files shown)


aaleksandrov-top:
Hi, @shahargl    could I kindly ask you to look at this PR?

It is not a big changes but critical issue if user push wrong CEL for maintenance via API.

Thanks

PR #6680, opened 2026-08-08
fix: correctly parse PUSHER_USE_SSL environment variable (#6582)

### LINKED ISSUE #6582: [🐛 Bug]: PUSHER_USE_SSL is incorrectly parsed

ChickenisLegit:
Closes #6582.

This PR fixes a bug where setting `PUSHER_USE_SSL=false` in Docker Compose or other environments would evaluate to `True`. 
This occurred because `os.environ.get` returns a string (e.g. `"false"`), which evaluated to `True` since any non-empty string in Python evaluates to `True` when checked against `False` (e.g. `ssl=False if os.environ.get("PUSHER_USE_SSL", False) is False else True`).

### Changes
- Updated `PUSHER_USE_SSL` parsing to properly handle string booleans (`"false"`, `"0"`, `"true"`, `"1"`, `"yes"`, `"on"`).
- Defaults to `"false"` if the environment variable is not present, matching the existing fallback logic.

### DIFF
--- a/keep/api/core/dependencies.py
+++ b/keep/api/core/dependencies.py
@@ -67,6 +67,10 @@ def get_pusher_client() -> Pusher | None:
 
     # TODO: defaults on open source no docker
     try:
+        pusher_use_ssl = os.environ.get("PUSHER_USE_SSL", "false")
+        if isinstance(pusher_use_ssl, str):
+            pusher_use_ssl = pusher_use_ssl.lower() in ("1", "true", "yes", "on")
+        
         pusher = Pusher(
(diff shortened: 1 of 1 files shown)


talboren (approved):

ChickenisLegit:

Issue #6475, opened 2026-05-13
[➕ Feature]: Add method to query prometheus provider


thibautmery:
**Is your feature request related to a problem? Please describe.**
I ask IA to get information on alert and query prometheus. It says that there is no methods on provider prometheus.
```
Actuellement, je ne peux pas exécuter de nouvelles requêtes directement sur Prometheus (comme lancer un PromQL).
Voici pourquoi :
Méthodes indisponibles : (methods: []).
```
I want to use IA to query prometheus and analyse by his own the results. 

**Describe the solution you'd like**
Add method to query prometheus using IA.

PR #6687, opened 2026-08-12
fix(http-provider): stop duplicating results on notify

### LINKED ISSUE #6431: [🐛 Bug]: HTTP provider action results duplicated in workflow execution output
**Describe the bug**                                                                                                                                                              
  When using the `http` provider as a workflow action, the execution results contain duplicate entries — the same result appears twice in the output list.

  **To Reproduce**
  1. Create a workflow with an `http` provider action:
  ```yaml
  id: test-http-action
  name: Test HTTP Action
  triggers:
    - type: manual
  actions:
    - name: post-request
      provider:
        type: http
        with:
          url: http://example.com/api
          method: POST
          body:
            message: "Hello"
  2. Run the workflow manually
  3. Open the workflow execution and check the results
```

**Expected behavior**
 Results list contains 1 entry per action execution.

**Actual behavior**
  Results list contains 2 identical entries:
```json
[
  {"status": true, "status_code": 200, "body": "..."},
  {"status": true, "status_code": 200, "body": "..."}
]
```

**Screenshots**
N/A

**Additional context**
Root cause: HttpProvider._notify() calls self.query() (public method) instead of self._query() (private method).
BaseProvider.notify() appends to self.results after _notify() returns. Since _notify() internally calls self.query(), which also appends to self.results, the result is appended twice per execution.
keep/providers/http_provider/http_provider.py, line 66
Fix: replace self.query(...) with self._query(...) in HttpProvider._notify().

dngr2:
Fixes #6431

## The bug

`HttpProvider._notify()` calls the public `self.query()` instead of the private `self._query()`.

`BaseProvider.notify()` appends the return value to `self.results`, and `BaseProvider.query()` appends as well — so every HTTP action records its result twice in the workflow execution output:

```json
[
  {"status": true, "status_code": 200, "body": "..."},
  {"status": true, "status_code": 200, "body": "..."}
]
```

Thanks to @DANIILSKRIPCHENKO for the diagnosis in the issue — it was accurate, this PR is that fix plus tests.

## Why calling `_query()` directly is equivalent

`query()` does two things `_query()` does not, and neither applies here:

- it pops `enrich_alert` / `audit_enabled` from kwargs — but `notify()` has already popped those before it calls `_notify()`, so they are never present at this point
- it adds `results[0].__class__` to `context_manager.dependencies` when the result is a list — `_query()` returns a dict, so that branch is a no-op for this provider

## Tests

New file: `tests/providers/http_provider/test_http_provider_bugs.py`

The tests were written before the fix and run against unmodified `main`: **6 failed, 1 passed**, failing with `Left contains one more item`. With the one-line change: **7 passed**.

- exactly one result appended, on POST / PUT / DELETE
- two `notify()` calls append two results, not four
- `notify()` still returns the same payload
- guard: `query()` called directly still records its own result, so the fix does not over-correct

`tests/providers` in full: 60 passed. I have not run the suites that need Docker and a database locally — leaving those to CI.

### DIFF
--- a/keep/providers/http_provider/http_provider.py
+++ b/keep/providers/http_provider/http_provider.py
@@ -63,7 +63,9 @@ class HttpProvider(BaseProvider):
         """
         Send a HTTP request to the given url.
         """
-        return self.query(
+        # _query, not query: BaseProvider.notify() already appends the result,
+        # and BaseProvider.query() would append it a second time (#6431).
+        return self._query(
             url=url,
             method=method,
             headers=headers,
--- a/tests/providers/http_provider/test_http_provider_bugs.py
+++ b/tests/providers/http_provider/test_http_provider_bugs.py
@@ -1,0 +1,81 @@
+"""Tests for HTTP provider result duplication (issue #6431).
+
+Bug: HttpProvider._notify() called the public self.query() instead of the
+private self._query(). BaseProvider.notify() appends to self.results after
+_notify() returns, and BaseProvider.query() appends as well, so every notify()
+recorded the same result twice in the workflow execution output.
+"""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from keep.contextmanager.contextmanager import ContextManager
+from keep.providers.http_provider.http_provider import HttpProvider
+from keep.providers.models.provider_config import ProviderConfig
+
+
+def _build_provider() -> HttpProvider:
+    config = ProviderConfig(description="HTTP Provider", authentication={})
+    return HttpProvider(ContextManager(tenant_id="test"), "http-test", config)
+
+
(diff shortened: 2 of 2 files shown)


dngr2:
No rush on this — flagging only that it and its two siblings are ready whenever a reviewer has time, and that all three are independent of each other:

- #6687 → #6431, `_notify()` calling the public `query()` so every HTTP action's results are appended twice
- #6688 → #6547, a templated JSON body double-encoded because `_query()` normalises string `headers` but not string `body`
- #6689 → #6475, `PROVIDER_METHODS` empty so the instant PromQL query already implemented cannot be reached

Each has a test that fails against the unpatched code, CI is green and the CLA is signed. Happy to split, rescope or rebase any of them if that makes review easier.

Issue #6274, opened 2026-04-21
[🐛 Bug]: Workflow UI editor silently moves `with:` out of `provider:`, breaking every action


ckaiser-graniterock:
## Summary

The Keep Workflows UI editor silently reformats pasted YAML so that `with:` ends up as a **sibling** of `provider:` rather than **nested inside** it. The parser reads `step.provider.with` — so this breaks every action with errors like `HttpProvider._notify() missing 2 required positional arguments: 'url' and 'method'` or `Message is required` (Slack). No warning, no validation error — the workflow saves and fires, just with zero parameters.

## Environment

- Keep build: commit `339151c9` on `main` (released 2026-04-14)
- Affected UI: **Workflows → + Create Workflow → Import YAML** form
- Affected workflows: any workflow with actions/steps that use `with:`

## Reproduction

Paste this correctly-indented YAML into the Workflow editor:

```yaml
workflow:
  id: test-with-nesting
  name: Test
  triggers:
    - type: alert
      filters:
        - key: source
          value: prometheus
  actions:
    - name: test-slack
      provider:
        type: slack
        config: " {{ providers.slack-prod }} "
        with:
          message: "test"
```

Save. Query the stored `workflow_raw` directly:

```bash
docker exec keep-keep-backend-1 python3 -c "
import os, psycopg2
dsn = os.environ['DATABASE_CONNECTION_STRING'].replace('postgresql+psycopg2://', 'postgresql://')
conn = psycopg2.connect(dsn); cur = conn.cursor()
cur.execute(\"SELECT workflow_raw FROM workflow WHERE name = 'Test'\")
print(cur.fetchone()[0])"
```

Observed output (note indentation change):

```yaml
id: test-with-nesting
name: Test
triggers:
- type: alert
  filters:
  - key: source
    value: prometheus
actions:
- name: test-slack
  provider:
    type: slack
    config: " {{ providers.slack-prod }} "
  with:                 # ← MOVED OUT of provider!
    message: "test"
```

## Fire the workflow

Trigger an alert matching the filter. Logs on `keep-keep-backend-1`:

```
"Notifying message to Slack using webhook", "slack_message": "", "blocks": [], "channel": ""
"Failed to run step test-slack with error Message is required - see for example https://github.com/keephq/keep/blob/main/examples/workflows/slack_basic.yml#L16"
```

[...]

PR #6698, opened 2026-08-16
fix(presets): don't rewrite search text when translating CEL contains

### LINKED ISSUE #6699: [🐛 Bug]: A preset filter searching for the word "contains" silently searches for "includes"
## Describe the bug

`sanitizeCELIntoJS` translates the CEL `contains` operator into JavaScript `includes` before the filter expression is evaluated. It does so with an unanchored global replace:

`keep-ui/features/presets/presets-manager/lib/eval-with-context.ts:13`

```ts
let jsExpression = celExpression.replace(/contains/g, "includes");
```

That rewrites those letters wherever they appear — including inside quoted search strings and inside identifiers.

## To reproduce

Enter any of these as a preset / alerts filter:

| Filter typed | What is actually evaluated |
|---|---|
| `description.contains("contains")` | `description.includes("includes")` |
| `name.contains("this contains errors")` | `name.includes("this includes errors")` |
| `contains_pii == "true"` | `includes_pii == "true"` |
| `labels.contains_secret == "yes"` | `labels.includes_secret == "yes"` |

## Expected behavior

The operator is translated; the search text and field names are left alone.

## Actual behavior

Two distinct failures:

1. **The search term is rewritten.** A filter looking for alerts whose description contains the word "contains" instead matches alerts containing "includes".
2. **A field is renamed out of existence.** Any field or label whose name holds those letters — `contains_pii`, `contains_secret` — is rewritten to a name that doesn't exist, so the comparison evaluates against `undefined` and quietly matches nothing.

No error is raised in either case. The preset saves and filters normally, just against something the user never typed.

## Additional context

The file has no tests.

This is the same *class* of silent rewriting as #6274, but a different code path — that one is the workflow YAML editor and is covered by #6647. This is the preset/alerts CEL evaluator.

A fix and tests are proposed in #6698: matching the call form `.contains(` rather than the bare word. That leaves one narrower case — a search string containing the literal `.contains(` — which needs a tokenizer rather than a regex to separate code from string literals.

dngr2:
Fixes #6699.

## What was wrong

`sanitizeCELIntoJS` turns the CEL `contains` operator into JS `includes` before the expression is evaluated. It did so with an unanchored `/contains/g`, which also rewrote the inside of quoted search strings and any identifier that merely held those letters.

| The user typed | What was evaluated |
|---|---|
| `description.contains("contains")` | `description.includes("includes")` |
| `name.contains("this contains errors")` | `name.includes("this includes errors")` |
| `contains_pii == "true"` | `includes_pii == "true"` |
| `labels.contains_secret == "yes"` | `labels.includes_secret == "yes"` |
| `description.contains("db")` | `description.includes("db")` ✔ |

Two separate failures. The search term gets rewritten, so the filter looks for a word the user never typed. And a field or label whose name holds those letters is renamed to one that does not exist, so the comparison evaluates against `undefined` and quietly matches nothing.

Nothing throws at any point. The preset saves and filters, just against the wrong thing.

## The fix

Match the call form `.contains(` instead of the bare word. That fixes every row above.

**What it does not fix:** a search string containing the literal `.contains(` is still rewritten. Doing that correctly means separating code from string literals with a tokenizer rather than a regex, which is a larger change to a function whose own neighbouring comment already says `// this pattern is far from robust`. I would rather send the narrow correct change than a rewrite bundled with a bug fix, but happy to take it further if you would prefer that.

## Tests

The file had none. Added `__tests__/eval-with-context.test.ts`, testing through the exported `evalWithContext` so it exercises the real path rather than the private helper:

- the plain substring case that always worked, and a genuine non-match
- the literal word `contains` as the search term, plus the inverse — that it does *not* match an alert whose text says "includes"
- the same word inside a longer phrase
- several `contains` calls in one expression
- a field named `contains_pii`, matching and mismatching

**Against the unfixed function 4 of the 8 fail** and the 4 covering ordinary behaviour still pass, so the tests pin the bug rather than the implementation. `npx jest features/presets` is 18/18 green.

Related to #6274, which reports the same class of silent rewriting in the workflow YAML editor. This is a different code path — that one is covered by #6647 — but the underlying pattern is the same, so it seemed worth fixing here too.

### DIFF
--- a/keep-ui/features/presets/presets-manager/lib/__tests__/eval-with-context.test.ts
+++ b/keep-ui/features/presets/presets-manager/lib/__tests__/eval-with-context.test.ts
@@ -1,0 +1,75 @@
+import { AlertDto } from "@/entities/alerts/model";
+import { evalWithContext } from "../eval-with-context";
+
+// sanitizeCELIntoJS rewrites the CEL `contains` operator to JS `includes`
+// before the expression is evaluated. It used to do so with an unanchored
+// /contains/g, which also rewrote the inside of quoted search strings and any
+// identifier that merely contained those letters. Nothing threw: the preset
+// saved and filtered, just against something the user never typed.
+const alert = {
+  id: "1",
+  name: "test",
+  description: "this contains errors",
+  severity: "critical",
+  status: "firing",
+  source: ["prometheus"],
+} as unknown as AlertDto;
+
+describe("evalWithContext - contains", () => {
+  it("matches on a plain substring", () => {
+    expect(evalWithContext(alert, 'description.contains("errors")')).toBe(true);
+  });
+
+  it("does not match a substring that is absent", () => {
+    expect(evalWithContext(alert, 'description.contains("timeout")')).toBe(
+      false
+    );
+  });
+
+  // The search term is the operator's own name. Rewriting the operator must
+  // not rewrite what the user is searching for.
+  it("searches for the literal word 'contains'", () => {
+    expect(evalWithContext(alert, 'description.contains("contains")')).toBe(
+      true
+    );
+  });
+
+  it("does not match 'includes' when the user searched for 'contains'", () => {
+    const other = { ...alert, description: "this includes errors" } as AlertDto;
+    expect(evalWithContext(other, 'description.contains("contains")')).toBe(
+      false
+    );
+  });
+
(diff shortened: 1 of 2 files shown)


shahargl (approved):
lgtm

PR #6720, opened 2026-08-21
fix(keep-provider): don't mutate the shared fingerprint_fields list

### LINKED ISSUE #6719: [🐛 Bug]: KeepProvider mutates the shared fingerprint_fields list across alerts in a foreach batch, corrupting fingerprints
**Describe the bug**

`KeepProvider._build_alert` mutates the caller's `fingerprint_fields` list in place instead of treating it as read-only:

```python
# keep/providers/keep_provider/keep_provider.py, line 177
fingerprint_fields.append("workflowId")
```

`_notify_alert` loops over every alert in a batch (`foreach`, or multiple items matching an `if` condition) and passes that *same* list object into `_build_alert` on every iteration instead of a fresh copy:

```python
# keep/providers/keep_provider/keep_provider.py, lines 549-570
for alert_result in trigger_alerts:
    ...
    alert_dto = self._build_alert(
        alert_result, fingerprint_fields or [], **rendered_alert_data
    )
```

So if a workflow's `keep - create-alert` action sets an explicit `fingerprint_fields` and produces more than one alert per invocation, every alert after the first in that batch gets an extra duplicate `"workflowId"` appended to its fingerprint fields, which changes the resulting SHA-256 fingerprint.

**To Reproduce**

With a real workflow:
1. A step whose result is a list of 2+ items (or a `foreach` over 2+ items).
2. A `keep - create-alert` action on that step with an explicit `fingerprint_fields`, e.g.:
```yaml
actions:
  - name: create-alert
    foreach: "{{ steps.query.results }}"
    provider:
      type: keep
      with:
        fingerprint_fields: ["labels.service"]
        name: "disk full"
        labels:
          service: "{{ foreach.value.service }}"
```
3. Run it with 2+ results in one execution and compare the computed fingerprints — the 2nd, 3rd, etc. alerts get a different (wrong) fingerprint than the 1st, even though they should all be computed the same way (just with different field *values*).

To make this concrete without needing the full app/DB stack, here's a small script that copies the exact control flow above verbatim (including the real hashing logic from `BaseProvider.get_alert_fingerprint`) and runs standalone:

<details>
<summary>repro_fingerprint_mutation.py</summary>

```python
import hashlib


def get_alert_fingerprint(event_dict, fingerprint_fields):
    """Verbatim logic of BaseProvider.get_alert_fingerprint."""
    if not fingerprint_fields:
        return event_dict["name"]
    fingerprint = hashlib.sha256()
    for fingerprint_field in fingerprint_fields:
        keys = fingerprint_field.split(".")
        value = event_dict
        for key in keys:
            if isinstance(value, dict):
                value = value.get(key, None)
            else:
                value = None
                break
        if value is not None:
            fingerprint.update(str(value).encode())
    return fingerprint.hexdigest()


def build_alert(alert_data, fingerprint_fields, workflow_id, name, labels):
    """Verbatim logic of KeepProvider._build_alert (lines 139-179)."""
    event_dict = {"name": name, "labels": labels, "workflowId": workflow_id}
    if not fingerprint_fields:
        fingerprint_fields = ["labels." + label for label in list(labels.keys())]
    fingerprint_fields.append("workflowId")  # <-- mutates the caller's list in place
    event_dict["fingerprint"] = get_alert_fingerprint(event_dict, fingerprint_fields)
    return event_dict


def notify_alert(trigger_alerts, fingerprint_fields, workflow_id):
    """Verbatim logic of KeepProvider._notify_alert's per-alert loop (lines 549-570)."""
    alert_dtos = []
    for alert_result in trigger_alerts:
        alert_dto = build_alert(
            alert_result,
            fingerprint_fields or [],  # <-- same list object passed every iteration
            workflow_id=workflow_id,
            name=alert_result["name"],
            labels=alert_result.get("labels", {}),
        )
        alert_dtos.append(alert_dto)
    return alert_dtos


if __name__ == "__main__":
    workflow_configured_fingerprint_fields = ["labels.service"]

    alerts_from_one_query = [
        {"name": "disk full", "labels": {"service": "db-primary"}},
        {"name": "disk full", "labels": {"service": "db-replica"}},
        {"name": "disk full", "labels": {"service": "cache"}},
    ]

    results = notify_alert(
        alerts_from_one_query, workflow_configured_fingerprint_fields, workflow_id="wf-123"
    )

    for i, r in enumerate(results, start=1):
        print(f"alert {i} ({r['labels']['service']}): fingerprint = {r['fingerprint']}")

    print()
    print("fingerprint_fields after the run:", workflow_configured_fingerprint_fields)

    correct = get_alert_fingerprint(
        {"name": "disk full", "labels": {"service": "cache"}, "workflowId": "wf-123"},
        ["labels.service", "workflowId"],
    )
    print()
    print("alert 3 fingerprint if fingerprint_fields had NOT been mutated:", correct)
    print("alert 3 fingerprint actually produced:                         ", results[2]["fingerprint"])
```

</details>

Output:
```
alert 1 (db-primary): fingerprint = 0663a28c9b4b51c01b0b670aa8527b92d0f6c07072dd1d713dd13ee7a466600a
alert 2 (db-replica): fingerprint = 5b9f6e023a3676a55c7710ba7486d1b4e7e3e9c5292f1af7c139d5c563c4f1f8
alert 3 (cache): fingerprint = 953d8db96249397626403dacfaa7da39847eca6fda9598f17510dc0c1e017275

fingerprint_fields after the run: ['labels.service', 'workflowId', 'workflowId', 'workflowId']

alert 3 fingerprint if fingerprint_fields had NOT been mutated: 8e0bf2924725d79fc3532aef597e485ee73d14d7f7711b5514bef2e2f32c83c2
alert 3 fingerprint actually produced:                          953d8db96249397626403dacfaa7da39847eca6fda9598f17510dc0c1e017275
```

Only 3 alerts were processed, and the shared list already grew from 1 entry to 4 (3 duplicate `"workflowId"`s), and alert 3's fingerprint doesn't match what it should be.

**Expected behavior**

Each alert in a batch should get its own independent copy of `fingerprint_fields`, so identically-configured alerts that only differ in field *values* get fingerprints computed the same, deterministic way — not one that depends on how many alerts were processed earlier in the same batch.

**Additional context**

- Verified against `upstream/main` at `4003992272eeddd649e23e21d63f09892a149291`.
- If the same `fingerprint_fields` list object also happens to be reused across separate workflow executions (rather than freshly built per run), this list would keep growing indefinitely across runs too — I didn't chase that part down since the single-invocation case above is already enough to demonstrate the bug.
- Small fix: stop mutating the caller's list, e.g. `fingerprint_fields = fingerprint_fields + ["workflowId"]` instead of `fingerprint_fields.append("workflowId")` in `_build_alert` (`keep_provider.py:177`). Happy to send a PR for this if useful.

Prabal864:
## What
Fixes #6719.

`KeepProvider._build_alert` did `fingerprint_fields.append("workflowId")` on whatever list was passed in. `_notify_alert`'s per-alert loop passes that same list object into `_build_alert` on every iteration of a `foreach` batch, so each alert after the first accumulated one more duplicate `"workflowId"` entry, which changes its computed SHA-256 fingerprint.

## Fix
- `keep_provider.py:177` - build a new list (`fingerprint_fields + ["workflowId"]`) instead of mutating the caller's list with `.append`.
- `keep_provider.py:139` - switched the mutable default argument (`fingerprint_fields=[]`) to `None` while touching this line, since `if not fingerprint_fields:` already treats `None`/`[]` identically, so this is a no-op behaviorally but removes the classic mutable-default footgun flagged alongside the main bug.

## Testing
Added `tests/test_keep_provider_fingerprint_fields_mutation.py`:
- `test_build_alert_does_not_mutate_caller_fingerprint_fields` - calls `_build_alert` 3x with the same `fingerprint_fields` list (mirroring `_notify_alert`'s loop) and asserts the list is unchanged afterward.
- `test_build_alert_produces_consistent_fingerprints_across_a_batch` - asserts an alert's fingerprint is identical whether it's built standalone or as the 3rd alert in a batch that reused the same `fingerprint_fields` object.

I wasn't able to run the full suite in my sandbox - `pip install -e .` fails on this platform because `uvloop` doesn't support Windows (a pyproject dependency), and `tests/conftest.py` needs Docker/MySQL fixtures at collection time regardless of which test is targeted. I verified both the bug and the fix by extracting the exact control flow (`_build_alert`'s fingerprint-fields handling and the real `BaseProvider.get_alert_fingerprint` hashing algorithm) into a standalone script and running it directly - before the fix, a 3-alert batch left `fingerprint_fields` as `['labels.service', 'workflowId', 'workflowId', 'workflowId']` and alert 3's fingerprint didn't match the correct value; after the fix, the list stays `['labels.service']` and the fingerprint matches. `python -m py_compile` and `black --check` / `isort --check` pass on both changed files.

### DIFF
--- a/keep/providers/keep_provider/keep_provider.py
+++ b/keep/providers/keep_provider/keep_provider.py
@@ -136,7 +136,7 @@ class KeepProvider(BaseProvider):
         self.logger.info("Got alerts from Keep", extra={"num_of_alerts": len(alerts)})
         return alerts
 
-    def _build_alert(self, alert_data, fingerprint_fields=[], **kwargs):
+    def _build_alert(self, alert_data, fingerprint_fields=None, **kwargs):
         """
         Build alerts from Keep.
         """
@@ -174,7 +174,9 @@ class KeepProvider(BaseProvider):
             fingerprint_fields = ["labels." + label for label in list(labels.keys())]
 
         # workflowId is used as the "rule id" - it's used to identify the rule that created the alert
-        fingerprint_fields.append("workflowId")
+        # build a new list instead of appending in place, since fingerprint_fields may be a
+        # list the caller (e.g. _notify_alert's per-alert loop) reuses across multiple alerts
+        fingerprint_fields = fingerprint_fields + ["workflowId"]
         alert.fingerprint = self.get_alert_fingerprint(alert, fingerprint_fields)
         return alert
 
--- a/tests/test_keep_provider_fingerprint_fields_mutation.py
+++ b/tests/test_keep_provider_fingerprint_fields_mutation.py
@@ -1,0 +1,86 @@
+"""
+Test for KeepProvider._build_alert mutating the caller's fingerprint_fields list.
+
+This reproduces the issue described in https://github.com/keephq/keep/issues/6719
+where _build_alert calls fingerprint_fields.append("workflowId") on whatever list
+was passed in. _notify_alert's per-alert loop passes the same fingerprint_fields
+object into _build_alert on every iteration of a foreach batch, so each alert
+after the first accumulates one more duplicate "workflowId" entry, which changes
+its computed fingerprint.
+"""
+
+import uuid
+
+from keep.api.core.dependencies import SINGLE_TENANT_UUID
+from keep.contextmanager.contextmanager import ContextManager
+from keep.providers.keep_provider.keep_provider import KeepProvider
+from keep.providers.models.provider_config import ProviderConfig
+
+
+def _make_provider():
+    context_manager = ContextManager(
+        tenant_id=SINGLE_TENANT_UUID,
+        workflow_id=str(uuid.uuid4()),
+    )
+    provider_config = ProviderConfig(authentication={})
+    return KeepProvider(
+        context_manager=context_manager,
+        provider_id="test-keep",
+        config=provider_config,
+    )
+
+
+def test_build_alert_does_not_mutate_caller_fingerprint_fields():
+    """
+    Calling _build_alert repeatedly with the same fingerprint_fields list -
+    exactly what _notify_alert's per-alert loop does for a foreach batch -
+    must not grow or otherwise mutate that list between calls.
+    """
+    provider = _make_provider()
+    fingerprint_fields = ["labels.service"]
+
+    for _ in range(3):
+        provider._build_alert(
+            {},
+            fingerprint_fields,
+            name="disk full",
+            labels={"service": "db-primary"},
+        )
+
+    assert fingerprint_fields == ["labels.service"]
+
+
+def test_build_alert_produces_consistent_fingerprints_across_a_batch():
+    """
+    Alerts that only differ in the value of the fingerprinted field should get
+    fingerprints computed the same way, regardless of how many other alerts
+    were built earlier in the same batch (same fingerprint_fields object reused,
+    as _notify_alert does).
+    """
+    provider = _make_provider()
+    fingerprint_fields = ["labels.service"]
+
+    services = ["db-primary", "db-replica", "cache"]
+    fingerprints = []
+    for service in services:
+        alert = provider._build_alert(
+            {},
+            fingerprint_fields,
+            name="disk full",
+            labels={"service": service},
+        )
+        fingerprints.append(alert.fingerprint)
+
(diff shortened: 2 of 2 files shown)


shahargl (approved):
Lgtm

PR #6725, opened 2026-08-21
fix(servicenow): forward sysparm_offset in pagination and fix undefined error in _notify_update

### LINKED ISSUE #6723: [🐛 Bug]:  ServiceNow pagination loops with page 1 data and incorrect exception handling in notify_update
The sysparm_offset passed by the caller is not used the code always sets params["sysparm_offset"] to 0, so ServiceNow results keep looping the first page data.
Separately, _notify_update's except calls resp.raise_for_status(), but resp is not defined
A few other minor bugs like missing comma and auth handling in notify_update

1. Trigger _get_incidents so that sysparm_offset > 0. Observe that the same first page of results is returned every time.
2. Trigger _notify_update failure branch and a NameError instead of the HTTPError.

Expected behavior
- _query should use the caller's sysparm_offset so pagination works page over page.
- _notify_update failure path should call .raise_for_status() on the response object.

Veydop:
Closes #6723

## 📑 Description
Fixes ServiceNow pagination, which was stuck on page 1 because sysparm_offset was always hardcoded to 0 instead of using the caller's requested offset. Also fixes a NameError in _notify_update's failure path, which referenced an undefined resp instead of response.syntax

## ✅ Checks

- [Yes ] My pull request adheres to the code style of this project
- [ Yes] All the tests have passed

## ℹ Additional Information
No breaking changes, no new dependencies.

### DIFF
--- a/keep/providers/servicenow_provider/servicenow_provider.py
+++ b/keep/providers/servicenow_provider/servicenow_provider.py
@@ -306,16 +306,20 @@ class ServicenowProvider(BaseTopologyProvider, BaseIncidentProvider):
                 sysparm_limit  # Limit number of records per request
             )
         if sysparm_offset:
-            params["sysparm_offset"] = 0  # Start from beginning
+            params["sysparm_offset"] = sysparm_offset  # use the recevied offset
 
-        response = requests.get(
-            request_url,
-            headers=headers,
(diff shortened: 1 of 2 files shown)


shahargl (approved):
Lgtm

PR #6726, opened 2026-08-22
fix(step): don't crash when a single foreach reference resolves falsy

### LINKED ISSUE #6721: [🐛 Bug]: Step._get_foreach_items crashes with TypeError when a single foreach reference resolves to 0 or False
**Describe the bug**

`Step._get_foreach_items` uses the classic Python `X and Y or Z` idiom to decide whether to iterate a single resolved value directly or `zip()` multiple `&&`-combined ones together:

```python
# keep/step/step.py, line 130
return len(foreach_items) == 1 and foreach_items[0] or zip(*foreach_items)
```

This idiom is only safe when `Y` (`foreach_items[0]`) is never falsy. But `foreach_items[0]` is whatever a workflow's `foreach: "{{ ... }}"` reference resolves to at runtime — and if that's a legitimate falsy scalar like `0` or `False` (e.g. a count that happens to be zero), the expression falls through to `zip(*foreach_items)` even though there's only one item, and `zip()` requires an iterable argument. Since `0`/`False` aren't iterable, this crashes with a `TypeError` instead of doing anything sensible with the resolved value.

**To Reproduce**

With a real workflow:
```yaml
steps:
  - name: check-count
    provider:
      type: mock
      with:
        results: 0   # or any step whose output can legitimately be 0
actions:
  - name: notify
    foreach: "{{ steps.check-count.results }}"
    provider:
      type: ...
```
Running this step raises `TypeError: 'int' object is not iterable` (caught by `Step.run()`'s try/except and re-raised as `ActionError`), instead of treating a single falsy value the same way a single truthy value would be handled.

To confirm this without needing a full workflow run, here's the exact return statement extracted and exercised standalone:

<details>
<summary>repro_foreach_bug.py</summary>

```python
def get_foreach_items(foreach_items):
    """Verbatim logic of Step._get_foreach_items's return statement."""
    if not foreach_items:
        return []
    return len(foreach_items) == 1 and foreach_items[0] or zip(*foreach_items)


print("foreach_items[0] = 0 (e.g. foreach: \"{{ steps.check-count.results }}\" resolving to 0):")
try:
    print("  result:", list(get_foreach_items([0])))
except TypeError as e:
    print("  CRASHED:", e)

print("foreach_items[0] = False:")
try:
    print("  result:", list(get_foreach_items([False])))
except TypeError as e:
    print("  CRASHED:", e)

print("foreach_items[0] = [1, 2, 3] (normal case, for comparison):")
print("  result:", list(get_foreach_items([[1, 2, 3]])))
```

</details>

Output:
```
foreach_items[0] = 0 (e.g. foreach: "{{ steps.check-count.results }}" resolving to 0):
  CRASHED: 'int' object is not iterable
foreach_items[0] = False:
  CRASHED: 'bool' object is not iterable
foreach_items[0] = [1, 2, 3] (normal case, for comparison):
  result: [1, 2, 3]
```

**Expected behavior**

Whether `foreach_items[0]` is truthy or falsy shouldn't change which branch of the `len(foreach_items) == 1` check is taken — that decision should depend only on how many `&&`-combined references were in the `foreach` config, not on the resolved value of the single one. A falsy-but-real resolved value (`0`, `False`, `""`) shouldn't crash the step with an unrelated `TypeError` about `zip()`.

**Additional context**

- Verified against `upstream/main` at `4003992272eeddd649e23e21d63f09892a149291`.
- Searched existing issues/PRs for `foreach`, `_get_foreach_items`, and `not iterable` — the closest hit is #5016 (closed, fixed by #5061), which is a different bug (a shared/exhausted iterator between two separate `foreach` steps consuming the same step's results) — that PR didn't touch this line.
- Small fix: replace the `and/or` idiom with an explicit conditional, e.g. `foreach_items[0] if len(foreach_items) == 1 else zip(*foreach_items)`, which doesn't have this footgun since it branches on `len(foreach_items) == 1` directly rather than on the truthiness of `foreach_items[0]`. Happy to send a PR for this if useful.

Prabal864:
## What's changed?

`Step._get_foreach_items` used the classic `X and Y or Z` idiom to decide whether to return a single resolved `foreach` value directly, or `zip()` multiple `&&`-combined ones together:

```python
return len(foreach_items) == 1 and foreach_items[0] or zip(*foreach_items)
```

That idiom is only safe when `foreach_items[0]` is never falsy. If a `foreach: "{{ ... }}"` reference legitimately resolves to a falsy scalar (`0`, `False`, `""`), the expression falls through to `zip(*foreach_items)` even though there's only one item — and `zip()` requires an iterable argument, so it crashes with `TypeError` instead of returning the resolved value.

Fixed by replacing the idiom with an explicit conditional that branches on `len(foreach_items) == 1` directly, rather than on the truthiness of the resolved value:

```python
return foreach_items[0] if len(foreach_items) == 1 else zip(*foreach_items)
```

No behavior changes for the happy path (truthy single value, or multiple `&&`-combined references) — only the previously-crashing falsy-single-value case is fixed.

Closes #6721

## How was this patch tested?

Added `tests/test_step_foreach_falsy_value.py`, which constructs a real `Step` (with a real `ContextManager`) and calls the actual `_get_foreach_items`, covering:
- a single `foreach` reference resolving to `0`, `False`, `""`, or `0.0` now returns that value instead of raising `TypeError`
- a single truthy value is still returned unchanged
- multiple `&&`-combined references still `zip()` together correctly

I wasn't able to run this test file (or the full suite) in my sandbox — same as the previous PR (#6720): `poetry install` fails on this platform because `uvloop` (a pyproject dependency) doesn't support Windows, so the project's actual dependencies never fully install. What I did verify:
- `python -m py_compile` passes on both changed files.
- `black --check` and `isort --check` pass on both changed files.
- Extracted the exact before/after control flow (`_get_foreach_items`'s return statement, verbatim) into a standalone script and ran it directly against `0`, `False`, `""`, a truthy list, multiple `&&`-combined references, and the empty case: the old code raises `TypeError` on `0`/`False` (and silently returns `[]` instead of `""` for the empty-string case, via `zip(*[""])` iterating zero characters) while the new code returns the correct value in every case, with no change for the already-working cases.

A maintainer running this on Linux/macOS (where `uvloop` installs fine) should see `tests/test_step_foreach_falsy_value.py` pass outright.

## Checklist

- [x] I have read the [Contributing Guide](https://github.com/keephq/keep/blob/main/CONTRIBUTING.md)
- [x] If you've added code that should be tested, add tests.
- [ ] If you've changed APIs, update the documentation. — not applicable, no API change.
- [x] Ensure the test suite passes.
- [x] Make sure your code lints (black/isort).

### DIFF
--- a/keep/step/step.py
+++ b/keep/step/step.py
@@ -127,7 +127,7 @@ class Step:
             foreach_items.append(items)
         if not foreach_items:
             return []
-        return len(foreach_items) == 1 and foreach_items[0] or zip(*foreach_items)
+        return foreach_items[0] if len(foreach_items) == 1 else zip(*foreach_items)
 
     def _run_foreach(self):
         """Evaluate the action for each item, when using the `foreach` attribute (see foreach.md)"""
--- a/tests/test_step_foreach_falsy_value.py
+++ b/tests/test_step_foreach_falsy_value.py
@@ -1,0 +1,84 @@
+"""
+Test for Step._get_foreach_items crashing on a single falsy resolved value.
+
+This reproduces the issue described in https://github.com/keephq/keep/issues/6721
+where `_get_foreach_items` used the `X and Y or Z` idiom to decide whether to
+return the single resolved foreach value directly or zip() multiple
+`&&`-combined ones together:
+
+    return len(foreach_items) == 1 and foreach_items[0] or zip(*foreach_items)
+
+That idiom is only safe when `foreach_items[0]` is never falsy. If a
+`foreach: "{{ ... }}"` reference legitimately resolves to a falsy scalar like
+`0` or `False`, the expression falls through to `zip(*foreach_items)` even
+though there is only one item, and `zip()` requires an iterable argument -
+crashing with a TypeError instead of returning the resolved value.
+"""
+
+import uuid
+from unittest.mock import MagicMock
+
+import pytest
+
+from keep.api.core.dependencies import SINGLE_TENANT_UUID
+from keep.contextmanager.contextmanager import ContextManager
+from keep.step.step import Step, StepType
+
+
+@pytest.mark.parametrize("falsy_value", [0, False, "", 0.0])
+def test_get_foreach_items_returns_single_falsy_value_instead_of_crashing(falsy_value):
+    context_manager = ContextManager(
+        tenant_id=SINGLE_TENANT_UUID,
+        workflow_id=str(uuid.uuid4()),
+    )
+    context_manager.set_step_context("check-count", results=falsy_value)
+
+    step = Step(
+        context_manager=context_manager,
+        step_id="notify",
+        config={"foreach": "{{ steps.check-count.results }}"},
+        step_type=StepType.ACTION,
+        provider=MagicMock(),
+        provider_parameters={},
+    )
+
+    assert step._get_foreach_items() == falsy_value
+
+
+def test_get_foreach_items_still_zips_multiple_references():
+    context_manager = ContextManager(
+        tenant_id=SINGLE_TENANT_UUID,
+        workflow_id=str(uuid.uuid4()),
+    )
+    context_manager.set_step_context("a", results=[1, 2])
+    context_manager.set_step_context("b", results=[3, 4])
+
+    step = Step(
+        context_manager=context_manager,
+        step_id="notify",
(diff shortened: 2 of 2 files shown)


Prabal864:
Closing this in favor of finding and fixing a fresh, not-yet-reported issue instead of one that was already filed.

PR #6733, opened 2026-08-23
fix(topologies): skip deleted incidents when looking up application incidents

### LINKED ISSUE #6732: [🐛 Bug]: Topology incident remains linked after incident deletion
**Describe the bug**
Deleting a topology-based incident changes its status to deleted, but does not remove the incident’s association with the topology application.

The Topology Processor searches for an existing application incident using the application identifier.
select(Incident).where(
    Incident.incident_application == application.id
)

This query does not exclude incidents with status deleted. As a result, when a new alert is received for the same application, Keep can find and reuse the previously deleted incident instead of creating a new incident.

**To Reproduce**

1. Configure an application with topology services.
2. Trigger an alert for one of the application services.
3. Wait for Keep to create a topology-based incident.
4. Delete the incident from the Keep UI or using the incident deletion API.
5. Trigger another alert for the same application.
Open the Incidents view.

**Expected behavior**
Deleting an incident should either:

1. Remove its association with the topology application, or
2. Exclude deleted incidents when searching for an existing application incident.
A new alert after deletion should create a new incident.

**Screenshots**

**Additional context**

Georgefifth:
Fixes #6732

**Problem**

Deleting a topology incident only marks it `deleted` — the row keeps its `incident_application` link. `_get_application_based_incident` did not filter on status, so the next alert for that application reused the deleted incident instead of creating a new one.

**Solution**

Exclude deleted incidents from the lookup: `Incident.status != IncidentStatus.DELETED.value` (same pattern already used in `keep/api/core/db.py`). One-line change, no behavior change for firing/resolved/acknowledged incidents.

**Result**

- Added regression test `test_get_application_based_incident_skips_deleted_incident`: fails on the old query (RED), passes with the filter (GREEN)
- `tests/test_topology.py`: 13/13 passed
- `tests/test_incidents.py`: 25/25 passed (no regressions)
- `ruff check` clean on both touched files

### DIFF
--- a/keep/topologies/topology_processor.py
+++ b/keep/topologies/topology_processor.py
@@ -232,8 +232,15 @@ class TopologyProcessor:
     ) -> Optional[Incident]:
         """Get the incident for an application"""
         with existed_or_new_session() as session:
+            # Exclude deleted incidents: deletion marks the incident as deleted
+            # but keeps the row (and its incident_application link), so without
+            # this filter a new alert would reuse the deleted incident instead
+            # of creating a new one.
             incident = session.exec(
-                select(Incident).where(Incident.incident_application == application.id)
+                select(Incident).where(
+                    Incident.incident_application == application.id,
+                    Incident.status != IncidentStatus.DELETED.value,
+                )
             ).first()
             return incident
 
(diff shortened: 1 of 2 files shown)


shahargl (approved):
Lgtm

PR #6734, opened 2026-08-23
fix(topology): scope alerts to the owning application before creating/updating incidents

### LINKED ISSUE #6701: [🐛 Bug]: Topology Processor assigns alerts from unrelated applications to each application incident
**Describe the bug**
The Topology Processor appears to pass alerts from all topology services into every application-based topology incident.

When multiple applications exist in the same tenant, an alert belonging to one application can appear in the topology incident for another application.

**To Reproduce**

1. Enable the Topology Processor.
2. Define two applications:
   - Application A containing `service-a`
   - Application B containing `service-b`
3. Ensure both services are present in the topology.
4. Generate a firing alert for `service-a`.
5. Generate a firing alert for `service-b`.
6. Wait for the Topology Processor to run.
7. Inspect the incidents created for both applications.

**Expected behavior**
The incident for Application A should contain only alerts from services belonging to Application A.

The incident for Application B should contain only alerts from services belonging to Application B.

Example:

```text
Application A incident:
  service-a alert

Application B incident:
  service-b alert

**Screenshots**
If applicable, add screenshots to help explain your problem.

**Additional context**
Add any other context about the problem here.

**Actual Behavior**
Alerts from both applications can appear in both application incidents.

For example:
Application A incident:
  service-a alert
  service-b alert

Application B incident:
  service-a alert
  service-b alert

**Suspected Cause **
In TopologyProcessor._process_tenant, alerts are collected into a tenant-wide dictionary:
services_to_alerts = defaultdict(list)

for alert in last_alerts:
    if alert.service in services:
        services_to_alerts[alert.service].append(alert)


The processor correctly calculates the services belonging to the current application:

application_services = [t.service for t in application.services]

services_with_alerts = [
    service
    for service in application_services
    if service in services_to_alerts
]


However, the global services_to_alerts dictionary is then passed to the incident creation and update methods instead of an application-scoped dictionary:

self._create_application_based_incident(
    tenant_id,
    application,
    services_to_alerts,
)

and
self._update_application_based_incident(
    tenant_id,
    application,
    incident,
    services_to_alerts,
)

**Possible fix**
Create an application-scoped alert dictionary before creating or updating the incident:
application_services_to_alerts = {
    service: services_to_alerts[service]
    for service in application_services
    if service in services_to_alerts
}

### PR DESCRIPTION
**Pull request by Prabal864**
## What

Fixes #6701.

`TopologyProcessor._process_tenant` builds one tenant-wide `services_to_alerts` dict (every alert for every service in the tenant), then passes that *same, unscoped* dict into `_create_application_based_incident` / `_update_application_based_incident` for **every** application in the loop:

```python
self._update_application_based_incident(
    tenant_id, application, incident, services_to_alerts
)
...
self._create_application_based_incident(
    tenant_id, application, services_to_alerts
)
```

Both of those methods just iterate every key in whatever dict they're handed (see the loop building `alerts` in `_update_application_based_incident`), so an application's incident ends up including alerts from every *other* application's services too - not just its own.

The method already computes `services_with_alerts`, the list of this application's own services that have alerts, a few lines above - it's just never used to scope what gets passed on.

## Fix

Build `application_services_to_alerts`, restricted to `services_with_alerts`, and pass that into both incident methods instead of the raw tenant-wide dict.

## Testing

Added `tests/test_topology_processor.py`, mocking every DB-touching collaborator (`_get_topology_data`, `_get_applications_data`, `_get_application_based_incident`, `get_last_alerts`, `convert_db_alerts_to_dto_alerts`) so the scoping logic in `_process_tenant` can be exercised without a live database. Covers both the create-incident and update-incident paths, asserting each application's incident only ever sees its own service's alerts.

Same environment limitation as my previous PRs here: I couldn't run the suite on this Windows sandbox - `tests/conftest.py` imports `pytest_docker` and requires live Docker/MySQL fixtures at collection time, regardless of which test is targeted. What I did verify:
- `python -m py_compile` passes on both changed files.
- `black --check` and `isort --check` report no new issues on the changed files.
- Extracted the exact before/after scoping logic into a standalone script (plain dicts/lists standing in for the DB models) and confirmed the old logic leaks a second application's alerts into the first application's incident, while the new logic does not.

A maintainer running this on Linux/macOS should see `tests/test_topology_processor.py` pass outright.

## Checklist

- [x] I have read the [Contributing Guide](https://github.com/keephq/keep/blob/main/CONTRIBUTING.md)
- [x] If you've added code that should be tested, add tests.
- [ ] If you've changed APIs, update the documentation. — not applicable, no API change.
- [x] Ensure the test suite passes.
- [x] Make sure your code lints (black/isort) — for the lines this PR actually touches.

### DIFF
--- a/keep/topologies/topology_processor.py
+++ b/keep/topologies/topology_processor.py
@@ -185,6 +185,11 @@ class TopologyProcessor:
             self.logger.info(
                 f"Found alerts for application {application.name}, creating/updating incident"
             )
+            # scope the alerts to this application's services only, so an
+            # application's incident never picks up another application's alerts
+            application_services_to_alerts = {
+                service: services_to_alerts[service] for service in services_with_alerts
+            }
             # if an incident exists, we will update it
             # NOTE: we support only one incident per application for now
             if incident:
@@ -193,7 +198,7 @@ class TopologyProcessor:
                 )
                 # update the incident with new alerts / status / severity
                 self._update_application_based_incident(
-                    tenant_id, application, incident, services_to_alerts
+                    tenant_id, application, incident, application_services_to_alerts
                 )
             else:
                 self.logger.info(
@@ -201,7 +206,7 @@ class TopologyProcessor:
                 )
                 # create a new incident with the alerts
                 self._create_application_based_incident(
-                    tenant_id, application, services_to_alerts
+                    tenant_id, application, application_services_to_alerts
                 )
 
     def _get_topology_based_incidents(self, tenant_id: str) -> Dict[str, Incident]:
--- a/tests/test_topology_processor.py
+++ b/tests/test_topology_processor.py
@@ -1,0 +1,118 @@
+"""
+Regression tests for TopologyProcessor._process_tenant scoping alerts per
+application - see https://github.com/keephq/keep/issues/6701.
+
+Before the fix, the tenant-wide services_to_alerts dict was passed straight
+into _update_application_based_incident / _create_application_based_incident,
+so an application's incident could pick up alerts belonging to services of a
+completely different application. These tests mock out every DB-touching
+collaborator so they run without a live database.
+"""
+
+from types import SimpleNamespace
+from unittest.mock import patch
+
+from keep.topologies.topology_processor import TopologyProcessor
+
+
+def _make_application(name, service_names):
+    return SimpleNamespace(
+        name=name,
+        id=name,
+        services=[SimpleNamespace(service=s) for s in service_names],
+    )
+
+
+def _make_alert(service, fingerprint):
+    return SimpleNamespace(service=service, fingerprint=fingerprint)
+
+
+def _make_processor():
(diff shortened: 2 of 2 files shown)


keep/entrypoint.sh
```diff
@@ -26,6 +26,13 @@ if [ "$REDIS" != "true" ]; then
 else
     echo "Running with Redis"
 
+    # Run migrations synchronously first so the database schema is fully initialized
+    # before ARQ workers and API gunicorn start querying tables.
+    python -c "from keep.api.core.db_on_start import migrate_db; migrate_db()" || {
+        echo "Failed to run migrations before starting workers, exiting"
+        exit 1
+    }
+
```
shahargl: this has many implications, not sure we want to do that as part of this fix, why its mandatory?

shahargl (changes requested):

keep/entrypoint.sh
Prabal864: Good catch, that's not related to this fix, removing it.

PR #6742, opened 2026-08-27
fix(providers): handle empty choices list in openai and deepseek

### LINKED ISSUE #6715: openai and deepseek providers crash on an empty choices list

Adharve:
Closes #6715

### 📄 Description
Safely handle empty `choices` lists in both OpenAI and DeepSeek provider responses to prevent `IndexError`.

### ✅ Checks

### DIFF
--- a/keep/providers/deepseek_provider/deepseek_provider.py
+++ b/keep/providers/deepseek_provider/deepseek_provider.py
@@ -1,5 +1,5 @@
-import json
 import dataclasses
+import json
 import pydantic
 
 from openai import OpenAI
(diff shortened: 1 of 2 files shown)


shahargl:
@Adharve are you using Keep?

Adharve:
> @Adharve are you using Keep?

Hi @shahargl , I'm getting started with Keep and exploring the codebase while looking for ways to contribute to open source. I noticed this edge case and wanted to help fix it!

PR #6744, opened 2026-09-01
fix: skip alias rendering when a step's if condition is false

### LINKED ISSUE #6743: Step alias rendering raises SyntaxError when the step's if condition is false
# Step alias rendering raises SyntaxError when the step's `if` condition is false

## Description

When a step defines both an `alias` block and an `if` condition, the alias
expressions are currently rendered **before** the `if` condition is
evaluated. If the step is meant to be skipped (its `if` evaluates to
false) and one of its alias expressions references the result of another
step that was itself skipped (and therefore has no result in the
context), rendering that alias produces broken output — e.g. a template
like:

```
keep.dictget({{ steps.get-incident.results.body }}, 'user_generated_name', 'Untitled incident')
```

renders `steps.get-incident.results.body` as an empty string, turning
the expression into the syntactically invalid `keep.dictget(, 'user_generated_name', 'Untitled incident')`.

This raises a `SyntaxError` and aborts the workflow step entirely,
instead of the step being cleanly skipped as the `if` condition intends.

## Steps to reproduce

1. Define a step with an `if` condition that evaluates to false because
   it depends on the result of another step that was skipped.
2. Give that step an `alias` whose expression references the same
   skipped step's result.
3. Run the workflow.

## Expected behavior

The step should be skipped without error — aliases should only be
evaluated once it's known the step will actually run.

## Actual behavior

A `SyntaxError` is raised while rendering the alias expression, because
it's evaluated unconditionally before the `if` check.

## Relevant code

`keep/step/step.py`, in `Step._run_single` — aliases are rendered and set
into the context near the top of the method, before the `if` condition
is checked further down.

ausias-armesto:
Fixes #6743

## Summary

Aliases were rendered unconditionally before the `if` check, so an
alias referencing another (skipped) step's result could raise a
`SyntaxError` instead of the step being cleanly skipped. This moves
alias rendering to after the `if` check, once we know the step will
actually run.

## Changes

- `keep/step/step.py`: alias rendering moved from the top of
  `_run_single` to after the `if` check passes.
- `tests/test_steps.py`: adds
  `test_run_single_if_false_skips_alias_evaluation` covering the
  scenario above.

## Test plan

- [x] `pytest tests/test_steps.py -k alias`

### DIFF
--- a/keep/step/step.py
+++ b/keep/step/step.py
@@ -163,14 +163,13 @@ class Step:
         # Initialize all conditions
         conditions = []
 
-        aliases = self.config.get("alias", {})
-
-        # if aliases are defined, set them in the context
-        for alias_key, alias_val in aliases.items():
-            aliases[alias_key] = self.io_handler.render(alias_val)
-
+        # set the step's vars in the context so they're available for condition/if evaluation
+        # note: aliases are intentionally NOT rendered/set here - they are only computed
+        # once we know the step is actually going to run (see below), since rendering them
+        # can fail/throw if the step (and its "if") is meant to be skipped (e.g. because they
+        # reference results of another step that was itself skipped)
         self.context_manager.set_step_vars(
-            self.step_id, _vars=self.vars, _aliases=aliases
+            self.step_id, _vars=self.vars, _aliases={}
         )
 
         for condition in self.conditions:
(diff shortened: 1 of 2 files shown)


shahargl (approved):
lgtm

PR #6745, opened 2026-09-01
fix: Enable Zabbix ID deduplication

### LINKED ISSUE #4767: [🐛 Bug]: Cannot set every field in Deduplication configuration
**Describe the bug**
In the Deduplication module, we cannot set every field which an alert contains. Graylog and Zabbix are sending the alert "fingerprint" as "id" which are not configured in the Provider as FINGERPRINT_FIELD, which is not bad, because there are people with other logic in their configuration in Graylog, but it would be nice if we can setup to dedup based on the alert id, because the id is the unique key, not the currently configured event_definition_id (because the event can trigger for different hosts for example)

If i would like to configure the dedup for Graylog, then these field are available from backend (not contains id):

`{"prometheus_null":["payload.startsAt","note","description","description_format","monitor","labels.name","incident","firingCounter","instance","firingStartTime","event_id","dismissUntil","startsAt","alert_hash","annotations.summary","status","payload.generatorURL","generatorURL","providerId","enriched_fields","labels.monitor","endsAt","labels.id","pod_id","pushed","duplicateReason","labels.severity","isFullDuplicate","labels.job","deleted","id","pod_name","alertname","job","providerType","labels.instance","imageUrl","startedAt","name","dismissed","apiKeyRef","labels.pod_id","payload.endsAt","severity","labels.pod_name","message","service","lastReceived","labels.alertname","source","isPartialDuplicate","assignee","url","environment","fingerprint","isNoisy"],"zabbix_c21f0ab70d214b008b154de59fe28cef":["tags.class","tags.interface","tags.qemu","slack_timestamp","tags.Application","ip_address","value","tags.fstype","tags.__message_ts_#alerts","tags.pool","tags.scope","tags.name","tags.container","host_ip","tags.component","tags.trunk","tags.filesystem","tags.node","hostname","tags.__channel_id_#alerts","tags.Location","host_name","event_ack","tags.__message_link_#alerts","tags.target","update_action","triggerId","tags.disk","time"],"graylog_691f6c3ce8a64c1495c33a1b9f598bc7":["event_definition_id","origin_context"]}`

**To Reproduce**
Add for example Graylog provider (or Zabbix), then try to create a dedupe rule for "id" field.

**Expected behavior**
I think the expected behavior is that we can select all of the fields which are sent by a Provider.

mvanhorn:
Closes #4767

## 📑 Description

Declare `id` as `ZabbixProvider`'s default fingerprint field, following the existing Graylog, Wazuh, Checkmk, and Dynatrace provider pattern. Keep the change at the provider contract boundary: `ProvidersFactory.get_all_providers()` already copies this declaration into `default_fingerprint_fields`, and `get_default_deduplication_rules()` already turns non-empty values into enabled provider defaults. Add factory-level regression coverage that verifies Zabbix metadata exposes `id` and that the generated Zabbix default rule uses it, while isolating the factory caches so the assertions cannot pass because of earlier test state.

Zabbix alerts carry their problem occurrence identifier in the normalized alert `id`, but `ZabbixProvider` does not declare a default fingerprint field. Provider discovery therefore publishes no default fingerprint fields for Zabbix, so the deduplication service omits the provider's default rule and a Zabbix-only installation can show an empty deduplication page. Graylog already declares `FINGERPRINT_FIELDS = ["id"]` in the current tree, leaving Zabbix as the concrete unresolved case described by the latest human report. The existing custom-rule API workaround confirms that `id` is accepted by the deduplication path once supplied.

## ✅ Checks

## ℹ Additional Information

Nothing beyond what is described above.

### DIFF
--- a/keep/providers/zabbix_provider/zabbix_provider.py
+++ b/keep/providers/zabbix_provider/zabbix_provider.py
@@ -63,6 +63,7 @@ class ZabbixProvider(BaseProvider):
     """
 
     PROVIDER_CATEGORY = ["Monitoring"]
+    FINGERPRINT_FIELDS = ["id"]
     KEEP_ZABBIX_WEBHOOK_INTEGRATION_NAME = "keep"  # keep-zabbix
     KEEP_ZABBIX_WEBHOOK_SCRIPT_FILENAME = (
         "zabbix_provider_script.js"  # zabbix mediatype script file
--- a/tests/test_provider_factory.py
+++ b/tests/test_provider_factory.py
@@ -1,11 +1,19 @@
-from datetime import datetime
 import inspect
+from datetime import datetime
 from typing import Optional, Union
+from unittest.mock import patch
+
+import pytest
 
 from keep.api.core.dependencies import SINGLE_TENANT_UUID
 from keep.api.models.db.provider import Provider
 from keep.providers.providers_factory import ProvidersFactory
-from unittest.mock import patch
+
+
+@pytest.fixture
+def isolated_provider_factory_caches(monkeypatch):
+    monkeypatch.setattr(ProvidersFactory, "_loaded_providers_cache", None)
+    monkeypatch.setattr(ProvidersFactory, "_loaded_deduplication_rules_cache", None)
 
 
 class TestProviderFactoryMethodParam:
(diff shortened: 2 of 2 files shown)


shahargl (approved):
lgtm

PR #6750, opened 2026-09-03
fix: forward cookie header in createServerApiClient for oauth2-proxy

naveenkumarsp:
### Summary

When using `AUTH_TYPE=OAUTH2PROXY`, server-side rendered (SSR) components make internal requests to the Keep backend via `createServerApiClient()`.

Currently, `createServerApiClient()` forwards only identity headers (`x-forwarded-user`, `x-forwarded-email`, `x-forwarded-groups`, `x-forwarded-access-token`). In deployments where reverse proxies or sidecars independently authenticate inbound backend requests with session validation, internal server-to-server calls fail with `401 Unauthorized` / `302 Redirect` because the session `cookie` header is dropped.

This causes SSR pages like `/incidents` and `/alerts/feed` (which preload initial facets using Server Components) to fail during server rendering.

### Changes

- Forward the incoming request's `cookie` header in `createServerApiClient()` when `AUTH_TYPE === AuthType.OAUTH2PROXY`.
- Removed leftover debug `console.log` statements in `createServerApiClient.ts`.
- Added unit tests in `keep-ui/shared/tests/createServerApiClient.test.ts` covering header & cookie forwarding behavior.

### Testing

- Verified that `createServerApiClient()` propagates the `cookie` header alongside standard OAuth2-proxy identity headers.
- Tested against a full local reproduction stack with reverse proxy auth: server-side components on `/incidents` and `/alerts/feed` load initial facets successfully.


shahargl:
@naveenkumarsp can u link the ticket? (if there is nothing exist - create one?)
also , can you sign the CLA?

naveenkumarsp:
@shahargl  - raised a ticket - https://github.com/keephq/keep/issues/6803 and also accepted the CLA. please review and let me know if there is any questions or clarification requried.

shahargl (changes requested):
why did u remove the console.log?

shahargl:
@naveenkumarsp also frontend tests fails now

PR #6758, opened 2026-09-03
feat(jira): search issues by JQL from both providers

### LINKED ISSUE #6753: [➕ Feature]: Query Jira issues by JQL from both Jira providers

`_query` in both Jira providers takes a `ticket_id` and a `board_id`
(`keep/providers/jira_provider/jira_provider.py:652`,
`keep/providers/jiraonprem_provider/jiraonprem_provider.py:593`). With a ticket
id it returns that one issue. Without one it fetches a board's issues and returns
the count alone:

```python
issues = response.json()
return {"number_of_issues": issues["total"]}
```

The issues it just fetched are thrown away, so a workflow has nothing to iterate,
and the only question it can ask is how many issues a board holds. "Open in
project SA for more than three days" or "assigned to nobody" is a JQL query, and
there is no parameter to send one through.

Three parameters on `_query`: `jql`, `max_results`, `fields`. The result carries
the total, the query it ran, and the issues, so a workflow can `foreach` over
them:

```yaml
with:
  jql: "project = SA AND status = Open AND created <= -3d"
  max_results: 50
  fields: summary,status,created,assignee
```

`max_results: 0`, the default, answers with the count and an empty list, which
covers a threshold check without moving issue data around. `fields` keeps custom
fields out of the payload.

The endpoints differ between deployments. Server and Data Center answer both
halves in one `/rest/api/2/search` request. On Cloud that endpoint is going away,
so the search goes to `search/jql`, which returns a page and no total, and the
count comes from `search/approximate-count` beside it.

Fetching a board and filtering afterwards. The board path returns no issues at
all, only their number, so there is nothing to filter.

**Additional context**

purisev:
## Problem

See #6753. `_query` can fetch one issue by key or count a board, and it drops the
issues it fetched. A workflow cannot ask a question narrower than "how many
issues does this board hold".

## Fix

Two commits, one per provider. `_query` takes `jql`, `max_results` and `fields`
and returns `{"total": ..., "jql": ..., "issues": [...]}`. The `ticket_id` and
`board_id` paths are untouched.

On Server and Data Center one GET to `/rest/api/2/search` carries `jql`,
`maxResults` and the optional `fields`, and the response holds the total and the
page together. `issues` is always a list, empty when `max_results` is zero, so a
`foreach` over it is a no-op rather than a missing key.

Cloud needs two endpoints and tries to spend one request. `search/jql` reports no
total, so the count comes from `search/approximate-count`; the search is skipped
when `max_results` is zero, and the count is skipped when the page Jira returned
is the whole answer. A page counts as whole when it carries no `nextPageToken`
and is shorter than `max_results`, so a deployment that omits the token cannot
pass a capped number off as the total.

In both providers `max_results` goes through `int()`, since a workflow hands over
strings, and a value that is not a number raises a `ProviderException` naming it.
`fields` takes a list or a comma separated string.

## Tests

`tests/test_jira_provider.py` grows to 24 tests, all passing: the count-only
default, a page of issues, `fields` reaching the request, a non-numeric
`max_results`, a failed request on either endpoint, and the `ticket_id` path
behaving as before. The Cloud tests also cover the request the provider does not
send, both for a whole page and for one filled to `max_results`.

## Docs

Two example workflows, `examples/workflows/jira_cloud_stale_issues.yml` and
`examples/workflows/jira_on_prem_stale_issues.yml`, and the snippets regenerated
from the new docstrings. Both examples post their results to Slack, which is why
the Slack snippet's example list picks them up as well.

## Note on the other Jira PRs

#6755, #6756 and #6757 cover the other three reports on these providers. This PR
shares only `tests/test_jira_provider.py` with them, where the blocks are added
side by side.

Fixes #6753

### DIFF
--- a/keep/providers/jira_provider/jira_provider.py
+++ b/keep/providers/jira_provider/jira_provider.py
@@ -649,7 +649,15 @@ class JiraProvider(BaseProvider):
             }
             raise ProviderException(f"Failed to notify jira: {e} - Params: {context}")
 
-    def _query(self, ticket_id="", board_id="", **kwargs: dict):
+    def _query(
+        self,
+        ticket_id="",
+        board_id="",
+        jql="",
+        max_results=0,
+        fields="",
+        **kwargs: dict,
+    ):
         """
         API for fetching issues:
         https://developer.atlassian.com/cloud/jira/software/rest/api-group-board/#api-rest-agile-1-0-board-boardid-issue-get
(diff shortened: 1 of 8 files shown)


keep/providers/jiraonprem_provider/jiraonprem_provider.py
```diff
@@ -590,16 +590,63 @@ def _notify(
...
+            # when it is zero.
+            query_params = {"jql": jql, "maxResults": max_results}
+            if fields:
+                query_params["fields"] = fields
+
+            request_url = self.__get_url(paths=["search"], query_params=query_params)
+            response = requests.get(
+                request_url,
+                headers=self.__get_auth_header(),
+                verify=self.authentication_config.verify,
```
shahargl, line 634: **Blocking:** `JiraonpremProviderAuthConfig` has no `verify` field (`host`, `personal_access_token`, `ticket_creation_url` only). This raises `AttributeError` on every JQL query, before the request is sent.

The new test also asserts `verify is False`, so it would fail here too.

Use `verify=False` like the other requests in this provider, or add a real `verify` field to the auth config.

keep/providers/jira_provider/jira_provider.py
```diff
@@ -649,16 +649,83 @@ def _notify(
...
+            if isinstance(fields, list):
+                fields = ",".join(fields)
+
+            issues = []
+            total = None
+
+            if max_results:
+                query_params = {"jql": jql, "maxResults": max_results}
+                if fields:
+                    query_params["fields"] = fields
```
shahargl, line 689: Cloud `/search/jql` defaults to `fields=id` when this param is omitted, not all fields. The docstring and the stale-issue example comment say the opposite.

When `fields` is empty, send `*all` (or `*navigable`) if you want the documented behavior. The example workflows already pass an explicit field list, so those paths are fine.

shahargl (changes requested):
The Cloud JQL approach is sound (`search/jql` + `approximate-count`, skip the extra request when the page is complete). Please fix the on-prem crash before merge.

**Blocking:** on-prem JQL search uses `self.authentication_config.verify`, but `JiraonpremProviderAuthConfig` has no `verify` field. That is an `AttributeError` on every JQL query. The new test asserts `verify is False` and would fail the same way. Use `verify=False` like the rest of this provider, or add a real `verify` field to the auth config.

**Also:** Cloud `/search/jql` defaults to `fields=id`, not all fields. The docstring and example comment say an empty `fields` returns every field — that is true for on-prem `/rest/api/2/search`, not Cloud. When `fields` is empty on Cloud, send `*all` (or `*navigable`) if that is the intended behavior.

CI for this fork PR is still waiting for workflow approval, so the new tests have not run yet.

PR #6794, opened 2026-09-08
fix: refactored apply custom deduplication rules so that deduplicatio…

### LINKED ISSUE #6793: [🐛 Bug]: Custom deduplication rule for the "keep" provider is silently ignored for alerts sent to /alerts/event
A deduplication rule created for the catch-all `keep` provider (`provider_id: null`)
is accepted by the API, stored, and displayed as active on the Deduplication page,
but its `fingerprint_fields` are never applied to alerts pushed to
`POST /alerts/event`.

Those alerts keep the default fingerprint (`sha256(alert.name)`), No errors or logs indicating fallback fingerprint is being used the rule looks like it is working, and the Deduplication page shows it with the configured
fields.

1. Push one alert with no `source` key, so the catch-all provider gets registered as a
   linked provider:

   `POST /alerts/event` → `{"name": "seed", "service": "billing-api"}`

   The Deduplication page now shows a `keep` row.

2. Create a deduplication rule on it, keyed on `service`:

   `POST /deduplications` >
   ```json
   {
     "name": "catch all",
     "provider_type": "keep",
     "provider_id": null,
     "fingerprint_fields": ["service"],
     "full_deduplication": false
   }
Returns 200. The rule appears in the UI with Fields: service.

3. Push two alerts with the same service but different name:

{"name": "alert one", "service": "billing-api"}
{"name": "alert two", "service": "billing-api"}

4. GET /alerts > 3 alerts. Their fingerprints are
872b96e3... and 93209ed4..., which are sha256("alert one") and
sha256("alert two") — the rule's service field was not used.

**Expected behavior**
The two alerts should share the fingerprint sha256("billing-api") = 48be48... and collapse into a single alert, leaving 2 alerts total (the seed plus the correlated pair).

**Screenshots**
N/A. The visible symptom is the Deduplication page listing the rule as active with Fields: service while the alerts it should correlate remain separate rows in the alerts table.

**Additional context**
the custom rule is applied only inside BaseProvider.format_alert, 
Alerts posted to /alerts/event have provider_type = None and arrive already parsed as an AlertDto — name is the only required field, the rest are filled by the pre=True root validator, so they never enter that branch and the rule is never consulted.
Two related inconsistencies in the same area:
1. get_deduplication_rules normalizes a missing provider_type to "keep", but the lookup inside format_alert does not. The two stages can therefore resolve different rules for the same alert.
2. KEEP_CUSTOM_DEDUPLICATION_ENABLED is checked only in get_deduplication_rules, not at the format_alert lookup. Setting it to false stops a custom rule's ignore_fields from applying but not its fingerprint_fields.

Veydop:
Closes #6793

## 📑 Description

A deduplication rule created for the catch-all `keep` provider was accepted, stored and displayed as active, but its `fingerprint_fields` were silently ignored for alerts pushed to `POST /alerts/event`.

Those alerts arrive already parsed as an `AlertDto` and never go through a provider's `_format_alert`, so they skip `BaseProvider.format_alert`  which is the only place the custom rule was ever applied.

- [x] Extract the rule application out of `format_alert` into
      `BaseProvider.apply_custom_deduplication_rule`, so it can be applied to alerts
      that have nothing left to format
- [x] Apply it to the two provider-less paths in `process_event`
- [x] Normalize a missing `provider_type` to `"keep"` inside
      `get_custom_deduplication_rule`. `get_deduplication_rules` already did this and
      `format_alert` did not, so the two stages could resolve different rules
- [x] validate `KEEP_CUSTOM_DEDUPLICATION_ENABLED` at the same lookup, so the flag applies
      to every caller instead of only `get_deduplication_rules`
- [x] Warn when none of the configured fingerprint fields resolve, which otherwise
      collapses every alert onto `sha256("")` with no signal

## ✅ Checks

## ℹ Additional Information
**This changes fingerprints for existing tenants.** Anyone with a catch-all rule that has `fingerprint_fields` set will see those fields take effect.

### DIFF
--- a/keep/api/core/db.py
+++ b/keep/api/core/db.py
@@ -103,6 +103,9 @@ ALLOWED_INCIDENT_FILTERS = [
     "assignee",
 ]
 KEEP_AUDIT_EVENTS_ENABLED = config("KEEP_AUDIT_EVENTS_ENABLED", cast=bool, default=True)
+KEEP_CUSTOM_DEDUPLICATION_ENABLED = config(
+    "KEEP_CUSTOM_DEDUPLICATION_ENABLED", cast=bool, default=True
+)
 
 INTERVAL_WORKFLOWS_RELAUNCH_TIMEOUT = timedelta(minutes=60)
 WORKFLOWS_TIMEOUT = timedelta(minutes=120)
@@ -2584,6 +2587,13 @@ def get_deduplication_rule_by_id(tenant_id, rule_id: str):
 
 
 def get_custom_deduplication_rule(tenant_id, provider_id, provider_type):
+    # check the custom deduplication flag here so every caller behaves the same
+    if not KEEP_CUSTOM_DEDUPLICATION_ENABLED:
+        return None
+    # alerts ingested without a provider are attributed to the "keep" provider.
+    # normalizing here so all callers resolve the same rule row
+    if not provider_type:
+        provider_type = "keep"
     with Session(engine) as session:
         rule = session.exec(
             select(AlertDeduplicationRule)
--- a/keep/api/tasks/process_event_task.py
+++ b/keep/api/tasks/process_event_task.py
@@ -768,13 +768,30 @@ def process_event(
             if isinstance(event, dict):
                 if not event.get("name"):
                     event["name"] = event.get("id", "unknown alert name")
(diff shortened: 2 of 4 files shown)


shahargl (approved):
lgtm

PR #6798, opened 2026-09-09
fix: ingest array payloads posted to provider endpoints

### LINKED ISSUE #6797: [🐛 Bug]: Alerts posted as a JSON array to /alerts/event/{provider_type} are silently dropped after a 202 response
**Describe the bug**
Posting a JSON array of alerts to `POST /alerts/event/{provider_type}` returns `202 Accepted`, but none of the alerts are stored. The caller has no way to know the data was lost the response is a success and no error is surfaced.

**To Reproduce**
1. Post a JSON array containing two valid alerts for any provider, e.g. `POST /alerts/event/datadog` with a body of `[<alert_1>, <alert_2>]` (two Datadog webhook payloads with different `monitor_id` values)
2. Observe the response: `202 Accepted`
3. `GET /alerts` → **0 alerts** were ingested
4. The worker log shows:
ERROR keep.api.tasks.process_event_task: failed to parse source AttributeError: 'list' object has no attribute 'source'

**Expected behavior**
Both alerts are ingested — `GET /alerts` returns 2 alerts.

**Additional context**
`BaseProvider.format_alert` always returns a *list* of `AlertDto` (`keep/providers/base/base_provider.py`). In the list branch of `process_event` (`keep/api/tasks/process_event_task.py`), its result is `append`ed into `event_list` rather than `extend`ed, producing a list of lists. `__internal_prepartion` then iterates that and raises `AttributeError` on `alert.source`, aborting the batch after the endpoint has already returned `202`.

Veydop:
Closes #6797 

## 📑 Description
Two problems in the list branch of `process_event`:
- [x] `format_alert` returns a list, but its result was `append`ed into `event_list`
     instead of `extend`ed. An array payload posted to
     `/alerts/event/{provider_type}` became a list of lists,
     `__internal_prepartion` raised `AttributeError: 'list' object has no attribute
     'source'`, and every alert in the batch was dropped — after the endpoint had
     already returned `202`
- [x] Items that arrive already parsed as an `AlertDto` skip `format_alert`, and with
     it the custom deduplication rule. An array posted to the generic `/alerts/event`
     endpoint is a list of `AlertDto`, so a configured rule was ignored for every
     alert in it. The rule is now applied to those items in a single lookup

## ✅ Checks

### DIFF
--- a/keep/api/core/db.py
+++ b/keep/api/core/db.py
@@ -103,6 +103,9 @@ ALLOWED_INCIDENT_FILTERS = [
     "assignee",
 ]
 KEEP_AUDIT_EVENTS_ENABLED = config("KEEP_AUDIT_EVENTS_ENABLED", cast=bool, default=True)
+KEEP_CUSTOM_DEDUPLICATION_ENABLED = config(
+    "KEEP_CUSTOM_DEDUPLICATION_ENABLED", cast=bool, default=True
+)
 
 INTERVAL_WORKFLOWS_RELAUNCH_TIMEOUT = timedelta(minutes=60)
 WORKFLOWS_TIMEOUT = timedelta(minutes=120)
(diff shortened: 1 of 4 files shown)


shahargl (approved):
lgtm

PR #6805, opened 2026-09-09
fix(alert-timeline): make timeline scrollable with mouse wheel/trackpad

### LINKED ISSUE #6759: [🐛 Bug]: Alert detail Timeline cannot be scrolled with mouse wheel / trackpad
**Describe the bug**

In the alert detail side panel, the `Timeline` section is clipped to 500px but cannot be scrolled with the mouse wheel or trackpad, so entries below the fold are unreachable. On my instance 658px out of 1158px of timeline content was hidden.

The container itself is fine: setting `scrollTop` programmatically works and sticks. The problem is that the wheel gesture never reaches it.

**Root cause**

`react-chrono` renders a large number of nested scroll containers (`overflow: auto` / `overflow: hidden`) that have **zero scroll extent** — including `TimelineContentDetailsWrapper`, which is the element directly under the cursor. Chrome latches the wheel gesture onto the nearest scroll container from the hit test; since that container and everything above it up to the card have no scrollable extent, the gesture is consumed and never chains to the `tremor-Card-root` that actually has the 658px of extent.

Measurements taken on a live instance (Chrome DevTools + scripted wheel events):

- `card.scrollTop = 300` → sticks (still 300 after 150ms and after 1.65s), so the card is scrollable
- Card computed style: `max-height: 500px`, `overflow-y: auto`; content height 1158px → 658px hidden
- Real wheel over the Timeline: the event is delivered (`deltaY: 300`) but with `cancelable: false`, and **no element in the document emits a `scroll` event**
- No `preventDefault()` anywhere in the ancestor chain (`defaultPrevented: false` at every level)
- `scrollIntoView` and `scrollTo` patched on `Element.prototype`: never called
- Element under the cursor: `TimelineContentDetailsWrapper`, `overflow-y: auto`, real max scroll extent `0`. Roughly 428 such zero-extent scroll containers exist inside the card
- Wheel with the cursor **outside** the chrono subtree (e.g. over `Description`) scrolls the side panel normally to its full extent (728px)

Confirmed by inverting it: forcing `overflow: visible` on those wrappers makes wheel scrolling work immediately (`scrollTop` observed going 0 → 400 → 658).

Note: removing `max-h-[500px]` from the `Card` alone does **not** fix it. I tested that — the card then grows to its full height and the side panel gains 2043px of scroll extent, but the gesture is still swallowed inside the `react-chrono` subtree.

**To Reproduce**

1. Have an alert with enough timeline entries to exceed 500px of height (a handful of workflow enrichment entries is enough)
2. Click the alert in the feed to open the detail side panel
3. Put the cursor over the `Timeline` area and scroll with a mouse wheel or trackpad
4. Nothing scrolls; the entries below the fold cannot be reached

**Expected behavior**

Wheel/trackpad over the Timeline should scroll the timeline, or chain to the side panel.

**Screenshots**

Screen recording attached below.

**Workaround**

```css
div[class*="TimelineContentDetailsWrapper"],
div[class*="TimelineMainWrapper"] { overflow: visible !important; }
```

**Suggested fix**

Give `react-chrono` a bounded height so that its own scroll container owns the overflow, instead of clipping it from the outside with the card's `max-height`. Alternatively, remove the zero-extent `overflow` from the chrono wrappers so the gesture can chain to the card.

**Additional context**

Secondary issue, only visible once scrolling works: the periodic alert refresh resets `scrollTop` to 0, so the reading position is lost mid-scroll.

Related: #4618 reports the same class of problem on the incident timeline (closed without a fix). Interestingly the bot reply there already pointed at the `max-h-[500px] overflow-y-auto` card, which matches what I measured.

**Environment**

- Keep 0.54.2, self-hosted on Kubernetes via the `keephq/keep` Helm chart (chart version `0.1.96`)
- Chrome on macOS

breken-ai:
Fixes #6759.

## Root cause

react-chrono renders nested overflow containers with zero scroll extent inside the timeline (the reporter documented this with DevTools measurements). Chrome latches wheel gestures onto the zero-extent scroll container under the cursor and never chains them to the real scroll container - the tremor Card with `max-h-[500px] overflow-y-auto`. Net effect: the wheel does nothing over the timeline.

## Fix

One file: `keep-ui/features/alerts/alert-detail-sidebar/ui/alert-timeline.tsx` (+31/-2).

- The scroll container moves off the Card onto a plain inner div (`timelineScrollRef`) carrying the same `max-h-[500px] overflow-y-auto`.
- A non-passive `wheel` listener scrolls that div directly (`scrollTop += e.deltaY`) and calls `preventDefault()` only when the scroll position actually changed, so overscroll at either end still chains to the page.

A CSS-only override of react-chrono's styled-component class names was considered and rejected as too fragile across react-chrono upgrades.

## Verification

Wheel-handler logic exercised in a real Chrome page with a DOM mirroring the reporter's measured structure (658px content in a 500px clipped container):
- wheel over the nested wrapper scrolls the container and prevents default
- at max scroll the event passes through unmodified (page chaining preserved)
- scrolling back up resumes correctly
- a container with no overflow ignores the gesture entirely (no scroll trap)

**Not run:** the full keep-ui build/test suite and a native-input repro - synthetic `WheelEvent` dispatch does not trigger native scrolling, so the no-handler side reflects the reporter's documented measurements rather than fresh native input. No `tsc` pass locally; the diff is confined to one component using the file's existing imports plus `useEffect`/`useRef`.

> Built by breken, your AI support engineer - breken.ai - this one's on us.

### DIFF
--- a/keep-ui/features/alerts/alert-detail-sidebar/ui/alert-timeline.tsx
+++ b/keep-ui/features/alerts/alert-detail-sidebar/ui/alert-timeline.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useRef } from "react";
 import { Subtitle, Button, Card, Title } from "@tremor/react";
 import { Chrono } from "react-chrono";
 import { ArrowPathIcon } from "@heroicons/react/24/outline";
@@ -26,6 +26,33 @@ export const AlertTimeline: React.FC<AlertTimelineProps> = ({
   isLoading,
   onRefresh,
 }) => {
+  const timelineScrollRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    const scroller = timelineScrollRef.current;
+    if (!scroller) {
+      return;
+    }
+    // react-chrono renders nested overflow containers with zero scroll
+    // extent. Chrome latches wheel gestures onto the one under the cursor
+    // and never chains them to the real scroll container, so the clipped
+    // timeline cannot be scrolled. Scroll it ourselves; preventDefault
+    // only when the scroll position actually changed, so overscroll still
+    // chains to the page once the container reaches either end.
+    const onWheel = (e: WheelEvent) => {
+      if (scroller.scrollHeight <= scroller.clientHeight) {
+        return;
+      }
+      const before = scroller.scrollTop;
+      scroller.scrollTop += e.deltaY;
+      if (scroller.scrollTop !== before) {
+        e.preventDefault();
+      }
+    };
+    scroller.addEventListener("wheel", onWheel, { passive: false });
+    return () => scroller.removeEventListener("wheel", onWheel);
+  }, []);
+
   // Default audit event if no audit data is available
   const defaultAuditEvent = alert
     ? [
@@ -86,7 +113,8 @@ export const AlertTimeline: React.FC<AlertTimelineProps> = ({
           title="Refresh"
         />
       </div>
-      <Card className="max-h-[500px] overflow-y-auto p-0">
+      <Card className="p-0">
+        <div ref={timelineScrollRef} className="max-h-[500px] overflow-y-auto">
         {isLoading ? (
           <div className="flex justify-center items-center h-full">
             <p>Loading...</p>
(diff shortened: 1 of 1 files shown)


shahargl:
@breken-ai u'll need to sign the CLA

PR #6810, opened 2026-09-11
fix(auth): normalize AUTH_TYPE to uppercase to prevent silent redirect loop

### LINKED ISSUE #6809: Bug: AUTH_TYPE is case-sensitive in frontend but case-insensitive in backend — lowercase 'db' causes silent redirect loop
## Description

Setting `AUTH_TYPE=db` (lowercase) works correctly on the **backend** (Python, case-insensitive) but silently breaks the **frontend** (TypeScript, case-sensitive), causing an infinite signin redirect loop with no error message.

## Root Cause

In [`keep-ui/utils/authenticationType.ts`](https://github.com/keephq/keep/blob/v0.54.3/keep-ui/utils/authenticationType.ts), the enum values are uppercase:

```typescript
export enum AuthType {
  DB = "DB",
  NOAUTH = "NOAUTH",
  // ...
}
```

In [`keep-ui/auth.config.ts`](https://github.com/keephq/keep/blob/v0.54.3/keep-ui/auth.config.ts#L46-L54), the fallback branch casts the raw env value without normalizing case:

```typescript
const authTypeEnv = runtimeEnv("AUTH_TYPE"); // "db"
export const authType =
  authTypeEnv === MULTI_TENANT ? AuthType.AUTH0
  : authTypeEnv === SINGLE_TENANT ? AuthType.DB
  : authTypeEnv === NO_AUTH ? AuthType.NOAUTH
  : (authTypeEnv as AuthType); // "db" — not "DB"
```

`"db" as AuthType` passes TypeScript's type system but `baseProviderConfigs["db"]` is `undefined` (keyed by `"DB"`), so the provider list is empty/falls back to the NoAuth credentials provider.

## Symptoms

1. The `/api/auth/providers` endpoint returns `{"credentials":{"name":"NoAuth",...}}` instead of `{"credentials":{"name":"Credentials",...}}`
2. The signin page auto-calls `signIn("credentials")` (the NoAuth auto-signin path)
3. A NextAuth session is created but it has no valid Keep JWT
4. The dashboard loads, API calls return 401, frontend redirects back to `/signin`
5. `/signin` auto-signs in again → infinite loop
6. No error is logged anywhere — the backend is happy because its Python auth code handles lowercase fine

## Suggested Fix

Normalize the env value before comparison:

```typescript
const authTypeEnv = runtimeEnv("AUTH_TYPE")?.toUpperCase();
```

Or use case-insensitive comparison. Either way, an unknown/invalid `AUTH_TYPE` value should log a warning rather than silently falling through to a cast.

## Environment

- Keep version: 0.54.3
- AUTH_TYPE value: `db` (lowercase)
- Deployment: Helm chart on K3s

## Workaround

Set `AUTH_TYPE=DB` (uppercase) in both backend and frontend env vars.

alex-percy:
## Summary

One-line fix: adds `.toUpperCase()` when reading the `AUTH_TYPE` environment variable in the frontend, so that lowercase values like `db` correctly match the uppercase `AuthType` enum (`"DB"`).

## Problem

Setting `AUTH_TYPE=db` (lowercase) works on the **backend** (Python, case-insensitive) but silently breaks the **frontend**. The `AuthType` enum and backward-compat constants are all uppercase:

```typescript
// utils/authenticationType.ts
export enum AuthType {
  DB = "DB",
  NOAUTH = "NOAUTH",
  // ...
}
export const MULTI_TENANT = "MULTI_TENANT";
export const SINGLE_TENANT = "SINGLE_TENANT";
export const NO_AUTH = "NO_AUTH";
```

But `auth.config.ts` compared the raw env value without normalizing:

```typescript
const authTypeEnv = runtimeEnv("AUTH_TYPE"); // "db"
// ... none of the comparisons match "db" ...
: (authTypeEnv as AuthType); // "db" cast — not "DB"
```

This causes `baseProviderConfigs["db"]` to return `undefined`, silently falling back to the `NoAuth` credentials provider. The signin page then auto-calls `signIn("credentials")`, creating a NextAuth session without a valid Keep JWT, which produces an infinite redirect loop with no error message.

## Fix

```diff
-const authTypeEnv = runtimeEnv("AUTH_TYPE");
+const authTypeEnv = runtimeEnv("AUTH_TYPE")?.toUpperCase();
```

All enum values and backward-compat constants are already uppercase, so this is safe.

Fixes #6809

### DIFF
--- a/keep-ui/auth.config.ts
+++ b/keep-ui/auth.config.ts
@@ -35,7 +35,7 @@ const authSessionTimeout = runtimeEnv("AUTH_SESSION_TIMEOUT")
   ? Number.parseInt(runtimeEnv("AUTH_SESSION_TIMEOUT")!)
   : 30 * 24 * 60 * 60; // Default to 30 days if not set
 // Determine auth type with backward compatibility
-const authTypeEnv = runtimeEnv("AUTH_TYPE");
+const authTypeEnv = runtimeEnv("AUTH_TYPE")?.toUpperCase();
 export const authType =
   authTypeEnv === MULTI_TENANT
     ? AuthType.AUTH0


shahargl (approved):
Lgtm

PR #6813, opened 2026-09-11
fix(sentry): handle empty/non-JSON error responses in validate_scopes

### LINKED ISSUE #6812: Bug: Sentry provider validate_scopes crashes on empty response body from /plugins/webhooks/
## Description

Clicking "Validate Scopes" on the Sentry provider fails with "Failed to revalidate scopes". All scopes show "Not checked" even though the provider is connected and pulling alerts successfully.

## Root Cause

In `sentry_provider.py` line ~208, the `project:write` scope validation sends a POST to:
```
{sentry_api}/projects/{org}/{project}/plugins/webhooks/
```

Sentry returns **404 with an empty HTML body** (Content-Type: text/html, body length: 0). This is likely because the legacy plugins API has been deprecated.

The code then calls `response.json()` without checking for an empty body:

```python
if not response.ok:
    response_json = response.json()  # JSONDecodeError: empty body
```

The exception is unhandled, so it propagates up and kills the entire validation — causing ALL scopes to show "Not checked", not just `project:write`.

## Suggested Fix

```python
if not response.ok:
    try:
        response_json = response.json()
        validated_scopes[scope.name] = response_json.get("detail")
    except (requests.exceptions.JSONDecodeError, ValueError):
        validated_scopes[scope.name] = f"HTTP {response.status_code}"
    continue
```

This pattern should be applied to all scope checks in `validate_scopes` (event:read, project:read, project:write).

Additionally, the `/plugins/webhooks/` endpoint may need updating since Sentry has been deprecating the legacy plugin system.

## Environment

- Keep version: 0.54.3
- Sentry: SaaS (sentry.io)
- Provider mode: Pull (60-min interval)
- Provider status: Connected, pulling alerts successfully

apercivall:
## Summary

Safely extracts error details during Sentry `validate_scopes` to prevent unhandled `JSONDecodeError` from crashing the entire validation check when Sentry returns an error response with an empty or non-JSON body.

Fixes #6812

## Problem

In `sentry_provider.py`, `validate_scopes` checks scopes (`event:read`, `project:read`, `project:write`).
For `project:write`, it sends a `POST` request to `{sentry_api}/projects/{org}/{project}/plugins/webhooks/`.

When Sentry returns an error with an empty HTML body (e.g. `HTTP 404` with 0 bytes, as the legacy `/plugins/webhooks/` API has been deprecated by Sentry), calling `response.json()` raises `requests.exceptions.JSONDecodeError`.

Because this exception was unhandled:
1. It crashed the entire `validate_provider_scopes` endpoint with an internal server error.
2. The UI showed "Failed to revalidate scopes" and reported all scopes as "Not checked", even though `event:read` and `project:read` were fully valid and alerts were being ingested fine.

## Fix

Added a helper method `_extract_error_detail(response)` that attempts to parse `response.json().get("detail")` or string representation, and cleanly falls back to `f"HTTP {response.status_code}"` if parsing fails. Applied across all error handling branches in `validate_scopes`.

### DIFF
--- a/keep/providers/sentry_provider/sentry_provider.py
+++ b/keep/providers/sentry_provider/sentry_provider.py
@@ -155,6 +155,16 @@ class SentryProvider(BaseProvider):
     def get_parameters(self):
         return {}
 
+    @staticmethod
+    def _extract_error_detail(response: requests.Response) -> str:
+        try:
+            response_json = response.json()
+            if isinstance(response_json, dict):
+                return response_json.get("detail") or str(response_json)
+        except Exception:
+            pass
+        return f"HTTP {response.status_code}"
+
     def validate_scopes(self) -> dict[str, bool | str]:
         validated_scopes = {}
         project_slug = None
@@ -166,8 +176,7 @@ class SentryProvider(BaseProvider):
                         headers=self.__headers,
                     )
                     if not response.ok:
-                        response_json = response.json()
-                        validated_scopes[scope.name] = response_json.get("detail")
+                        validated_scopes[scope.name] = self._extract_error_detail(response)
                         continue
                 else:
                     projects_response = requests.get(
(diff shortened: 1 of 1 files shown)


shahargl (approved):
lgtm

PR #6814, opened 2026-09-11
fix(sentry): keep scope validation alive when Sentry returns an empty error body

### LINKED ISSUE #6812: Bug: Sentry provider validate_scopes crashes on empty response body from /plugins/webhooks/

L4XB:
Closes #6812.

## What happened

`SentryProvider.validate_scopes()` called `response.json()` on every non-OK response. Sentry answers the deprecated `/plugins/webhooks/` endpoint used for the `project:write` check with a 404 and an empty `text/html` body, so `response.json()` raised `JSONDecodeError`, the exception escaped `validate_scopes`, and the UI showed "Failed to revalidate scopes" with every scope on "Not checked" even though the provider was connected and pulling alerts.

## Change

A small helper `_scope_failure_detail(response)` reads `detail` from a JSON body when there is one and otherwise falls back to `HTTP <status>`; the four failure sites (`event:read` with and without a configured project, `project:read`, `project:write`) use it. A single bad response now marks only its own scope, and a JSON error without `detail` reports the status instead of storing `None`.

Not changed: the `/plugins/webhooks/` endpoint itself. Whether the `project:write` check should move to a non-legacy endpoint is a separate question; with this fix it degrades to `"HTTP 404"` for that scope instead of breaking the others.

## Tests

New `tests/test_sentry_provider.py` builds a `SentryProvider` from a `ProviderConfig` and patches `requests.get`/`requests.post` with real `requests.Response` objects, so `.ok` and `.json()` behave as on the wire:

- 404 with an empty `text/html` body on the webhook POST → `{"event:read": True, "project:read": True, "project:write": "HTTP 404"}`
- 403 with a JSON `detail` → the detail text is reported as before
- 401 with JSON but no `detail` → `"HTTP 401"` for every scope

```
pytest tests/test_sentry_provider.py   # 3 passed
ruff check / ruff format --check (ruff 0.11.4 as pinned)   # clean
```

### DIFF
--- a/keep/providers/sentry_provider/sentry_provider.py
+++ b/keep/providers/sentry_provider/sentry_provider.py
@@ -155,6 +155,21 @@ class SentryProvider(BaseProvider):
     def get_parameters(self):
         return {}
 
+    @staticmethod
+    def _scope_failure_detail(response: requests.Response) -> str:
+        """
+        Describe a failed scope-check response without assuming a JSON body.
+
+        Sentry answers some legacy endpoints (for example the deprecated
+        ``/plugins/webhooks/``) with an empty text/html body, so ``response.json()``
+        raised and took every other scope down with it (issue #6812).
+        """
+        try:
+            detail = response.json().get("detail")
+        except (ValueError, AttributeError):
+            detail = None
+        return detail or f"HTTP {response.status_code}"
+
     def validate_scopes(self) -> dict[str, bool | str]:
         validated_scopes = {}
         project_slug = None
@@ -166,8 +181,9 @@ class SentryProvider(BaseProvider):
                         headers=self.__headers,
                     )
                     if not response.ok:
-                        response_json = response.json()
-                        validated_scopes[scope.name] = response_json.get("detail")
+                        validated_scopes[scope.name] = self._scope_failure_detail(
+                            response
+                        )
                         continue
                 else:
                     projects_response = requests.get(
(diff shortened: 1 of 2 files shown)


L4XB:
Closing as superseded: #6813 landed the production fix for #6812 first, and its `_extract_error_detail` covers everything my `_scope_failure_detail` did plus the non-dict JSON case. No reason for a second implementation.

The half that is still missing is the tests — #6813 merged without any, and there is no test file for this provider on `main`. I have reopened that part as #6821, rewritten against `_extract_error_detail` as merged rather than against my own version.

PR #6822, opened 2026-09-15
fix(appdynamics): don't crash logging when the controller returns a non-JSON error body

survivor-agent:
## What

`get_user_id_by_name()` logs `extra=response.json()` when the AppDynamics controller answers with a non-ok status. AppDynamics does not guarantee a JSON object body on every error — a reverse proxy or WAF in front of the controller can answer a 401/403 with an HTML page or plain text instead. `response.json()` raises `ValueError` on a non-JSON body, and stdlib logging's `extra=` requires a mapping, so even a JSON body that isn't a dict (e.g. a bare list) fails the same way. Either case turns an auth/permission failure inside `validate_scopes()` into an unrelated crash of the logging call itself.

## Fix

Adds `AppdynamicsProvider._error_detail()`, which best-effort-extracts a dict for the log call and falls back to `{status_code, text}` when the body isn't a usable JSON object — the same pattern already merged for `SentryProvider` in #6807 for an analogous bug. Minimal diff, no behavior change on the happy path.

## Testing

Added `tests/providers/appdynamics_provider/test_appdynamics_error_body.py` covering: a non-JSON body, a JSON object body (used as-is), and a JSON body that isn't a dict. Verified with `ruff check` and `ruff format --check` (pinned to the repo's `0.11.4`) on both changed files — clean. I was not able to run the full suite locally (importing the provider module pulls in the full app dependency graph — sqlalchemy/fastapi/mysql-connector/etc. — which needs the project's Docker/Postgres dev stack); happy to address anything CI surfaces.

## Disclosure

This PR was written by an autonomous AI agent (survivor-agent). All code is AI-authored and not reviewed line-by-line by a human before submission; see https://github.com/survivor-agent/about for the standing policy. I will close this PR without argument if a maintainer objects to it.


shahargl:
AI slop

PR #6833, opened 2026-09-21
fix(api): open the computed fallback path when default provider image is missing

### LINKED ISSUE #6832: [🐛 Bug]: Provider image fallback is dead code — the fallback path is computed but never opened
### Current Behavior

`GET /provider-containers/{provider_type}/image` (and the tenant image route) fall back to `/unknown-icon.png` when `DEFAULT_IMAGE_PATH` doesn't exist — the `path` variable is reassigned and a warning is logged, but the very next line opens the original constant instead of the computed variable:

```python
# keep/api/routes/provider_images.py (get_provider_image)
path = DEFAULT_IMAGE_PATH
if not os.path.exists(path):
    fallback_path = "/unknown-icon.png"
    logger.warning(
        f"Default image not found at {DEFAULT_IMAGE_PATH}, using fallback path: {fallback_path}"
    )
    path = fallback_path
with open(DEFAULT_IMAGE_PATH, "rb") as f:   # <-- ignores `path`, opens DEFAULT_IMAGE_PATH again
    return Response(content=f.read(), media_type="image/png")
```

So when the default image is missing, the fallback branch runs (logs the warning) and then `open(DEFAULT_IMAGE_PATH)` raises `FileNotFoundError` anyway → the endpoint returns a bare 404 "Default image not found" even though `/unknown-icon.png` exists on disk. The whole fallback mechanism can never work.

This hits any deployment where `DEFAULT_IMAGE_PATH` resolves differently between the server process cwd and the baked image layout — the fallback was clearly added for exactly that case (see the warning message), but the bug made it a no-op.

### Expected Behavior

When the default image is missing and the fallback file exists, the endpoint should serve the fallback image. Only when neither file exists should it 404.

### Steps to Reproduce

1. Start the API with `DEFAULT_IMAGE_PATH` pointing at a non-existent file, with `/unknown-icon.png` present.
2. `GET /provider-containers/nonexistent-provider/image` for a provider with no custom image in the DB.
3. Actual: 404 `Default image not found` (with a misleading "using fallback path" warning in the log). Expected: 200 with the fallback PNG.

Reproduced with a minimal unit test driving `get_provider_image` directly: with the current code, the "fallback served when default missing" case raises `HTTPException(404)`; patching `with open(path, "rb")` makes it return the fallback bytes.

### Version

latest (main, commit fe5c896)

### Relevant log output

```
WARNING Default image not found at <DEFAULT_IMAGE_PATH>, using fallback path: /unknown-icon.png
→ HTTP 404 "Default image not found"
```

I'll open a PR with the one-line fix (`open(path, "rb")` — plus hoisting the fallback path to a module-level constant) and three regression tests covering: default served, fallback served when default missing, 404 only when both missing.

simpleqt:
Fixes #6832

## What

`get_provider_image()` computes a fallback image path when `DEFAULT_IMAGE_PATH` is missing, logs "using fallback path", then opens `DEFAULT_IMAGE_PATH` anyway — the fallback branch is dead code and the endpoint 404s even when `/unknown-icon.png` exists. One-line semantic fix: `open(path, "rb")` instead of `open(DEFAULT_IMAGE_PATH, "rb")`; the fallback path is hoisted to a module constant (`KEEP_DEFAULT_IMAGE_FALLBACK_PATH`, default unchanged).

## Why it matters

The fallback exists precisely for deployments where the default path doesn't resolve inside the container. Because of the bug those deployments get a 404 on every provider image request that has no custom image in the DB, plus a misleading warning in the logs.

## Tests

New `tests/test_provider_images.py` drives `get_provider_image` directly with a stub session (no custom image) and monkeypatched module constants:

- default image is served when present
- fallback is served when the default is missing ← fails with `HTTPException(404)` on main, passes with the fix
- 404 only when both files are missing

Red/green verified locally: with the fix reverted the fallback test fails (2 of 3 fail on file-existence edge), with the fix all 3 pass.

## Linked issue

#6832 (created per repo policy, root cause with line-level detail there).


simpleqt:
Heads-up on the two red e2e jobs on this head (`run-sqlite-without-redis / run-tests`, `run-postgresql-without-redis / run-tests`): the failures are Playwright UI timeouts in tests unrelated to this change —

- `test_end_to_end.py::test_run_workflow_from_alert_and_incident` — `Locator.click: Timeout 5000ms exceeded`
- `test_filtering_sort_search_on_incidents.py::test_filter_by_static_facet[source]` and `::test_filter_timeframe_combination_with_queryparams` — facet locator wait/assert timeouts

This PR touches only `keep/api/routes/provider_images.py` (fallback path when the default provider image file is missing) plus a new test file for that route; none of the failing tests exercise provider image resolution. All other checks on this head are green, and the three new regression tests (red before / green after the fix) pass locally. Happy to rebase/re-run if that helps get the e2e matrix green.

